PR #1, opened 2024-07-04
[DOCS] Create docs that explains interfaces we want to use

hubgan:


packages/docs/interfaces.md
```diff
@@ -0,0 +1,99 @@
+# React Native Audio Context
+
```
michalsek: I would move the file into `/docs/interfaces.md` instead of `packages` and reserve potential `packages/docs` for doc website 🙃

packages/docs/interfaces.md
```diff
@@ -0,0 +1,99 @@
...
+## Project goal
+
+The main goal of the project is to reproduce the Web Audio API as accurate as possible in the React Native environment.
```
michalsek: Can you add it also to `/README.md`

PR #9, opened 2024-07-09
set up and created basic jsi::HostObject config with example

maciejmakowski2003:
Added setup for JSI host objects with some example.

Closes: #10

### DIFF
--- a/example/android/build.gradle
+++ b/example/android/build.gradle
@@ -1,23 +1,22 @@
 buildscript {
-    ext {
-        buildToolsVersion = "34.0.0"
-        minSdkVersion = 23
-        compileSdkVersion = 34
-        targetSdkVersion = 34
-        ndkVersion = "26.1.10909125"
-        kotlinVersion = "1.9.22"
-      kotlin_version = '2.0.0'
-    }
-    repositories {
-        google()
-        mavenCentral()
-    }
-    dependencies {
-        classpath("com.android.tools.build:gradle")
-        classpath("com.facebook.react:react-native-gradle-plugin")
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin")
-      classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-    }
+  ext {
+    buildToolsVersion = "34.0.0"
+    minSdkVersion = 23
+    compileSdkVersion = 34
+    targetSdkVersion = 34
+    ndkVersion = "26.1.10909125"
+    kotlin_version = '2.0.0'
+  }
+  repositories {
+    google()
+    mavenCentral()
+  }
+  dependencies {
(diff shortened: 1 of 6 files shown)


example/android/build.gradle
```diff
@@ -6,6 +6,7 @@ buildscript {
         targetSdkVersion = 34
         ndkVersion = "26.1.10909125"
         kotlinVersion = "1.9.22"
+      kotlin_version = '2.0.0'
```
michalsek: Nitpick: indentation. Also to we really need kotlin version set up two times? 🤔 
One with `kotlinVersion`, one with `kotlin_version`?

example/android/build.gradle
```diff
@@ -15,6 +16,7 @@ buildscript {
         classpath("com.android.tools.build:gradle")
         classpath("com.facebook.react:react-native-gradle-plugin")
         classpath("org.jetbrains.kotlin:kotlin-gradle-plugin")
+      classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
```
michalsek: Nitpick: indentation

src/JSIExample.ts
```diff
@@ -0,0 +1,68 @@
+import { NativeModules, Platform } from 'react-native';
```
michalsek: To keep things clean i would create separate directory and split this file into multiple.

```
 src/
   JSIExample/
     index.ts - export everyting that should be available to JSIExample lib "client"
     types.ts - declares and exports all types exposed by JSIExample
     types.d.ts - not necessary, but usualy it is better to keep declarations (especially those which modify global or window/document scopes) in declaration (d.ts) files.
     JSExample.ts - contains all of the code that configures the JSIExample library
```

src/JSIExample.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+// global func declaration for JSI functions
+declare global {
+  function nativeCallSyncHook(): unknown;
+  var __JSIExampleProxy: JSIExampleWrapper | undefined;
+}
+
+// Check if the constructor exists. If not, try installing the JSI bindings.
+if (global.__JSIExampleProxy == null) {
```
michalsek: I think it would be nice to have all of this code inside a function (or more than one), so we could structure the code and early exists better. F.e.

```ts
function createJSIExampleProxy(): JSIExampleWrapper {
  if (global.__JSIExampleProxy) {
    return global.__JSIExampleProxy;
  }
  
  verifyExpoGo();
  
  const JSIExampleModule = getJSIExample();

  verifyOnDevice(JSIExampleModule);
  installModule(JSIExampleModule);
  verifyInstallation();
  
  return global.__JSIExampleProxy;
}

// Call the creator and export what it returns
export default createJSIExampleProxy();
```

PR #11, opened 2024-07-10
chore: ios with jsi setup

hubgan:


src/JSIExample.ts
```diff
@@ -1,7 +1,7 @@
 import { NativeModules, Platform } from 'react-native';
 
 export interface JSIExampleWrapper {
-  helloWorld(): string;
+  helloWorld(num1: number, num2: number): string;
```
michalsek, line 4: can we rename it to multiply in this case? 🙃

Issue #16, opened 2024-07-10
Implement OscillatorNode on Android


maciejmakowski2003:
-Implement oscillator class in Kotlin
-Implement HybridClass in Cpp for oscillator using fbjni
-Implement HostObject representing this oscillator in order to access it on Javascript side

PR #24, opened 2024-07-11
Feat/android/oscillator

maciejmakowski2003:
-Implemented oscillator class in Kotlin containing whole sound generating logic and AudioContex native module which provides method for creating oscillator instance on JS side( `createOscillator( )`)
-Implemented c++ HybridClass for this oscillator
-Implemented jsi HostObject for this oscillator and injected it into js runtime

Issue: #16


android/src/main/java/com/audiocontext/Oscillator.kt
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  init {
+    mHybridData = initHybrid(reactContext.javaScriptContextHolder!!.get())
+
+    val bufferSize = AudioTrack.getMinBufferSize(
+      44100,
+      AudioFormat.CHANNEL_OUT_MONO, AudioFormat.ENCODING_PCM_16BIT)
+    this.audioTrack = AudioTrack(
+      AudioManager.STREAM_MUSIC, 44100, AudioFormat.CHANNEL_OUT_MONO,
```
michalsek: Would to be good to move `44100` to be a constant

android/src/main/java/com/audiocontext/Oscillator.kt
```diff
@@ -0,0 +1,82 @@
...
+    audioTrack.stop()
+    playbackThread?.join()
+  }
+
+  private fun generateSound() {
+    var wavePhase = 0.0
+    var phaseChange: Double
+
+    while(isPlaying) {
+      phaseChange = 2 * Math.PI * (frequency + detune) / 44100
```
michalsek: and use it also here :)

android/src/main/java/com/audiocontext/Oscillator.kt
```diff
@@ -0,0 +1,82 @@
...
+    this.audioTrack = AudioTrack(
+      AudioManager.STREAM_MUSIC, 44100, AudioFormat.CHANNEL_OUT_MONO,
+      AudioFormat.ENCODING_PCM_16BIT, bufferSize, AudioTrack.MODE_STREAM
+    )
+  }
+
+  external fun initHybrid(l: Long): HybridData?
+
+  fun start() {
+    if(isPlaying) return
```
michalsek:
```suggestion
    if (isPlaying) {
      return
    }
```
:)

android/src/main/java/com/audiocontext/Oscillator.kt
```diff
@@ -0,0 +1,82 @@
...
+    isPlaying = true
+    audioTrack.play()
+    playbackThread = Thread { generateSound() }.apply{ start()}
+  }
+
+  fun stop() {
+    if(!isPlaying) return
```
michalsek:
```suggestion
    if (!isPlaying) {
      return
    }
```
:)

PR #25, opened 2024-07-15
Feat/android/oscillator node

maciejmakowski2003:
-Implemented oscillator class in Kotlin containing whole sound generating logic and AudioContex native module which provides method for creating oscillator instance on JS side( `createOscillator( )`)
-Implemented c++ HybridClass for this oscillator
-Implemented jsi HostObject for this oscillator and injected it into js runtime
-Integrate it with AudioContext class having HybridClass na HostObject

Closes: #16 
Closes: #21


android/src/main/java/com/audiocontext/nodes/oscillator/OscillatorNode.kt
```diff
@@ -0,0 +1,113 @@
...
+
+    while(isPlaying) {
+      phaseChange = 2 * Math.PI * (frequency + detune) / context.sampleRate
+
+      for(i in buffer.indices) {
+        buffer[i] = when(waveType) {
+          WaveType.SINE -> (sin(wavePhase) * Short.MAX_VALUE).toInt().toShort()
+          WaveType.SQUARE -> ((if (sin(wavePhase) >= 0) 1 else -1) * Short.MAX_VALUE).toShort()
+          WaveType.SAWTOOTH -> ((2 * (wavePhase / (2 * Math.PI) - floor(wavePhase / (2 * Math.PI) + 0.5))) * Short.MAX_VALUE).toInt().toShort()
+          WaveType.TRIANGLE -> ((2 * abs(2 * (wavePhase / (2 * Math.PI) - floor(wavePhase / (2 * Math.PI) + 0.5))) - 1) * Short.MAX_VALUE).toInt().toShort()
```
michalsek: Can you move calculation of each type to separate function or method? So the `when` statement will be clean?

```
buffer[i] = when(waveType) {
  WaveType.SINE -> sineWave(wavePhase) // Wave.sine(wavePhase)
  WaveType.SQUARE -> squareWave(wavePhase) // Wave.square(wavePhase)
  etc -> itd
}
```

android/src/main/java/com/audiocontext/nodes/oscillator/WaveType.kt
```diff
@@ -0,0 +1,20 @@
+package com.audiocontext.nodes.oscillator
+
+enum class WaveType {
+  SINE,
+  SQUARE,
+  SAWTOOTH,
+  TRIANGLE;
+
+  companion object {
+    fun switchWaveType(type: String): WaveType {
```
michalsek:
```suggestion
    fun fromString(type: String): WaveType {
```

example/src/App.tsx
```diff
@@ -1,45 +1,46 @@
...
+
+  const startOscillator = () => {
+    oscillator.start();
+    oscillator2.start();
+  };
+
+  const stopOscillator = () => {
+    oscillator.stop();
+    oscillator2.stop();
   };
```
michalsek:
```suggestion
  const oscillatorRef = useRef<??? | null>(null);
  const secondaryOscillatorRef = useRef<??? | null>(null);
  
  useEffect(() => {
    const context = new AudioContext();
    
    const oscillator = context.createOscillator();
    const secondaryOscillator = context.createOscillator();
    
    oscillator.frequency = 200;
    
    oscillatorRef.current = oscillator;
    secondaryOscillatorRef.current = oscillator;
    
    return () => {
      // TODO: cleanup audio context and oscillators?
    };
  }, []);

  const startOscillator = () => {
    oscillatorRef.current?.start();
    secondaryOscillatorRef.current?.start();
  };

  const stopOscillator = () => {
    oscillatorRef.current?.stop();
    secondaryOscillatorRef.current?.stop();
  };
```

michalsek: This way, audio context and the oscillators will be created only on first render.

Note: in development builds, because of [reacts strict mode every](https://react.dev/reference/react/StrictMode) effect is created twice>

The empty function returned by `useEffect` is a cleanup function that will be used when particular effect is unmounted. More info here: https://react.dev/reference/react/useEffect

src/types.d.ts
```diff
@@ -0,0 +1,8 @@
+export interface AudioContextWrapper {
+  createOscillator(): Oscillator;
+}
+
+declare global {
+  function nativeCallSyncHook(): unknown;
```
michalsek: Do we need it typed? It seems to be unused in the code

src/index.ts
```diff
@@ -1,2 +1,21 @@
...
+import type { Float } from 'react-native/Libraries/Types/CodegenTypes';
+const { AudioContextModule } = NativeModules;
+
+interface Oscillator {
+  frequency: Float;
+  wave: 'sine' | 'square' | 'sawtooth' | 'triangle';
+  detune: Float;
+  start: () => void;
+  stop: () => void;
+}
```
michalsek:
```suggestion
type WaveType = 'sine' | 'square' | 'sawtooth' | 'triangle';

interface Oscillator {
  frequency: Float;
  wave: WaveType;
  detune: Float;
  start: () => void;
  stop: () => void;
}
```

also it would be great to move this types (and export them) to separate file `types.ts` (not `.d.ts`)

cpp/AudioContextHostObject.cpp
```diff
@@ -0,0 +1,29 @@
...
+        return propertyNames;
+    }
+
+    jsi::Value AudioContextHostObject::get(jsi::Runtime& runtime, const jsi::PropNameID& propNameId) {
+        auto propName = propNameId.utf8(runtime);
+
+        if(propName == "createOscillator") {
+            return jsi::Function::createFromHostFunction(runtime, propNameId, 0, [this](jsi::Runtime& runtime, const jsi::Value& thisValue, const jsi::Value* arguments, size_t count) -> jsi::Value {
+                return audiocontext_->createOscillator();
+            });
```
michalsek: I think it might be good to move the code to a named lambda outside of `get` method. Potentially this method will have a lot of other method getters, so best would be to keep the implementation of each getter short.

src/types.d.ts
maciejmakowski2003: if i remove it from that file, i will get a warning in index.ts 
"Element implicitly has an 'any' type because type 'typeof globalThis' has no index signature.ts(7017)"

Issue #18, opened 2024-07-10
Implement Oscillator class on IOS side


hubgan:
We should be able to start and stop oscillator from RN side

PR #27, opened 2024-07-16
feat(oscillator): Setup IOS Oscillator

hubgan:
Create AudioContext class which contains createOscillator method.
We can create multiple oscillators which we can start, stop and change frequency.

Issue: #18

### DIFF
--- a/ios/nodes/Oscillator/OscillatorNode.m
+++ b/ios/nodes/Oscillator/OscillatorNode.m
@@ -46,7 +46,7 @@
     AVAudioFrameCount bufferFrameCapacity = (AVAudioFrameCount)self.sampleRate;
     
     // Convert cents to HZ
-    double detuneRatio = pow(2.0, self.detune / 1200.0);
+    double detuneRatio = pow(2.0, self.detune / OCTAVE_IN_CENTS);
     double detunedFrequency = detuneRatio * self.frequency;
     double phaseIncrement = FULL_CIRCLE_RADIANS * detunedFrequency / self.sampleRate;
     double phase = 0.0;
--- a/ios/nodes/Oscillator/OscillatorNode.h
+++ b/ios/nodes/Oscillator/OscillatorNode.h
@@ -3,6 +3,8 @@
 #import "WaveType.h"
 #import "AudioScheduledSourceNode.h"
 
+static const double OCTAVE_IN_CENTS = 12 * 100;
+
 @interface OscillatorNode : AudioScheduledSourceNode
 
 @property (nonatomic, strong) AVAudioEngine *audioEngine;


src/AudioContext.ts
```diff
@@ -0,0 +1,9 @@
+import Oscillator from './Oscillator/Oscillator';
+
+export class AudioContext {
+  constructor() {}
+
+  public createOscillator(frequency: number) {
+    return Oscillator(frequency);
+  }
```
hubgan: I think I should add TS OscillatorNode class

src/utils/install.ts
```diff
@@ -0,0 +1,50 @@
...
+    );
+  }
+}
+
+export function installModule(Module: any) {
+  const result = Module.install();
+  if (result !== true)
+    throw new Error(
+      `Failed to install react-native-audio-context: The native Oscillator Module could not be installed! Looks like something went wrong when installing JSI bindings: ${result}`
+    );
```
michalsek:
```suggestion
  if (result !== true) {
      throw new Error(
        `Failed to install react-native-audio-context: The native Oscillator Module could not be installed! Looks like something went wrong when installing JSI bindings: ${result}`
      );
    };
```

src/AudioContext.ts
michalsek: It might be good to also create some kind of `ACBase` class which will implement code that will have to be present in each class of the rnac.

f.e.
```
class ACBase {
  constructor() {
    verifyInstallation();
  }
}

class Oscillator extends ACBase {
  constructor() {
    super();
    
    this.hakuna = 'matata';
  }
}
```

ios/PlatformOscillator.mm
```diff
@@ -0,0 +1,22 @@
+#include <PlatformOscillator.h>
+#include <iostream>
+
+namespace audiocontext {
+
+PlatformOscillator::PlatformOscillator(const float frequency) {
```
michalsek: Nit: ident for everything in the namespace?

ios/IOSOscillator.m
```diff
@@ -0,0 +1,61 @@
...
+}
+
+- (void)stop {
+    [self.playerNode stop];
+}
+
+- (void)setBuffer {
+    AVAudioFrameCount bufferFrameCapacity = (AVAudioFrameCount)self.sampleRate;
+
+    double thetaIncrement = 2.0 * M_PI * self.frequency / self.sampleRate;
```
michalsek: magic number detected!

ios/IOSOscillator.m
```diff
@@ -0,0 +1,61 @@
...
+    double theta = 0.0;
+    float *audioBufferPointer = self.buffer.floatChannelData[0];
```
michalsek: are we sure that we can always access the `[0]` of `buffer.floatChannelData`? 👀

ios/AudioContextModule.mm
```diff
@@ -0,0 +1,53 @@
...
+            0,
+            [](jsi::Runtime& runtime, const jsi::Value& thisVal, const jsi::Value* args, size_t count) -> jsi::Value {
+                const float frequency = static_cast<float>(args[0].asNumber());
+
+                auto cppOscillatorHostObjectPtr = std::make_shared<audiocontext::OscillatorHostObject>(frequency);
+
+                const auto& jsiOscillatorHostObject = jsi::Object::createFromHostObject(runtime, cppOscillatorHostObjectPtr);
+
+                return jsi::Value(runtime, jsiOscillatorHostObject);
+            }
```
michalsek: named lambda? :)

example/src/App.tsx
```diff
@@ -1,15 +1,56 @@
...
+  const [frequency, setFrequency] = useState('440');
+  const osc = useRef<OscillatorNode | null>(null);
+
+  const start = useCallback(() => {
+    if (!osc.current) {
+      const context = new AudioContext();
+      osc.current = context.createOscillator(+frequency);
+    }
+
+    osc.current.start();
```
michalsek: both oscillator and audio context would be better to be created in useEffect (and cleaned up if required / possible)

example/src/App.tsx
```diff
@@ -1,15 +1,56 @@
...
+  }, [frequency]);
+
+  const stop = useCallback(() => {
+    if (!osc.current) {
+      return;
+    }
+
+    osc.current.stop();
```
michalsek:
```suggestion
    osc.current?.stop();
```
ts to the rescue!

example/src/App.tsx
```diff
@@ -1,15 +1,56 @@
...
+  }, [frequency]);
+
+  const getFrequency = useCallback(() => {
+    if (!osc.current) {
+      return;
+    }
+
+    console.log(osc.current.frequency);
+
+    return osc.current.frequency;
```
michalsek:
```suggestion
    return osc.current?.frequency ?? defaultFrequency;
```

example/src/App.tsx
```diff
@@ -1,15 +1,56 @@
...
     <View style={styles.container}>
-      <Text>{multiply()}</Text>
+      <Button title="Start1" onPress={start} />
+      <Button title="Stop1" onPress={stop} />
+      <TextInput
+        value={frequency}
+        onChangeText={(text) => setFrequency(text)}
+      />
+      <Button title="Change frequency" onPress={changeFrequency} />
+      <Button title="Get current frequency" onPress={getFrequency} />
```
michalsek: Having get/set buttons seems kinda weird (even in a test/example application :)). 
I would just go with `useEffect` that reacts to state changes or implement a wrapper around `setFrequency` that will update both state and and oscillator property

example/src/App.tsx
```diff
@@ -1,15 +1,56 @@
...
-import JSIExample from '../../src/JSIExample/JSIExample';
+import React, { useCallback, useRef, useState } from 'react';
+import { StyleSheet, View, Button, TextInput } from 'react-native';
+import { AudioContext, type OscillatorNode } from 'react-native-audio-context';
 
 const App: React.FC = () => {
-  const multiply = () => {
-    return JSIExample.multiply(2, 3);
-  };
+  const [frequency, setFrequency] = useState('440');
```
michalsek: why string? :)

ios/IOSOscillator.m
hubgan: Yes, we create AVAudioFormat using initStandardFormatWithSampleRate:channels:

image

According to documentation floatChannelData will return pointer if the buffer format is 32-bit float and nil otherwise.

image

I think it is fine for now.

example/src/App.tsx
```diff
@@ -1,15 +1,46 @@
...
+  const osc2 = useRef<OscillatorNode | null>(null);
+
+  useEffect(() => {
+    if (!osc.current) {
+      const context = new AudioContext();
+      osc.current = context.createOscillator(+frequency);
+      osc2.current = context.createOscillator(+frequency);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
hubgan: Temporary if until we will have an option for cleaning memory using useEffect cleanup function

src/modules/global.d.ts
```diff
@@ -0,0 +1,11 @@
+import type { Oscillator } from '../types';
+
+type AudioContext = {
+  createOscillator: () => Oscillator;
+  destination: AudioDestinationNode | null;
+};
```
hubgan, line 6: Temporary nulls until destination is implemented on ios

cpp/AudioDestinationNode/AudioDestinationNodeWrapper.h
```diff
@@ -8,7 +8,6 @@
 #endif
 
 namespace audiocontext {
-    using namespace facebook;
```
michalsek, line 11: Have you checked if it won't break anything on android?

ios/nodes/Oscillator/IOSOscillator.mm
```diff
@@ -0,0 +1,43 @@
...
+
+	void IOSOscillator::changeFrequency(const float frequency) const {
+		[OscillatorNode_ changeFrequency:frequency];
+	}
+
+	float IOSOscillator::getFrequency() const {
+		return [OscillatorNode_ getFrequency];
+	}
+
+    void IOSOscillator::changeDetune(const float detune) const {
```
michalsek: odd formating

ios/nodes/Oscillator/OscillatorNode.m
```diff
@@ -0,0 +1,90 @@
...
+        [self.audioEngine connect:self.playerNode to:self.audioEngine.mainMixerNode format: self.format];
+        [self setBuffer];
+
+        NSError *error = nil;
+        if (![self.audioEngine startAndReturnError:&error]) {
+            NSLog(@"Error starting audio engine: %@", [error localizedDescription]);
+        }
+    }
+
+    return self;
```
michalsek:
```suggestion
    if (self != [super init]) {
      return self;
    }
 
    self.frequency = 440;
    self.detune = 0;
    self.sampleRate = 44100;
    self.waveType = WaveTypeSine;

    self.audioEngine = [[AVAudioEngine alloc] init];
    self.playerNode = [[AVAudioPlayerNode alloc] init];

    self.format = [[AVAudioFormat alloc] initStandardFormatWithSampleRate:self.sampleRate channels:1];
    AVAudioFrameCount bufferFrameCapacity = (AVAudioFrameCount)self.sampleRate;
    self.buffer = [[AVAudioPCMBuffer alloc] initWithPCMFormat:self.format frameCapacity:bufferFrameCapacity];
    self.buffer.frameLength = bufferFrameCapacity;

    [self.audioEngine attachNode:self.playerNode];
    [self.audioEngine connect:self.playerNode to:self.audioEngine.mainMixerNode format: self.format];
    [self setBuffer];

    NSError *error = nil;

    if (![self.audioEngine startAndReturnError:&error]) {
      NSLog(@"Error starting audio engine: %@", [error localizedDescription]);
    }

    return self;
```

ios/nodes/Oscillator/OscillatorNode.m
```diff
@@ -0,0 +1,90 @@
...
+
+- (void)stop {
+    [self.playerNode stop];
+}
+
+- (void)setBuffer {
+    AVAudioFrameCount bufferFrameCapacity = (AVAudioFrameCount)self.sampleRate;
+    
+    // Convert cents to HZ
+    double detuneRatio = pow(2.0, self.detune / 1200.0);
```
michalsek: magic numbers ;-)

hubgan: It's math formula i found on wikipedia: https://en.wikipedia.org/wiki/Cent_(music)

image

I don't know if there is a good way to name these constants.

cpp/AudioDestinationNode/AudioDestinationNodeWrapper.h
hubgan: Yes, this namespace comes from jsi header file which is not used in this file.

image

ios/nodes/Oscillator/OscillatorNode.m
michalsek: OctaveInCents :)

michalsek: best defined as `dobule OctaveInCents = 12 * 100`

Issue #34, opened 2024-07-18
Implement c++ wrappers for AudioContext objects


maciejmakowski2003:
Implement c++ wrappers  (wrapper header file) providing interface for methods used in HostObjects. Each platform(ios, android) is going to have own implementation of those methods (wrapper source file). This approach would be helpful in parallel work on both platforms and would make each HostObject implementation platform independent .

PR #35, opened 2024-07-18
Feat/android/audio context wrapper

maciejmakowski2003:
-Implemented wrappers' header files for `AudioContext`, `AudioDestinationNode` and `OscillatorNode`
-Provided wrappers implementation on Android
-Introduced new cpp folder structure
-Refactored jni and jsi classes in order to be consistent with `Single Responsibility Principle` and maintain great abstraction level (HybridClass is responsible for interaction with kotlin side(fbjni), wrappers play a compositional role and whole jsi logic is locked in HostObject's classes)

Closes: #34

### DIFF
--- a/cpp/OscillatorNode/OscillatorNodeHostObject.cpp
+++ b/cpp/OscillatorNode/OscillatorNodeHostObject.cpp
@@ -22,7 +22,7 @@ namespace audiocontext
 
     if (propName == "start")
     {
-        return jsi::Function::createFromHostFunction(runtime, propNameId, 0, [this](jsi::Runtime &rt, const jsi::Value &thisValue, const jsi::Value *args, size_t count) -> jsi::Value
+        return jsi::Function::createFromHostFunction(runtime, propNameId, 1, [this](jsi::Runtime &rt, const jsi::Value &thisValue, const jsi::Value *args, size_t count) -> jsi::Value
         {
             auto time = args[0].getNumber();
             wrapper_->start(time);
@@ -32,7 +32,7 @@ namespace audiocontext
 
     if (propName == "stop")
     {
-        return jsi::Function::createFromHostFunction(runtime, propNameId, 0, [this](jsi::Runtime &rt, const jsi::Value &thisValue, const jsi::Value *args, size_t count) -> jsi::Value
+        return jsi::Function::createFromHostFunction(runtime, propNameId, 1, [this](jsi::Runtime &rt, const jsi::Value &thisValue, const jsi::Value *args, size_t count) -> jsi::Value
         {
             auto time = args[0].getNumber();
             wrapper_->stop(time);


android/src/main/cpp/OscillatorNode.h
```diff
@@ -29,8 +28,8 @@ namespace audiocontext {
             });
         }
 
-        void start();
-        void stop();
+        void start(double time);
```
michalsek, line 31: why `double` not `int`?

cpp/AudioContext/AudioContextHostObject.cpp
```diff
@@ -14,11 +14,17 @@ namespace audiocontext {
         auto propName = propNameId.utf8(runtime);
 
         if(propName == "createOscillator") {
-            return createOscillator(runtime, propNameId);
+            return jsi::Function::createFromHostFunction(runtime, propNameId, 0, [this](jsi::Runtime& runtime, const jsi::Value& thisValue, const jsi::Value* arguments, size_t count) -> jsi::Value {
+                auto oscillator = wrapper_->createOscillator();
+                auto oscillatorHostObject = OscillatorNodeHostObject::createFromWrapper(oscillator);
+                return jsi::Object::createFromHostObject(runtime, oscillatorHostObject);
+            });
```
michalsek, line 21: I think that `createFromHostFunction` is unnecesary also here 🤔

cpp/OscillatorNode/OscillatorNodeHostObject.cpp
```diff
@@ -0,0 +1,102 @@
...
+    return propertyNames;
+  }
+
+  jsi::Value OscillatorNodeHostObject::get(jsi::Runtime &runtime, const jsi::PropNameID &propNameId)
+  {
+    auto propName = propNameId.utf8(runtime);
+
+    if (propName == "start")
+    {
+        return jsi::Function::createFromHostFunction(runtime, propNameId, 0, [this](jsi::Runtime &rt, const jsi::Value &thisValue, const jsi::Value *args, size_t count) -> jsi::Value
```
michalsek:
```suggestion
        return jsi::Function::createFromHostFunction(runtime, propNameId, 1, [this](jsi::Runtime &rt, const jsi::Value &thisValue, const jsi::Value *args, size_t count) -> jsi::Value
```

cpp/OscillatorNode/OscillatorNodeHostObject.cpp
```diff
@@ -0,0 +1,102 @@
...
+        {
+            auto time = args[0].getNumber();
+            wrapper_->start(time);
+            return jsi::Value::undefined();
+        });
+    }
+
+    if (propName == "stop")
+    {
+        return jsi::Function::createFromHostFunction(runtime, propNameId, 0, [this](jsi::Runtime &rt, const jsi::Value &thisValue, const jsi::Value *args, size_t count) -> jsi::Value
```
michalsek:
```suggestion
        return jsi::Function::createFromHostFunction(runtime, propNameId, 1, [this](jsi::Runtime &rt, const jsi::Value &thisValue, const jsi::Value *args, size_t count) -> jsi::Value
```

android/src/main/cpp/OscillatorNode.h
maciejmakowski2003: time is in seconds according to Web Audio API, so for example we could want to start playing after 0.5 second

cpp/AudioContext/AudioContextHostObject.cpp
maciejmakowski2003: If I remove it I have to use this function in js like below
`__AudioContext.createOscillator` instead of `__AudioContext.createOscillator()`. For me it is a bit more intuitive to treat creating oscillator as a function call, however if you think it is better approach, I will change it.

PR #39, opened 2024-07-24
Feat/android/gain node

maciejmakowski2003:
-Implemented classes for GainNode and whole necessary logic in AudioContext classes
-Implemented get method in AudioNodeHostObject in order to call it from derived classes
-Refactored imports, friend class marks, etc.

Closes: #22


android/src/main/cpp/AudioDestinationNode.cpp
```diff
@@ -1,7 +1 @@
 #include "AudioDestinationNode.h"
-
-namespace audiocontext {
-
-    using namespace facebook::jni;
```
michalsek, line 5: do we need this file then? 🤔

example/src/App.tsx
```diff
@@ -1,71 +1,156 @@
...
+        <Text>Gain: {gain.toFixed(1)}</Text>
+        <Slider
+          containerStyle={styles.slider}
+          value={gain}
+          onValueChange={(value) =>
+            handleSliderChange(value, 0.0, setGain, gainRef, 'gain')
+          }
+          minimumValue={0.0}
+          maximumValue={1.0}
+          step={0.1}
```
michalsek: Nitpick: why only 10 steps instead of 100 with step of `0.01` or even go with `[0, 100]` scope and normalize it when setting a gainRef property

example/src/App.tsx
```diff
@@ -1,71 +1,156 @@
...
+    setValue: (value: number) => void,
+    ref: React.MutableRefObject<{ [key: string]: any } | null>,
+    propName: string
+  ) => {
+    const newValue = value[0] || defaultValue;
+    setValue(newValue);
+    if (ref.current) {
+      ref.current[propName] = newValue;
+    }
   };
```
michalsek: Didn't notice is earlier, why so over-complicated?
```suggestion
  const handleSliderChange = (value: number[]) => {
    const newValue = value[0] || 0.0;
 
    setValue(newValue);
    
    ganinRef.current?.gain = value;
  };
```

if we would like to have generic way of modify'ing nodes via sliders we could create a hook and/or components for that

```ts
useNodeValue(node: AudioNode) {
   // state
   // callbacks
   // return values?
}

...

useNodeValue(gainNodeRef.current);
```

android/src/main/cpp/AudioDestinationNode.cpp
maciejmakowski2003: you are right

example/src/App.tsx
maciejmakowski2003: Currently I have implemented three handlers each for parameter. In future I will introduce sth like a hook that u mentioned.

Issue #20, opened 2024-07-10
Implement StereoPannerNode on IOS side


hubgan:
We should be able to change pan value from RN side

PR #51, opened 2024-07-29
Feat/ios/create stereo panner node

hubgan:
Issue: #20


ios/nodes/GainNode/GainNode.m
```diff
@@ -31,13 +33,13 @@ - (void)deprocess:(AVAudioPCMBuffer *)buffer playerNode:(AVAudioPlayerNode *)pla
     }
 }
 
-- (void)changeGain:(float)gain {
-    self.gain = gain;
+- (void)setGain:(float)gain {
+    _gain = gain;
     [self.context processNodes];
 }
```
hubgan: `@property (nonatomic, assign) float gain;`

Turns out that above line create function setGain and getGain which i'm overriding. Using `self.gain = gain;` inside it will cause calling function recursively. For safety i change self to _ wherever is possible.

Issue #42, opened 2024-07-24
Implement methods for managing value/object state in given interval/point of time on Android


maciejmakowski2003:
-Refactor `start(time)` and `stop(time)` methods for `AudioScheduledSourceNode`
-Implement `setValueAtTime(value, startTime)` method
-Implement `linearRampToValueAtTime(value, endTime)` method
-Implement `exponentialRampToValueAtTime(value, endTime)` method
-Implement `setTargetAtTime(target, startTime, timeConstant)` method
-Implement `setValueCurveAtTime(values, startTime, duration)` method

PR #54, opened 2024-07-29
Feat/android/time scheduling

maciejmakowski2003:
- [x] Refactored start(time) and stop(time) methods for AudioScheduledSourceNode
- [x] Implemented setValueAtTime(value, startTime) method
- [x] Implemented linearRampToValueAtTime(value, endTime) method
- [x] Implemented exponentialRampToValueAtTime(value, endTime) methodhod

Closes: #42

### DIFF
--- a/android/src/main/java/com/audiocontext/nodes/GainNode.kt
+++ b/android/src/main/java/com/audiocontext/nodes/GainNode.kt
@@ -9,7 +9,7 @@ import com.audiocontext.nodes.parameters.PlaybackParameters
 class GainNode(context: BaseAudioContext): AudioNode(context) {
   override val numberOfInputs: Int = 1
   override val numberOfOutputs: Int = 1
-  private var gain: AudioParam = AudioParam(context,1.0, 1.0, 0.0)
+  private var gain: AudioParam = AudioParam(context,0.5, 1.0, 0.0)
     get() = field
     set(value) {
       field = value
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -42,7 +42,7 @@ const App: React.FC = () => {
 
       panRef.current = audioContextRef.current.createStereoPanner();
       panRef.current.pan.value = INITIAL_PAN;
-      
+
       oscillatorRef.current.connect(gainRef.current);
       gainRef.current.connect(panRef.current);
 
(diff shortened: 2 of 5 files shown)


android/src/main/java/com/audiocontext/nodes/oscillator/OscillatorNode.kt
```diff
@@ -1,66 +1,25 @@
...
 import com.audiocontext.nodes.AudioScheduledSourceNode
-import com.audiocontext.nodes.parameters.PlaybackParameters
-import com.facebook.jni.HybridData
+import kotlin.math.pow
 
 class OscillatorNode(context: BaseAudioContext) : AudioScheduledSourceNode(context) {
-  override val numberOfInputs: Int = 0
-  override val numberOfOutputs: Int = 1
-  private var frequency: AudioParam = AudioParam(440.0, 1200.0, 100.0)
+  private var frequency: AudioParam = AudioParam(context, 440.0, context.sampleRate/2.0, 0.0)
```
michalsek, line 12: Can you write short comment why we pass `context.sampleRate / 2.0`?

android/src/main/java/com/audiocontext/nodes/parameters/AudioParam.kt
```diff
@@ -29,10 +40,79 @@ class AudioParam(defaultValue: Double, maxValue: Double, minValue: Double) {
     return value
   }
 
+  @RequiresApi(Build.VERSION_CODES.N)
+  fun getValueAtTime(time: Double): Double {
+    if (changeQueue.isNotEmpty()) {
+      if (!currentChange.isPresent || currentChange.get().endTime < time) { //when the current change is over or there is no current change
+        currentChange = changeQueue.poll()?.let { Optional.of(it) }!!
+      }
+    }
```
michalsek:
```suggestion
    if (changeQueue.isNotEmpty()) {
      // when the current change is over or there is no current change
      if (!currentChange.isPresent || currentChange.get().endTime < time) {
        currentChange = changeQueue.poll()?.let { Optional.of(it) }!!
      }
    }
```

android/src/main/java/com/audiocontext/nodes/oscillator/OscillatorNode.kt
maciejmakowski2003: sure, I will attach link to Web Audio API docs, which defines it.

PR #98, opened 2024-08-22
Feat/ios/stereo panner node

### LINKED ISSUE #77: Create Stereo Panner Node compatible with AVAudioSourceNode on IOS
Since we're deleted AVAudioPlayerNode there is no obvious way to create StereoPannerNode. Probably we will need to implement it from scratch.

### LINKED ISSUE #82: [iOS] Validate and fix buffer creation and management in Oscillator node

maciejmakowski2003:
- [x] Implement StereoPannerNode
- [x] Created PlaybackParameters class, which stores all paramteres passed through audio processing flow
- [x] Update API roadmap
- [x] Refactor nodes to use PlaybackParameters

Closes #82 
Closes #77

### DIFF
--- a/ios/nodes/BiquadFilterNode/BiquadFilterNode.m
+++ b/ios/nodes/BiquadFilterNode/BiquadFilterNode.m
@@ -17,9 +17,9 @@
         self.numberOfOutputs = 1;
         _filterType = FilterTypeLowpass;
     }
-    
+
     [self resetCoefficients];
-    
+
     return self;
 }
 
@@ -28,7 +28,7 @@
     _x2 = 0;
     _y1 = 0;
     _y2 = 0;
-    
+
     _a0 = 1;
     _a1 = 0;
     _a2 = 0;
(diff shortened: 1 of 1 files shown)


ios/nodes/Oscillator/OscillatorNode.m
```diff
@@ -17,8 +17,9 @@ - (instancetype)initWithContext:(AudioContext *)context {
     self.numberOfOutputs = 1;
     self.numberOfInputs = 0;
 
-    _format = [[AVAudioFormat alloc] initStandardFormatWithSampleRate:self.context.sampleRate channels:1];
+    _format = [[AVAudioFormat alloc] initStandardFormatWithSampleRate:self.context.sampleRate channels:2];
```
michalsek, line 20: Some things to consider:
- According to [specification](https://webaudio.github.io/web-audio-api/#OscillatorNode) Oscillator is always 2-channel, so we should make it also 2 channel on android 
- As our input node is 2-channel, we should go through every audio node and make sure it is able to work with both channels, so no data is lost
- Specification says that StereoPanner accepts 2-channel input and 2-channel output, but in case the input has less or more channels its using [up or down-mixing](https://webaudio.github.io/web-audio-api/#channel-up-mixing-and-down-mixing)
- Also interesting to read might be the [panning algorithm](https://webaudio.github.io/web-audio-api/#Spatialization-panning-algorithm) described in the spec

PR #103, opened 2024-08-27
buffer and conquer: free-play with buffers and audio nodes

michalsek:


example/ios/AudioContextExample.xcodeproj/project.pbxproj
```diff
@@ -419,6 +419,7 @@
 			baseConfigurationReference = 5B7EB9410499542E8C5724F5 /* Pods-AudioContextExample-AudioContextExampleTests.debug.xcconfig */;
 			buildSettings = {
 				BUNDLE_LOADER = "$(TEST_HOST)";
+				DEVELOPMENT_TEAM = WC59N2X3FV;
```
michalsek, line 422: nope

PR #130, opened 2024-09-20
feat: implemented navigation

### LINKED ISSUE #109: Add react-navigation to demo app

maciejmakowski2003:
- Added navigation
- Reorganize examples

Closes #109

### DIFF
--- a/apps/common-app/src/App.tsx
+++ b/apps/common-app/src/App.tsx
@@ -1,43 +1,43 @@
 import React from 'react';
+import type { FC } from 'react';
 import { ScrollView, StyleSheet, Text, TouchableOpacity } from 'react-native';
 import { NavigationContainer, useNavigation } from '@react-navigation/native';
 import { createStackNavigator } from '@react-navigation/stack';
-import { Container } from './components/Container';
+
+import Container from './components/Container';
 import { Examples } from './examples';
 import { layout, colors } from './styles';
 
-//TODO add absolute imports
-
 const Stack = createStackNavigator();
 
-function HomeScreen() {
+const HomeScreen: FC = () => {
   const navigation = useNavigation();
 
   return (
     <Container centered={false}>
-      <ScrollView contentContainerStyle={{ padding: layout.spacing * 2 }}>
-        {Object.keys(Examples).map((key) => (
+      <ScrollView contentContainerStyle={styles.scrollView}>
+        {Object.entries(Examples).map(([key, example]) => (
           <TouchableOpacity
             onPress={() => navigation.navigate(key as never)}
             key={key}
             style={styles.button}
           >
-            <Text style={styles.title}>{Examples[key].title}</Text>
-            <Text style={styles.subtitle}>{Examples[key].subtitle}</Text>
+            <Text style={styles.title}>{example.title}</Text>
+            <Text style={styles.subtitle}>{example.subtitle}</Text>
           </TouchableOpacity>
         ))}
       </ScrollView>
     </Container>
   );
-}
+};
 
-export default function App() {
+const App: FC = () => {
   return (
     <NavigationContainer>
       <Stack.Navigator
         screenOptions={{
           headerStyle: {
-            backgroundColor: colors.blue,
+            backgroundColor: colors.darkblue,
           },
           headerTintColor: colors.white,
         }}
@@ -58,7 +58,7 @@ export default function App() {
       </Stack.Navigator>
     </NavigationContainer>
   );
-}
+};
 
 const styles = StyleSheet.create({
   title: {
@@ -77,4 +77,9 @@ const styles = StyleSheet.create({
     borderStyle: 'dashed',
     borderRadius: layout.radius,
   },
+  scrollView: {
+    padding: layout.spacing * 2,
+  },
 });
+
+export default App;
--- a/apps/common-app/src/components/Container.tsx
+++ b/apps/common-app/src/components/Container.tsx
@@ -1,18 +1,15 @@
 /* eslint-disable react-native/no-inline-styles */
-import type { ReactNode } from 'react';
+import type { PropsWithChildren, FC } from 'react';
 import { StyleProp, ViewStyle } from 'react-native';
-
 import { SafeAreaView } from 'react-native-safe-area-context';
 
-export function Container({
-  children,
-  style,
-  centered = true,
-}: {
-  children: ReactNode;
+type ContainerProps = PropsWithChildren<{
   style?: StyleProp<ViewStyle>;
   centered?: boolean;
-}) {
+}>;
+
+const Container: FC<ContainerProps> = (props) => {
+  const { children, style, centered } = props;
   return (
     <SafeAreaView
       style={[
@@ -27,4 +24,6 @@ export function Container({
       {children}
     </SafeAreaView>
   );
-}
+};
+
+export default Container;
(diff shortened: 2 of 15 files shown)


apps/common-app/src/examples/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    title: 'Metronome 🎸',
+    subtitle: 'Keep time with the beat',
+    screen: Metronome,
+  },
+  DrumMachine: {
+    title: 'Drum Machine 🥁',
+    subtitle: 'Create drum patterns',
+    screen: DrumMachine,
+  },
+} as const;
```
michalsek, line 27: You can export a type of this map to be used in the button list array:

```suggestion
export const Examples: Record<string, Example> = {
  Oscillator: {
    title: 'Oscillator 🔉',
    subtitle: 'Generate sound waves',
    screen: Oscillator,
  },
  Metronome: {
    title: 'Metronome 🎸',
    subtitle: 'Keep time with the beat',
    screen: Metronome,
  },
  DrumMachine: {
    title: 'Drum Machine 🥁',
    subtitle: 'Create drum patterns',
    screen: DrumMachine,
  },
} as const;
```
```suggestion
export const Examples: Record<string, Example> = {
  Oscillator: {
    title: 'Oscillator 🔉',
    subtitle: 'Generate sound waves',
    screen: Oscillator,
  },
  Metronome: {
    title: 'Metronome 🎸',
    subtitle: 'Keep time with the beat',
    screen: Metronome,
  },
  DrumMachine: {
    title: 'Drum Machine 🥁',
    subtitle: 'Create drum patterns',
    screen: DrumMachine,
  },
} as const;

export type ExampleKey = keyof typeof Examples;
```

apps/common-app/tsconfig.json
```diff
@@ -1,3 +1,9 @@
 {
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "@/*": ["./src/*"]
+    }
+  },
```
michalsek: For this import to work you also need to add something similar to babel.config of each example application, so react-native bundler will understand the pathing

apps/common-app/src/examples/Metronome/Metronome.tsx
```diff
@@ -0,0 +1,10 @@
+import { Container } from '../../components/Container';
+import { Text } from 'react-native';
```
michalsek: Incorrect import order -> I guess I might have broken eslint setup with this monorepo refactor. Usually it would complain right ahead

apps/common-app/src/App.tsx
```diff
@@ -1,241 +1,80 @@
...
-
-    if (!clapRef.current) {
-      clapRef.current = new Clap(audioContextRef.current);
-    }
-
-    clapRef.current.play(audioContextRef.current.currentTime);
-  };
+  return (
+    <Container centered={false}>
+      <ScrollView contentContainerStyle={{ padding: layout.spacing * 2 }}>
```
michalsek: could be moved to the stylesheet at the bottom of the file

apps/common-app/src/App.tsx
```diff
@@ -1,241 +1,80 @@
...
+        {Object.keys(Examples).map((key) => (
```
michalsek: `Object.entries` couldbe usefull here :)

apps/common-app/src/App.tsx
```diff
@@ -1,241 +1,80 @@
...
-import { Kick } from './soundEngines/Kick';
-import { HiHat } from './soundEngines/HiHat';
-import { Clap } from './soundEngines/Clap';
+const Stack = createStackNavigator();
 
-const INITIAL_FREQUENCY = 440;
-const INITIAL_DETUNE = 0;
-const INITIAL_GAIN = 1.0;
-const INITIAL_PAN = 0;
+function HomeScreen() {
```
michalsek: Usually it is better to write functional components using arrow functions, f.e. you can easily use `React.FC` to type them, f.e.
```ts
import React from 'react';
import type { FC, PropsWithChildren } from 'react';


const HomeScreen: React.FC = () => null; // this preciselly tells that component doesn't accept any props, as well as enforces you to return something from it.

const HomeScreen: FC<HomeScreenProps> = (props) => {
  const { spread, the } = props;
  return null;
} 

const HomeScreen: React.FC<PropsWithChildren<HomeScreenProps>> = (props) => {
  const { children, spread the } = props;
  return null;
}
```

apps/common-app/src/components/Container.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export function Container({
+  children,
+  style,
+  centered = true,
+}: {
+  children: ReactNode;
+  style?: StyleProp<ViewStyle>;
+  centered?: boolean;
+}) {
```
michalsek: It is cleaner to define the prop type before the component

```suggestion
type ContainerProps = PropsWithChildren<{
  style?: StyleProps<ViewStyle>;
  centered?: boolean;
}>;

const Container: FC<ContainerProps> = (props) => {
  const { children, style, centered = true } = props;
  
  // rest of the code, don't forget about imports! :)
}
```

apps/common-app/src/examples/DrumMachine/DrumMachine.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  useEffect(() => {
+    if (!audioContextRef.current) {
+      audioContextRef.current = new AudioContext();
+    }
+
+    return () => {
+      audioContextRef.current?.close();
+    };
+  }, []);
```
michalsek: Its more of a preference rather than convention, but I usually (and most of the devs too) order the component +- this way:

```ts
const Component: FC<Props> = (props) => {
   coonst { props, spread } = props;
   // state / data sources / refs hooks calls: useState, useRef, useReducer, useImaginaryAPIProvider
   
   // callbacks and functions that are passed as props to components below

   // useEffect and useLayoutEffect calls or other similar "reaction" hooks

   // render functions, such as renderItem often used when rendering arrays (and required in react native when using a Flat/SectionList
   
   return <TheFinalReturnStatement />;   
};
```

apps/common-app/src/examples/DrumMachine/DrumMachine.tsx
```diff
@@ -0,0 +1,69 @@
+import React from 'react';
+import { useRef, useEffect } from 'react';
+import { Button } from 'react-native';
+
+import { AudioContext } from 'react-native-audio-api';
+
+import { Container } from '../../components/Container';
+import { Kick, HiHat, Clap } from '../SharedUtils';
+
+export function DrumMachine() {
```
michalsek: If it is a the only component/function exported from given file (or main one) it is better to use `export default`.

Btw. when using `export default`, the `import/export * from 'x';` won't capture it :)

apps/common-app/tsconfig.json
maciejmakowski2003: what exactly should I add?

apps/common-app/src/App.tsx
```diff
@@ -1,240 +1,84 @@
...
-    }
-
-    clapRef.current.play(audioContextRef.current.currentTime);
-  };
+  return (
+    <Container centered={false}>
+      <ScrollView contentContainerStyle={styles.scrollView}>
+        {Object.entries(Examples).map(([key, example]) => (
+          <TouchableOpacity
+            onPress={() => navigation.navigate(key as never)}
```
michalsek, line 21: You should be able now to use `ExampleKey` instead of never, might require typing the function in `.map`

```
.map([key, example]: [ExampleKey, Example]) => (
```

PR #131, opened 2024-09-23
Feat/oscillator example

### LINKED ISSUE #127: [Android] Fix slider
or implement your own

maciejmakowski2003:
- Implemented custom silder
- Reorganize oscillator example
- Added switching oscillator type

Closes #127

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -15,6 +15,5 @@
     "forceConsistentCasingInFileNames": true,
     "resolveJsonModule": true,
     "allowJs": true,
-    "checkJs": true,
   }
 }


apps/common-app/src/App.tsx
```diff
@@ -16,16 +16,18 @@ const HomeScreen: FC = () => {
...
-            style={styles.button}
-          >
-            <Text style={styles.title}>{example.title}</Text>
-            <Text style={styles.subtitle}>{example.subtitle}</Text>
-          </TouchableOpacity>
-        ))}
+        {Object.entries(Examples).map(
+          ([key, example]: [ExampleKey, Example]) => (
+            <TouchableOpacity
+              onPress={() => navigation.navigate(key as never)}
```
michalsek:
```suggestion
              onPress={() => navigation.navigate(key)}
```

apps/common-app/src/examples/Oscillator/Oscillator.tsx
```diff
@@ -133,21 +152,23 @@ const Oscillator: FC = () => {
...
+      <CustomSlider
         style={styles.slider}
         value={detune}
         onValueChange={handleDetuneChange}
         minimumValue={0}
         maximumValue={100}
         step={1}
       />
+      <Text>Oscillator type: {oscillatorType}</Text>
+      <Button title={`Click to change type`} onPress={toggleOscillatorType} />
```
michalsek: This carousel toggle seems a bit wierd, can we have row of small buttons for each type + highlight for the currently active one?

apps/common-app/src/components/CustomSlider.tsx
```diff
@@ -0,0 +1,116 @@
...
+  maximumValue: number;
+  step: number;
+  style?: StyleProp<ViewStyle>;
+}
+
+const CustomSlider: FC<CustomSliderProps> = (props) => {
+  const { value, onValueChange, minimumValue, maximumValue, step, style } =
+    props;
+  const x = useSharedValue(0);
+  const [sliderWidth, setSliderWidth] = useState<number>(300);
```
michalsek: If you want the width to be defined by the user of the component, it should be also used in initialization. Preferably as a separate prop instead of extracting from `style` prop.

michalsek: also it would be better to have sliderWidth as share value, state changes generate component re-renders, usually when handling animations we would like to avoid that as much as possible.

apps/common-app/src/components/CustomSlider.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  const convertXToValue = useCallback(
+    (xValue: number) => {
+      const fraction = xValue / sliderWidth;
+      const newValue = minimumValue + fraction * (maximumValue - minimumValue);
+      runOnJS(onValueChange)(Math.round(newValue / step) * step);
```
michalsek: `runOnJS` call isn't necessary here: `convertXToValue` is always called inside runOnJS in the gesture callbacks

michalsek: Personally I would separate the calculation from setting a value / calling the onValueChange callback

apps/common-app/src/components/CustomSlider.tsx
```diff
@@ -0,0 +1,116 @@
...
+      ],
+    };
+  });
+
+  useEffect(() => {
+    if (style && 'width' in style) {
+      setSliderWidth(style.width as number);
+    }
+
+    runOnJS(convertValueToX)(value);
```
michalsek: `useEffect` isn't a reanimated related hook and runs only on the JS thread. Thus we don't need the `runONJS` function

```suggestion
    convertValueToX(value);
```

apps/common-app/src/components/CustomSlider.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }, [style, value, minimumValue, maximumValue, convertValueToX]);
+
+  return (
+    <View style={[styles.default, style]}>
+      <GestureDetector gesture={panGesture}>
+        <View
+          style={[styles.slider, styles.default, { width: sliderWidth }]}
```
michalsek: If we want to have variable-size slider, best approach I think would be to:
- use `sliderWidth` as an reanimated shared value
- add an `onLayout` property that sets the `sliderWidth` shared value

- or add ref to the top view component, add `useLayoutEffect` hook, that would call `measure` function on the view ref and sets up the `sliderWidth` shared value: https://reactnative.dev/docs/direct-manipulation

apps/common-app/src/components/CustomSlider.tsx
```diff
@@ -0,0 +1,116 @@
...
+          hitSlop={hitSlop}
+        >
+          <Animated.View style={[styles.progress, { width: x }]} />
```
michalsek: Animation of width is slower than translation! (width changes forces react-native to re-calculate layout of the component)

Instead you can make this view as same width as its container, add a `overflow: 'hidden'` to the parent and instead of animating the width, animate the translation of this View:

for a value equal to minimum value it would be translated on x by 100% of container width,
for a value equal to max value it would be translated on x by 0.

apps/common-app/src/components/CustomSlider.tsx
```diff
@@ -0,0 +1,116 @@
...
+        {
+          translateX: x.value,
+        },
+      ],
+    };
+  });
+
+  useEffect(() => {
+    if (style && 'width' in style) {
+      setSliderWidth(style.width as number);
```
michalsek: This won't work, in react native `style` can be one of 4 types:
- plain js object with styling properties passed directly to the component `style={{ width: 300 }}`
- style created with `StyleSheet.create({ })`, in this case, the the style property will be an integer - id of the registered style.
- a boolean :), people often use something like `style={amIInsane && styles.insanityBox}`
- an array of combination of the 3 above

In order to be able to access property in such mess, you have to call `StyleSheet.flatten` on the `style` property first https://reactnative.dev/docs/stylesheet#flatten

But I think in this case I would be misuse of the flatten function

apps/common-app/src/examples/Oscillator/Oscillator.tsx
maciejmakowski2003: ofc!

apps/common-app/src/App.tsx
```diff
@@ -16,16 +16,18 @@ const HomeScreen: FC = () => {
...
+              onPress={() => navigation.navigate(key)}
+              key={key}
+              style={styles.button}
+            >
+              <Text style={styles.title}>{example.title}</Text>
+              <Text style={styles.subtitle}>{example.subtitle}</Text>
+            </TouchableOpacity>
+          )
+        )}
       </ScrollView>
```
michalsek: In case of array of items of same shape etc. it is better to use `FlatList` component

apps/common-app/src/examples/Oscillator/Oscillator.tsx
```diff
@@ -134,21 +146,57 @@ const Oscillator: FC = () => {
...
 };
 
 const styles = StyleSheet.create({
-  slider: {
-    width: 300,
-    padding: 8,
+  oscillatorTypeContainer: {
+    flexDirection: 'row',
+    marginTop: 20,
+  },
```
michalsek:
```ts
  oscillatorTypeContainer: {
    flexDirection: 'row',
    marginTop: 20,
  },
```

Not something to fix or change, but small hint, towards such components. Usually I prefer to separate margin around/between components for the components style itself. It helps to reuse same component when there are different margin between components in different places. So for example:

```ts
return (
  <Screen>
    <FirstComponent />
    <Spacer.Vertical size={20} />
    <Row>
      {buttons.map(/* ... *.)}
    </Row>
  </Screen>
);
```

where spacer is something like:
```ts
const Vertical: FC<Props> = ({ size }) => {
  return <View style={{ height: size }} />
};

/* ... */

export default {
  Vertical,
  Horizontal,
}
```

PR #136, opened 2024-09-26
fix: expo 51 and rn 0.75 coexistence

michalsek:


tsconfig.json
```diff
@@ -16,7 +16,5 @@
     "resolveJsonModule": true,
     "allowJs": true,
     "checkJs": true,
-    // We set "noEmit" to `true` only in packages which emit files.
-    "noEmit": true
```
michalsek: Not sure if this is still relevant change 🤔

maciejmakowski2003: what do you mean?

PR #146, opened 2024-10-02
Refactor/introduce oboe

maciejmakowski2003:
- Rewrote whole audio processing logic to c++ using oboe library
- Refactored wrappers' and HostObjects' classes
- Changed general audio processing flow architecture. Currently access to system audio API is placed only in `AudioContext` class.

ToDo:
- [x] Refactor Oscillator callback
- [x] Implement AudioBuffor
- [x] Implement AudioBufforSourceNode
- [x] Implement AudioScheduledSourceNode
- [x] Conduct general refactoring

### DIFF
--- a/packages/react-native-audio-api/android/src/main/cpp/AudioBuffer/AudioBuffer.h
+++ b/packages/react-native-audio-api/android/src/main/cpp/AudioBuffer/AudioBuffer.h
@@ -3,6 +3,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <algorithm>
 
 namespace audioapi {
 


apps/common-app/src/examples/SharedUtils/soundEngines/HiHat.ts
```diff
@@ -14,41 +14,40 @@ class HiHat implements SoundEngine {
...
+
+      highpassFilter.type = 'highpass';
+      highpassFilter.frequency.value = this.highpassFilterFrequency;
+
+      gain.gain.setValueAtTime(0.0001, time);
+      gain.gain.exponentialRampToValueAtTime(this.volume, time + 0.02);
+      gain.gain.exponentialRampToValueAtTime(this.volume * 0.33, time + 0.03);
+      gain.gain.exponentialRampToValueAtTime(this.volume * 0.0001, time + 0.3);
+      gain.gain.setValueAtTime(0, time + 0.3 + 0.001);
+      //number of inputs of filter is 1 on android- check it
```
michalsek, line 44: So I did some digging in the webkit and labsound:

The reason it works correctly - having multiple oscillators connected to single filter - is that their `AudioBus` implementation does both source and channel mixing if it is necessary for given node.

And you've guessed right: it is just addition of values (and max value based normalization at the end)

Although one thing we should consider when implementing the source/channel mixing is the use of SSE opreations similarly to Webkit: you can check them in https://github.com/WebKit/WebKit/blob/main/Source/WebCore/platform/audio/VectorMath.cpp

michalsek: For the time being, we can live with having separate Biquads for every oscillator though

packages/react-native-audio-api/android/src/main/cpp/AudioBufferSourceNode/AudioBufferSourceNode.cpp
```diff
@@ -1,31 +1,54 @@
 #include "AudioBufferSourceNode.h"
+#include "AudioContext.h"
 
 namespace audioapi {
 
-using namespace facebook::jni;
+AudioBufferSourceNode::AudioBufferSourceNode(AudioContext *context)
+    : AudioScheduledSourceNode(context), loop_(false), bufferIndex_(0) {
+  numberOfInputs_ = 0;
+  buffer_ = std::make_shared<AudioBuffer>(2, 428, context->getSampleRate());
```
michalsek: why the particular value of `428`?

packages/react-native-audio-api/android/src/main/cpp/AudioBuffer/AudioBuffer.cpp
```diff
@@ -2,58 +2,87 @@
...
+      duration_(static_cast<double>(length) / sampleRate) {
+  if (numberOfChannels != 1 && numberOfChannels != 2) {
+    throw std::invalid_argument("Invalid number of channels");
+  }
 
-int AudioBuffer::getSampleRate() const {
-  static const auto method =
-      javaClassStatic()->getMethod<jint()>("getSampleRate");
-  return method(javaPart_);
+  channels_ = static_cast<float **>(malloc(numberOfChannels * sizeof(float *)));
```
michalsek: Since we are in C++, we can use 
- `new float*[numberOfChannels]` for allocation
- `delete[] channels_[i]` for deconstruction
- and `std::copy` instead of memcpy;

I've used the old-style `c` memory management as C++ operators arent available in ObjectiveC and they are far more convinent to use that ObjectiveC `NSMutableData` or other Nstypes :)

packages/react-native-audio-api/android/src/main/cpp/AudioBufferSourceNode/AudioBufferSourceNode.cpp
maciejmakowski2003: I forgot about it. I will encapsulate it in optional

PR #242, opened 2024-12-18
Feat/analyser node

### LINKED ISSUE #168: Implement AnalyserNode
https://developer.mozilla.org/en-US/docs/Web/API/AnalyserNode

https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/webaudio/AnalyserNode.cpp
https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/webaudio/RealtimeAnalyser.cpp

maciejmakowski2003:
Closes #168 

## Introduced changes

- Implemented `AnalyserNode` 
- Refactored `FFTFrame`
- Added `AudioVisualizer` example using AnalyserNode(early version)

## Checklist

- [x] Linked relevant issue


apps/common-app/src/examples/AudioVisualizer/Canvas.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const CanvasContext = createContext<CanvasContext>({
+  initialized: false,
+  size: { width: 0, height: 0 },
+  canvasPan: { x: 40, y: 0 },
+});
+
+const Canvas: React.FC<PropsWithChildren> = ({ children }) => {
+  const [size, setSize] = useState<Size>({ width: 0, height: 0 });
+  const [canvasPan] = useState<Position>({ x: 40, y: 0 });
```
michalsek: `canvasPan` can be removed (unless you want to use it later). I've planned to add pan & zoom effects to the spy app, but due to the bugs in mac catlyst+reanimated i've never finished that.

apps/common-app/src/examples/AudioVisualizer/Canvas.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const stableContext = useMemo(
+    () => ({
+      initialized: true,
+      size: { width: size.width, height: size.height },
+      canvasPan: { x: canvasPan.x, y: canvasPan.y },
+    }),
+    [size.width, size.height, canvasPan.x, canvasPan.y]
+  );
+
+  const context = useThrottledValue(stableContext, 300);
```
michalsek: Similar things goes to the throttling of the canvas values. I've written that so resizing the window of the catalyst mode wouldn't create a large number of resize onLayout events. 

In case of our example app, the only "resize" events we have to worry about is when user rotates the phone and as we don't really support that in other examples I guess we can disable this in xcode project settings :)

apps/common-app/src/examples/AudioVisualizer/Charts.tsx
```diff
@@ -0,0 +1,94 @@
...
+      />
+    </>
+  );
+};
+
+const Grid: React.FC = () => {
+  const { size, canvasPan } = useCanvas();
+
+  const inc = useMemo(() => {
+    return Math.max(size.width / 80, size.height / 80);
```
michalsek: The chart grid looks way too dense, something more light this should be more mobile-friendly :)

```suggestion
    return Math.max(size.width / 16, size.height / 16);
```

apps/common-app/src/examples/AudioVisualizer/AudioVisualizer.tsx
```diff
@@ -0,0 +1,128 @@
...
+    <Container>
+      <Layout.Row>
+        <Layout.Box>
+          <FreqTimeChart data={times} />
+        </Layout.Box>
+      </Layout.Row>
+      <Layout.LineHorizontal />
+      <Layout.Row>
+        <Layout.Box>
+          <FreqTimeChart data={freqs} />
```
michalsek: Looks like this chart is mostly rendering outside its viewport, maybe it should have a param like "maxValue" so it would be easier to scale the data, so everything will stay visible?

apps/common-app/src/examples/Piano/Piano.tsx
```diff
@@ -20,13 +20,12 @@ const Piano: FC = () => {
     notesRef.current?.[key].stop();
   };
 
-  useEffect(() => {}, []);
```
michalsek, line 23: 👏

apps/common-app/src/styles.ts
```diff
@@ -17,6 +17,7 @@ export const colors = {
   gray: '#d7d7d7',
 
   background: '#222222',
+  backgeoundDark: '#1f2020',
```
michalsek: typo! :)
```suggestion
  backgroundDark: '#1f2020',
```

packages/react-native-audio-api/common/cpp/core/AnalyserNode.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+  const float magnitudeScale = 1.0f / static_cast<float>(fftSize_);
+
+  for (size_t i = 0; i < magnitudeBuffer_->getSize(); i++) {
+    std::complex<float> c(realFFTFrameData[i], imaginaryFFTFrameData[i]);
+    auto scalarMagnitude = std::abs(c) * magnitudeScale;
+    magnitudeBuffer_->getData()[i] = static_cast<float>(
+        smoothingTimeConstant_ * magnitudeBuffer_->getData()[i] +
+        (1 - smoothingTimeConstant_) * scalarMagnitude);
+  }
```
michalsek: Could you add some comments here, so it would be easier to get back to this code?

F.e. why tempBuffer is required, why this particular start and stop values when coping data to it etc

michalsek (approved):
Looks great! 🔥 Can't wait to play with it myself 😈

packages/react-native-audio-api/src/core/AnalyserNode.ts
```diff
@@ -0,0 +1,85 @@
...
+  ];
+
+  public get fftSize(): number {
+    return (this.node as IAnalyserNode).fftSize;
+  }
+
+  public set fftSize(value: number) {
+    if (!AnalyserNode.allowedFFTSize.includes(value)) {
+      throw new IndexSizeError(
+        `Provided value (${value}) must be a power of 2 2^5 and 2^15`
```
michalsek:
```suggestion
        `Provided value (${value}) must be a power of 2 between 32 and 32768`
```

PR #338, opened 2025-03-05
feat: OfflineAudioContext RNAA-11

### LINKED ISSUE #256: Implement OfflineAudioContext
### Feature Request

OfflineAudioContext will be usefull for automated testing

the spec: https://developer.mozilla.org/en-US/docs/Web/API/OfflineAudioContext

orange4glace:
Closes #256

## Introduced changes

Implementing [OfflineAudioContext](https://developer.mozilla.org/ko/docs/Web/API/OfflineAudioContext)

Still working on it.

-

## Checklist

- [x] Linked relevant issue
- [x] Added support for web


maciejmakowski2003:
Aru you going to continue your work?

maciejmakowski2003:
Of course, It was not my intention to hurry you up. Just checkin ;))

orange4glace:
Implemented [`resume()`](https://developer.mozilla.org/en-US/docs/Web/API/OfflineAudioContext/resume) and [`suspend()`](https://developer.mozilla.org/en-US/docs/Web/API/OfflineAudioContext/suspend) api.
`suspend()` locks`stateLock` and check whether it should be suspended on every quantum frame.

packages/react-native-audio-api/common/cpp/audioapi/core/OfflineAudioContext.cpp
```diff
@@ -0,0 +1,149 @@
...
+        "or equal to the total " +
+        "render duration of " +
+        std::to_string(numFrames_) + " frames (" +
+        std::to_string(totalRenderDuration) + 
+        " seconds)"
+    );
+  }
+
+  int32_t frame = when * sampleRate_;
+  frame = RENDER_QUANTUM_SIZE *
```
maciejmakowski2003, line 74: why???

packages/react-native-audio-api/common/cpp/audioapi/core/OfflineAudioContext.cpp
```diff
@@ -0,0 +1,149 @@
...
+      "cannot schedule more than one suspend at frame " +
+        std::to_string(frame) + " (" + std::to_string(when) +
+        " seconds)"
+    );
+  }
+
+  scheduledSuspends_.emplace(frame, callback);
+}
+
+void OfflineAudioContext::resumeRendering() {
```
maciejmakowski2003, line 101:
```suggestion
void OfflineAudioContext::renderAudio() {
```

packages/react-native-audio-api/common/cpp/audioapi/core/OfflineAudioContext.cpp
maciejmakowski2003: leave a comment about suspension precision, cause it is not obvious for everyone who will read the code

packages/react-native-audio-api/common/cpp/audioapi/core/OfflineAudioContext.cpp
```diff
@@ -0,0 +1,149 @@
...
+  int32_t frame = when * sampleRate_;
+  frame = RENDER_QUANTUM_SIZE *
+          ((frame + RENDER_QUANTUM_SIZE - 1) / RENDER_QUANTUM_SIZE);
+  
+  if (frame < currentSampleFrame_) {
+    int32_t currentFrameClamped = std::min(currentSampleFrame_, numFrames_);
+    double currentTime = currentSampleFrame_ / (double)sampleRate_;
+    double totalTime = numFrames_ / (double)sampleRate_;
+    double currentTimeClamped = std::min(currentTime, totalTime);
+    throw std::runtime_error(
```
maciejmakowski2003, line 82: we do not handle parameter errors in cpp. suggest some way how we can do it on TS side please.

orange4glace:
> I have prepared branch with your code, but updated to be compatible with current API (you made a fork with 0.4.x, currently we working on 0.5.x) and applied some minor fixes and improvements.

Great! I think we can proceed on #389

packages/react-native-audio-api/common/cpp/audioapi/core/OfflineAudioContext.cpp
orange4glace: I'll add comment for that.
In fact, most of the code is referenced from Chromium, including error messages.

packages/react-native-audio-api/common/cpp/audioapi/core/OfflineAudioContext.cpp
orange4glace: I'm concerned that if we can handle it on the JS side.
Since the time input from JS, which is seconds, should be converted to size of render quantum time, which means we need `RENDER_QUANUTM_SIZE` to calculate.
If the error message should be handled in JS side, we should expose `RENDER_QUANTUM_TIME` to JS interface, which is not a part of spec in `OfflineAudioContext`.

packages/react-native-audio-api/common/cpp/audioapi/core/OfflineAudioContext.cpp
maciejmakowski2003: okey

packages/react-native-audio-api/common/cpp/audioapi/core/OfflineAudioContext.cpp
maciejmakowski2003: We can leave it for know, but I will think about better soultion

maciejmakowski2003:
Overall, great work! It is not the easiest repo to make contributions. Congrats!

maciejmakowski2003:
Overall, great work! It is not the easiest repo to make contributions. Congrats!

PR #351, opened 2025-03-12
Feat/modifing data by get channel data

### LINKED ISSUE #348: Enable direct modification of `AudioBuffer` data
### Feature Request

`getChannelData(i)[j] = x` should modify audio data

maciejmakowski2003:
Closes #348

## Introduced changes

- Enabled direct `AudioBuffer's` data modification via `getChannelData` method.

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/jsi/AudioArrayBuffer.h
```diff
@@ -0,0 +1,23 @@
...
+ public:
+  AudioArrayBuffer(float *data, size_t size);
+  ~AudioArrayBuffer() override = default;
+
+  [[nodiscard]] size_t size() const override;
+  uint8_t *data() override;
+
+ private:
+  float *data_;
+  size_t size_;
```
tomekzaw:
```suggestion
  const float *data_;
  const size_t size_;
```

packages/react-native-audio-api/common/cpp/audioapi/jsi/AudioArrayBuffer.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <jsi/jsi.h>
+
+namespace audioapi {
+
+using namespace facebook;
+
+class AudioArrayBuffer : public jsi::MutableBuffer {
+ public:
+  AudioArrayBuffer(float *data, size_t size);
```
tomekzaw:
```suggestion
  AudioArrayBuffer(const float *data, const size_t size);
```

packages/audiodocs/docs/sources/audio-buffer.mdx
```diff
@@ -32,7 +32,7 @@ Once you have data in `AudioBuffer`, audio can be played by passing it to [`Audi
 
 ### `getChannelData`
 
-Above method returns array with PCM data from given channel.
+Above method returns modifiable array with PCM data from given channel.
```
tomekzaw, line 35:
```suggestion
The above method returns a modifiable array with PCM data from given channel.
```

packages/react-native-audio-api/common/cpp/audioapi/jsi/AudioArrayBuffer.h
```diff
@@ -0,0 +1,23 @@
...
+  ~AudioArrayBuffer() override = default;
+
+  [[nodiscard]] size_t size() const override;
```
tomekzaw, line 14: Just to confirm, `size()` returns the size in bytes, is that correct? So if we have an array buffer of length of 10 float numbers, size is 40?

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioBufferHostObject.h
```diff
@@ -52,16 +53,12 @@ class AudioBufferHostObject : public JsiHostObject {
     auto length = static_cast<int>(audioBuffer_->getLength());
     auto size = static_cast<int>(length * sizeof(float));
 
-    auto arrayBufferCtor = runtime.global().getPropertyAsFunction(runtime, "ArrayBuffer");
-    auto arrayBuffer = arrayBufferCtor.callAsConstructor(runtime, size).getObject(runtime).getArrayBuffer(runtime);
+    auto audioArrayBuffer = std::make_shared<AudioArrayBuffer>(channelData, size);
```
tomekzaw, line 58: Just FYI, if you try to access the array buffer from JS when `channelData` is already deallocated, this will crash. The whole point of using shared pointers is that you don't need to worry about such cases. On the other hand, I understand that the lifetime of `channelData` is already controlled externally so this would require a bigger refactor to achieve that. I guess this is okay for now but let's keep that in mind.

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioBufferHostObject.h
```diff
@@ -52,16 +53,12 @@ class AudioBufferHostObject : public JsiHostObject {
...
+    auto arrayBuffer = jsi::ArrayBuffer(runtime, audioArrayBuffer);
```
tomekzaw, line 59: This is exactly what I suggested 😄 Looks good!

packages/react-native-audio-api/common/cpp/audioapi/jsi/AudioArrayBuffer.cpp
```diff
@@ -0,0 +1,16 @@
...
+
+AudioArrayBuffer::AudioArrayBuffer(float *data, size_t size)
+    : data_(data), size_(size) {}
+
+size_t AudioArrayBuffer::size() const {
+  return size_;
+}
+
+uint8_t *AudioArrayBuffer::data() {
+  return reinterpret_cast<uint8_t *>(data_);
```
tomekzaw: Another option would be to use `reinterpret_cast` in constructor (instead of in `data()` method) and store `data_` as `uint8_t *` instead of `float *`. Which option do you think is better?

packages/react-native-audio-api/common/cpp/audioapi/jsi/AudioArrayBuffer.cpp
```diff
@@ -0,0 +1,16 @@
+#include <audioapi/jsi/AudioArrayBuffer.h>
+
+namespace audioapi {
+
+AudioArrayBuffer::AudioArrayBuffer(float *data, size_t size)
+    : data_(data), size_(size) {}
```
tomekzaw: Just FYI, C++ constructors with empty method body, only with argument list, are usually implemented in header file. Similarly, we could inline the implementation of `size()` and `data()` methods by moving them to the header file as well. However, in such case all methods of `AudioArrayBuffer` would be inline so there would be no key function (pure non-line function) and there could be some problems with RTTI across shared library boundary, i.e. if you create `AudioArrayBuffer` in `libfoo.so`, then `dynamic_cast` would not work correctly in `libbar.so`, it would just always return `null`. So perhaps it's better to have `size()` and `data()` in .cpp file.

packages/react-native-audio-api/common/cpp/audioapi/jsi/AudioArrayBuffer.h
maciejmakowski2003: thats right

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioBufferHostObject.h
maciejmakowski2003: I will leave a comment about it, just in case

packages/react-native-audio-api/common/cpp/audioapi/jsi/AudioArrayBuffer.cpp
maciejmakowski2003: I guess it would be more intuitive to store it as `uint8_t`, cause `size` property is number of bytes

packages/react-native-audio-api/common/cpp/audioapi/jsi/AudioArrayBuffer.h
maciejmakowski2003: if type of data_ is `const uint8_t*`, I have to return `const_cast<uint8_t *>(data_);` in `data()` method, cause cv-qualification are different.

tomekzaw (approved):
Looks good to me!

michalsek (approved):

PR #376, opened 2025-03-29
Memory management improvements

michalsek:
Closes #

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Updated relevant documentation
- [x] Added/Conducted relevant tests
- [x] Performed self-review of the code
- [ ] Updated Web Audio API coverage
- [x] Added support for web

PR #378, opened 2025-03-31
Fix/memory manage improvements

maciejmakowski2003:
Closes RNAA-44
Closes RNAA-45

## Introduced changes

- Changes from #376 

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/apps/common-app/src/examples/DrumMachine/constants.ts
+++ b/apps/common-app/src/examples/DrumMachine/constants.ts
@@ -9,7 +9,7 @@ export const padding = 24;
 export const maxSize =
   Math.min(screenSize.width, screenSize.height) - padding * 2;
 
-export const initialBpm = 120;
+export const initialBpm = 240;
 export const numBeats = 16;
 export const noteDensity = 2;
 
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.cpp
@@ -266,8 +266,6 @@ void AudioNode::onInputDisconnected(AudioNode *node) {
 
   if (position != inputNodes_.end()) {
     inputNodes_.erase(position);
-  } else {
-    assert(false);
   }
 }
 
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.cpp
@@ -12,7 +12,9 @@ AudioNode::AudioNode(BaseAudioContext *context) : context_(context) {
 }
 
 AudioNode::~AudioNode() {
-  isInitialized_ = false;
+  if (isInitialized_) {
+    cleanup();
+  }
 }
 
 int AudioNode::getNumberOfInputs() const {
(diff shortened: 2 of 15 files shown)


apps/common-app/src/examples/DrumMachine/constants.ts
```diff
@@ -9,7 +9,7 @@ export const padding = 24;
 export const maxSize =
   Math.min(screenSize.width, screenSize.height) - padding * 2;
 
-export const initialBpm = 120;
```
michalsek, line 12: I think we can stay with 120 bpm :)

packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.cpp
```diff
@@ -239,30 +258,28 @@ void AudioNode::onInputDisconnected(AudioNode *node) {
...
 
-  if (isEnabled()) {
-    node->onInputDisabled();
+  auto position = inputNodes_.find(node);
+
+  if (position != inputNodes_.end()) {
+    inputNodes_.erase(position);
+  } else {
+    assert(false);
   }
```
michalsek:
```suggestion
  if (position != inputNodes_.end()) {
    inputNodes_.erase(position);
  }
```

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferSourceNode.cpp
```diff
@@ -149,19 +149,21 @@ void AudioBufferSourceNode::processNode(
...
-  }
+    if (!pitchCorrection_) {
+      processWithoutPitchCorrection(processingBus, framesToProcess);
+    } else {
+      processWithPitchCorrection(processingBus, framesToProcess);
+    }
 
-  handleStopScheduled();
+    handleStopScheduled();
+  }
```
michalsek, line 163:
```suggestion
  } else {
    processingBus->zero();
  }
```

michalsek: there is also unnecessary if at the beginning of the function (we would like to try to aquire the lock only once here)

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.h
```diff
@@ -30,6 +30,8 @@ class AudioNodeManager {
   void cleanup();
 
  private:
+  friend class AudioNode;
```
michalsek: I think this is unnecessary (used this access during debugging)

michalsek (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.cpp
```diff
@@ -189,21 +204,25 @@ std::shared_ptr<AudioBus> AudioNode::applyChannelCountMode(
...
+  auto position = outputNodes_.find(node);
+
+  if (position == outputNodes_.end()) {
+    outputNodes_.insert(node);
+    node->onInputConnected(this);
+  }
 }
 
 void AudioNode::disconnectNode(const std::shared_ptr<AudioNode> &node) {
   outputNodes_.erase(node);
```
michalsek, line 225: just in case, won't it throw if node does not exist?

michalsek (approved):
LGTM, just one last comment

packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.cpp
maciejmakowski2003: node will always exists in this case, cause we are passing it by `shared_ptr`

PR #399, opened 2025-04-14
Chore/rn audio api plugin

maciejmakowski2003:
Part of RNAA-17

## Introduced changes

- Added expo plugin setting `ForegroundService` permissions on Android and `Background permissions` on iOS

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/src/plugin/withAudioAPI.ts
```diff
@@ -0,0 +1,38 @@
...
+    'android.permission.FOREGROUND_SERVICE',
+    'android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK',
+  ]);
+};
+
+const withAudioAPI: ConfigPlugin = (config) => {
+  config = withBackgroundAudio(config);
+  config = withAndroidPermissions(config);
+  return config;
+};
```
michalsek: I think the entire thing should be customizable on client-app app.json / app.config.(ts/js) level.

f.e.
```
plugins: [
  [
    'react-native-audio-api',
    {
       iosBackgroundMode: true,
       androidForegroundService: true,
       androidFSPermissions: ['PERMISSION_1', 'PERMISSION_2'],
    },
  ],
]
```

alternatively `androidForegroundService` could be a `false` or list of permissions

michalsek (changes requested):

packages/react-native-audio-api/src/plugin/withAudioAPI.ts
```diff
@@ -0,0 +1,38 @@
...
+    }
+    return iosConfig;
+  });
+};
+
+const withAndroidPermissions: ConfigPlugin = (config) => {
+  return AndroidConfig.Permissions.withPermissions(config, [
+    'android.permission.FOREGROUND_SERVICE',
+    'android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK',
+  ]);
```
michalsek: Also I'm not sure if setting this only on the app level manifest will be enough, as I've pointed out those also exists on the foreground service/lib level manifest, which also affects the review process.

maciejmakowski2003: so, what we have to add?

packages/react-native-audio-api/src/plugin/withAudioAPI.ts
maciejmakowski2003: will do

apps/fabric-example/app.json
```diff
@@ -1,4 +1,22 @@
 {
   "name": "FabricExample",
-  "displayName": "FabricExample"
+  "displayName": "FabricExample",
+  "expo": {
+    "plugins": [
+      [ "react-native-audio-api",
```
tomekzaw:
```suggestion
      [
        "react-native-audio-api",
```

maciejmakowski2003: here the way how we will add this plugin

```js
"plugins": [
      [ 
        "react-native-audio-api",
        {
          "iosBackgroundMode": true,
          "androidForegroundService": true,
          "androidFSPermissions" : [
            "android.permission.FOREGROUND_SERVICE",
            "android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK"
          ],
          "androidFSTypes": [
            "mediaPlayback",
            "dataSync"
          ]
        }
      ]
```

packages/react-native-audio-api/src/plugin/withAudioAPI.ts
```diff
@@ -0,0 +1,92 @@
...
+      ];
+    } else {
+      iosConfig.modResults.UIBackgroundModes = ['audio'];
+    }
+    return iosConfig;
+  });
+};
+
+// 'android.permission.FOREGROUND_SERVICE',
+// 'android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK',
```
michalsek: cleanup :)

packages/react-native-audio-api/src/plugin/withAudioAPI.ts
```diff
@@ -0,0 +1,92 @@
...
+  return withInfoPlist(config, (iosConfig) => {
+    if (iosConfig.modResults.UIBackgroundModes) {
+      iosConfig.modResults.UIBackgroundModes = [
+        ...Array.from(
+          new Set([...iosConfig.modResults.UIBackgroundModes, 'audio'])
+        ),
+      ];
+    } else {
+      iosConfig.modResults.UIBackgroundModes = ['audio'];
+    }
```
lukmccall:
```suggestion
    iosConfig.modResults.UIBackgroundModes = [
      ...Array.from(
        new Set([...(iosConfig.modResults.UIBackgroundModes ?? []), 'audio'])
      ),
    ];
```
It should give the same result, but in my option is more readable. It's up to you ;)

packages/react-native-audio-api/src/plugin/withAudioAPI.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    mainApplication.service.push(serviceElement);
+
+    return mod;
+  });
+};
+
+const withAudioAPI: ConfigPlugin<{
+  iosBackgroundMode: boolean;
+  androidForegroundService: boolean;
```
lukmccall:
```suggestion
  iosBackgroundMode?: boolean;
  androidForegroundService?: boolean;
```
These should be optional since you are providing default values for them.

packages/react-native-audio-api/src/plugin/withAudioAPI.ts
```diff
@@ -0,0 +1,92 @@
...
+  androidFSPermissions: string[];
+  androidFSTypes: string[];
```
lukmccall: These should probably be optional or converted to a union type, since if `androidForegroundService` is false, the value is likely undefined.

lukmccall (approved):

packages/react-native-audio-api/src/plugin/withAudioAPI.ts
maciejmakowski2003: will do

michalsek (approved):

PR #405, opened 2025-04-21
fix: create thread on constructor, after others are initialized

orange4glace:
## Introduced changes

- `AudioNodeDestructor::thread_` is initialized on initialization list, while its thread captures `this` itself.
I've experienced app crash on iOS, at line `cv_.wait(lock, [this] {`

```cpp
void AudioNodeDestructor::process() {
  std::unique_lock<std::mutex> lock(mutex_);
  while (!isExiting_) {
    cv_.wait(lock, [this] {
      return isExiting_ || !nodesForDeconstruction_.empty();
    });

    if (isExiting_)
      break;

    if (!isExiting_ && !nodesForDeconstruction_.empty()) {
      nodesForDeconstruction_.clear();
    }
  }
}
```

with error 
```
libc++abi: terminating due to uncaught exception of type std::__1::system_error: condition_variable wait failed: Invalid argument
```

It seems like a thread trying to access some uninitialized value of `this` .
After moving code to initialize thread on its constructor, no more crash happened.

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeDestructor.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeDestructor.cpp
@@ -4,9 +4,9 @@
 
 namespace audioapi {
 
-AudioNodeDestructor::AudioNodeDestructor()
-    : thread_(std::thread(&AudioNodeDestructor::process, this)),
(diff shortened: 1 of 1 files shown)


maciejmakowski2003 (approved):
Great job!

PR #410, opened 2025-04-23
fix: remove playing nodes from destruction

### LINKED ISSUE #400: `AudioNode` is garbage collected even it is connected with another node.
### Description

Though it is hard to suggest a code to reproduce,
I've experienced `AudioBufferSourceNode` is garbage collected and stop playing (by being `~AudioNode()` is called)

For example,

```ts
const context = new AudioContext();
{
  const sourceNode = context.createBufferSourceNode();
  sourceNode.buffer = myBuffer;
  sourceNode.connect(context.destination);
  sourceNode.start();
}
```

In this code, `sourceNode` should not be garbage collected since it is connected with `AudioContext.destination`, but `sourceNode` can be garbage collected thus `sourceNode` is cleaned up, finally disconnected from `AudioContext`.

### Steps to reproduce

_

_

4ccd750

### React Native version

0.75.3

### Platforms

iOS, Android

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

mdydek:
Closes #400 
Closes RNAA-62

## Introduced changes

- added checks during destruction in order not to destroy playing source nodes

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/sources/OscillatorNode.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/sources/OscillatorNode.cpp
@@ -51,7 +51,7 @@ void OscillatorNode::processNode(
 
   updatePlaybackInfo(processingBus, framesToProcess, startOffset, offsetLength);
 
-  if (!isPlaying()) {
+  if (!isPlaying() && !isStopScheduled()) {
     processingBus->zero();
     return;
   }
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.h
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.h
@@ -11,6 +11,7 @@
 namespace audioapi {
 
 class AudioNode;
+class AudioScheduledSourceNode;
 
 class AudioNodeManager {
  public:
@@ -27,7 +28,8 @@ class AudioNodeManager {
       const std::shared_ptr<AudioNode> &to,
       ConnectionType type);
 
-  void addNode(const std::shared_ptr<AudioNode> &node);
+  void addProcessingNode(const std::shared_ptr<AudioNode> &node);
+  void addSourceNode(const std::shared_ptr<AudioScheduledSourceNode> &node);
 
   void cleanup();
 
@@ -36,7 +38,8 @@ class AudioNodeManager {
   AudioNodeDestructor nodeDeconstructor_;
 
   // all nodes created in the context
-  std::unordered_set<std::shared_ptr<AudioNode>> nodes_;
+  std::unordered_set<std::shared_ptr<AudioScheduledSourceNode>> sourceNodes_;
+  std::unordered_set<std::shared_ptr<AudioNode>> processingNodes_;
 
   // connections to be settled
   std::vector<std::tuple<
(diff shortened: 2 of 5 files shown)


maciejmakowski2003 (approved):
Great work! Before merge please check if it works well on drums example with bpm set to 1000.

michalsek (approved):
Good job! 🔥

mdydek:
Unfortunately previous approach using dynamic cast turned out to be not efficient enough in stressful environment.

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.h
```diff
@@ -46,6 +49,7 @@ class AudioNodeManager {
       audioNodesToConnect_;
 
   void settlePendingConnections();
+  void cleanupNode(const std::shared_ptr<AudioNode> &node);
```
maciejmakowski2003, line 52: Where it is used?

mdydek: In AudioNodeManager.cpp to not repeat lines: 
  nodeDeconstructor_.addNodeForDeconstruction(node);
  node.get()->cleanup();

maciejmakowski2003 (approved):

packages/react-native-audio-api/common/cpp/audioapi/core/sources/OscillatorNode.cpp
```diff
@@ -51,7 +51,7 @@ void OscillatorNode::processNode(
 
   updatePlaybackInfo(processingBus, framesToProcess, startOffset, offsetLength);
 
-  if (!isPlaying()) {
+  if (!isPlaying() && !isStopScheduled()) {
```
michalsek, line 54: wont it cause the oscillator to immediately stop playing once it is scheduled?

mdydek: from my understanding `stopScheduled` is a state where the time of finishing playing is set but it is yet to come, so practically `stopScheduled` state is also kind of playing state @maciejmakowski2003

michalsek: oh wait, didn't notice the `!`

buut I've followup question why this condition is needed? I can understand it as
"is not playing and stop is not scheduled" while I don't see a reason why it would be in the state where it is not playing and stop is scheduled 🤔

mdydek: It IS playing in both of states, but in plain playing state we don’t know the time of ending. It was introduced by @maciejmakowski2003, because he needed differentiation in some of his features.

michalsek:
Good to go! :)

packages/react-native-audio-api/common/cpp/audioapi/core/sources/OscillatorNode.cpp
maciejmakowski2003: Exactly like @mdydek explained. I have introduced new state of ASSN while refactoring frame accurate stop. We need to know if node is playing and stop is scheduled to happen or it is not.

PR #413, opened 2025-04-24
Fix/audio-engine

maciejmakowski2003:
Closes RNAA-75
Part of RNAA-72

## Introduced changes

- Changed initial `iOS` options in order to prevent `AVAudioEngine` crash when playing before using microphone
- Switch from tap to `AVAudioSinkNode`
- Refactor start/stop and resume/pause logic in `AVAudioEngine` and node management
- Replaced `AudioRecorder` interop class with interface which is implemented by `AndroidAudioRecorder` and `iOSAudioRecorder`

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web


mdydek (changes requested):
I think the problem needs further investigation. Some apps are not working as intended when their screen is run first. The situation is gone when user goes back and reopen this or another screen.

maciejmakowski2003:
Thats right, this PR only fixes initial settings, but still we are facing bug mentioned by you.

mdydek:
Thought it resolves everything, but nevertheless I think that this is a good place to fix them.

michalsek (approved):

packages/react-native-audio-api/ios/audioapi/ios/system/AudioSessionManager.mm
```diff
@@ -24,9 +24,10 @@ - (instancetype)initPrivate
   if (self = [super init]) {
     self.audioSession = [AVAudioSession sharedInstance];
 
-    self.sessionCategory = AVAudioSessionCategoryPlayback;
+    self.sessionCategory = AVAudioSessionCategoryPlayAndRecord;
     self.sessionMode = AVAudioSessionModeDefault;
-    self.sessionOptions = AVAudioSessionCategoryOptionDuckOthers | AVAudioSessionCategoryOptionAllowBluetooth;
+    self.sessionOptions = AVAudioSessionCategoryOptionAllowAirPlay | AVAudioSessionCategoryOptionAllowBluetoothA2DP |
+        AVAudioSessionCategoryOptionDefaultToSpeaker;
```
michalsek: I think we should stay with 

AVAudioSessionCategoryPlayback
AVAudioSessionModeDefault
AVAudioSessionCategoryOptionAllowAirPlay

as default, and only change that for microphone example

maciejmakowski2003: will do

michalsek: Reason:
- we want to stay as close as to default configuration, in our default settings (do not set something lib user doesn't expect)
- libs main purpouse is playback not p&r

PR #414, opened 2025-04-25
feat: useSystemVolume hook for android

mdydek:
Closes RNAA-55

## Introduced changes

- Added hook that allows user track volume value in android

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/AudioManagerModule.kt
+++ b/packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/AudioManagerModule.kt
@@ -57,5 +57,8 @@ class AudioManagerModule(
   @ReactMethod(isBlockingSynchronousMethod = true)
   fun observeAudioInterruptions(enable: Boolean) = MediaSessionManager.observeAudioInterruptions(enable)
 
+  @ReactMethod(isBlockingSynchronousMethod = true)
+  fun observeVolumeChanges(enable: Boolean) = MediaSessionManager.observeVolumeChanges(enable)
+
   override fun getName(): String = NAME
 }
--- a/packages/react-native-audio-api/src/system/AudioManager.ts
+++ b/packages/react-native-audio-api/src/system/AudioManager.ts
@@ -3,6 +3,7 @@ import {
   LockScreenInfo,
   RemoteEventName,
   RemoteEventCallback,
+  RemoteCommandName,
 } from './types';
 import { AudioManagerModule, eventEmitter } from '../specs';
 import { EmitterSubscription } from 'react-native';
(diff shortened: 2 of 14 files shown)


maciejmakowski2003 (approved):
LGTM!

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/AudioManagerModule.kt
```diff
@@ -57,5 +57,8 @@ class AudioManagerModule(
   @ReactMethod(isBlockingSynchronousMethod = true)
   fun observeAudioInterruptions(enable: Boolean) = MediaSessionManager.observeAudioInterruptions(enable)
 
+  @ReactMethod(isBlockingSynchronousMethod = true)
+  fun observeVolumeChanges(enable: Boolean) = MediaSessionManager.observeVolumeChanges(enable)
```
michalsek, line 61: Do we need blocking synchronous method here? (Same goes to the `observeAudioInterruptions`

packages/react-native-audio-api/src/system/AudioManager.ts
```diff
@@ -42,10 +32,14 @@ class AudioManager {
...
+  observeVolumeChanges(enabled: boolean) {
+    AudioManagerModule.observeVolumeChanges(enabled);
+  }
+
+  enableRemoteCommand<Name extends RemoteEventName>(
+    name: Name,
     enabled: boolean,
-    callback?: (event: AudioManagerEventType) => void
+    callback?: RemoteEventCallback<Name>
   ): EmitterSubscription | null {
```
michalsek: I would split this function into two: f.e. `enableRemoteCommand` and `enableRemoteEvent` as it the `remote-command` might be misleading with events like interruption or volume change.

If we wan't to stay with only one method for adding event handlers I would go with `addEventListener`.

Also It would be great to get rid of the `enabled` parameter

i.e.

```
useEffect(() => {
  const listener = AudioManager.addEventListener('onVolumeChange', (volume) => {
    // do stuff
  });
  
  return () => {
    listener.remove();
  }
}, [deps]);
```

packages/react-native-audio-api/src/system/useSystemVolume.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from 'react';
```
michalsek: Can you move this file to `src/hooks`?

michalsek (changes requested):

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/AudioManagerModule.kt
mdydek: In my implementation you are probably right, but I'm not sure in case of `observeAudioInterruptions`, which was @maciejmakowski2003 implementation.

maciejmakowski2003: It is temporary. This will be changed with PR introducing module refactoring. We are going to have one TurboModule(interface will be generated with Codegen) and own implementation of AudioEventEmitter like in Reanimated. In this case, method does not have to be blocking synchronous

michalsek (approved):

PR #415, opened 2025-04-25
feat: useSystemVolume hook for ios

mdydek:
Closes RNAA-54

## Introduced changes

- Added hook that allows user track volume value in ios

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/ios/audioapi/ios/system/NotificationManager.mm
```diff
@@ -59,6 +60,34 @@ - (void)observeAudioInterruptions:(BOOL)enabled
   self.audioInterruptionsObserved = enabled;
 }
 
+- (void)observeVolumeChanges:(BOOL)enabled
```
michalsek, line 64: Lets also add info about this not working in sim in a comment here :)

maciejmakowski2003 (approved):

PR #430, opened 2025-05-05
feat: first implementation of new audio param

mdydek:
Closes RNAA-79

- Refactored Audio Param in order to prepare architecture for possibility of connection Nodes to Params.

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/core/v2/AudioNode.h
```diff
@@ -0,0 +1,84 @@
...
+
+  int getNumberOfInputs() const;
+  int getNumberOfOutputs() const;
+  int getChannelCount() const;
+  std::string getChannelCountMode() const;
+  std::string getChannelInterpretation() const;
+  void connect(const std::shared_ptr<AudioNode> &node);
+  void disconnect();
+  void disconnect(const std::shared_ptr<AudioNode> &node);
+  virtual std::shared_ptr<AudioBus> processAudio(const std::shared_ptr<AudioBus> &outputBus, int framesToProcess, bool checkIsAlreadyProcessed);
```
mdydek: for now it is only change in node compared to v1, I need to have an access to that function in param so I moved it to the public scope

maciejmakowski2003:
so far looks good

packages/react-native-audio-api/common/cpp/audioapi/core/AudioParam.cpp
```diff
@@ -256,6 +267,48 @@ void AudioParam::cancelAndHoldAtTime(double cancelTime) {
...
+  auto processingBus = audioBus_;
+  processingBus->zero();
+  if (inputNodes_.empty()) {
+    processParamNoInput(processingBus, framesToProcess, time, sampleRate);
+  } else {
+    processInputs(processingBus, framesToProcess, true);
+    mixInputsBuses(processingBus);
+  }
+  // processingBus is a mono bus
+  return std::make_shared<AudioArray>(processingBus->getChannel(0));
```
michalsek, line 309: One of the main trobules when working with the audio thread, is that we are not allowed to allocate data on it (single primitive values like int, float, bool are ok though).

As it might result in crackling audio.

michalsek: I would presonally go for returning `std::shared_ptr<AudioBus>` from this function

michalsek:
```
std::shared_ptr<AudioBius> AudioParam::processParam(int framesToProcess, double time, float sampleRate) {
  audioBus_->zero();

  if (inputNodes_.empty()) {
    processParamNoInput(audioBus_, framesToProcess, time, sampleRate);
  } else {
    processInputs(audioBus_, framesToProcess, true);
    mixInputsBuses(audioBus_);
  }

  return audioBus_;
}
```

packages/react-native-audio-api/common/cpp/audioapi/core/AudioParam.cpp
```diff
@@ -256,6 +267,48 @@ void AudioParam::cancelAndHoldAtTime(double cancelTime) {
...
+  for (size_t i = 0; i < framesToProcess; i++) {
+    auto sample = getValueAtTime(
+        dsp::timeToSampleFrame(time, sampleRate) + i / sampleRate);
+    // outputBus is a mono bus
+    (*outputBus->getChannel(0))[i] = sample;
+  }
+}
+
+std::shared_ptr<AudioArray>
+AudioParam::processParam(int framesToProcess, double time, float sampleRate) {
```
michalsek, line 299: Do we really need sampleRate here?

packages/react-native-audio-api/common/cpp/audioapi/core/AudioParam.cpp
mdydek: great catch, resolved in #432, this will be main pr for this feature

packages/react-native-audio-api/common/cpp/audioapi/core/AudioParam.cpp
mdydek: You're right, we can use sampleRate from context, which is already defined in param

mdydek: fixed in #432 :)

PR #453, opened 2025-05-19
Feat/onended/added state to payload

maciejmakowski2003:
Closes #

## Introduced changes

- Added `state` prop to the payload of `onended` event. It will allow us to distinguish if `ABSN` has been stopped or it has ended by itself

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
@@ -152,9 +152,7 @@ void AudioScheduledSourceNode::disable() {
   }
 
   std::unordered_map<std::string, EventValue> body = {
-          {"value", getStopTime()},
-          {"state", state}
-  };
+      {"value", getStopTime()}, {"state", state}};
 
   context_->audioEventHandlerRegistry_->invokeHandlerWithEventBody(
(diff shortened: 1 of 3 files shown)


packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
```diff
@@ -144,7 +144,15 @@ void AudioScheduledSourceNode::updatePlaybackInfo(
 void AudioScheduledSourceNode::disable() {
   AudioNode::disable();
 
-  std::unordered_map<std::string, EventValue> body = {{"value", getStopTime()}};
+  std::string state = "stopped";
+
+  // if it has not been stopped, it is ended
+  if (stopTime_ < 0) {
```
michalsek, line 150: what if stopTime_ == startTime_ + buffer.duration? :)

michalsek (approved):
One nit, besides that looks good :)

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
maciejmakowski2003: Still it is finished by calling stop. However we can distinguish that case in the callback body.

PR #457, opened 2025-05-20
fix: added mutex in param for thread safety

mdydek:
Closes RNAA-56

## Introduced changes

- added mutex and logic behind locking functionalities that use eventsQueue, in order to not allow race condition between JS and audio thread

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


mdydek:
We are currently investigating possible options to achieve this, since approach introduced in this PR lowers efficiency of library ~10%.

PR #466, opened 2025-05-26
Refactor/system settings and audio engine

maciejmakowski2003:
Closes RNAA-100
Closes RNAA-105

## Introduced changes

- Refactor resume/suspend (start/stop) audio player logic
- Fixed some audio system settings on iOS
- Refactored `enableSystemEvent` -> introduced `enableRemoteCommand` and `addSystemEventListener` for all system events
- Implemented `setAudioSessionActivity`

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/src/events/types.ts
+++ b/packages/react-native-audio-api/src/events/types.ts
@@ -34,10 +34,10 @@ interface RemoteCommandEvents {
   remoteChangePlaybackRate: EventTypeWithValue;
   remoteNextTrack: EventEmptyType;
   remotePreviousTrack: EventEmptyType;
-  remoteSkipForward: EventEmptyType;
-  remoteSkipBackward: EventEmptyType;
-  remoteSeekForward: EventTypeWithValue;
-  remoteSeekBackward: EventTypeWithValue;
+  remoteSkipForward: EventTypeWithValue;
+  remoteSkipBackward: EventTypeWithValue;
+  remoteSeekForward: EventEmptyType;
+  remoteSeekBackward: EventEmptyType;
   remoteChangePlaybackPosition: EventTypeWithValue;
 }
 


michalsek (approved):

packages/react-native-audio-api/src/events/types.ts
```diff
@@ -26,23 +26,26 @@ interface OnRouteChangeEventType {
...
   remoteNextTrack: EventEmptyType;
   remotePreviousTrack: EventEmptyType;
-  remoteSkipForward: EventEmptyType;
-  remoteSkipBackward: EventEmptyType;
-  remoteSeekForward: EventTypeWithValue;
-  remoteSeekBackward: EventTypeWithValue;
+  remoteSkipForward: EventTypeWithValue;
+  remoteSkipBackward: EventTypeWithValue;
+  remoteSeekForward: EventEmptyType;
+  remoteSeekBackward: EventEmptyType;
```
mdydek, line 40: Is there a reason why they switch return types?

packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.cpp
mdydek, line 1: So we are basically deleting functionalities from  audioPlayers that were responsible for suspending and resuming player. But instead, on invoking resume and suspend on audioContext we are calling start and stop respectively. What is the use case here?

maciejmakowski2003: That approach was incorrect due to system constraints, so the current one provide better reliability

packages/react-native-audio-api/src/events/types.ts
maciejmakowski2003: We had a typo and current types are correct

PR #469, opened 2025-05-27
Enable Native External Audio Processing via TurboModules

jerryseigle:
This PR introduces a clean and modular system for injecting native audio processing logic into react-native-audio-api through the use of ExternalAudioProcessor and a global registry.

Developers can now:
	•	Register and unregister external processors from a TurboModule at runtime
	•	Write custom DSP (digital signal processing) code in C++
	•	Perform advanced buffer-level audio manipulation directly within the native render cycle
	•	Avoid any reliance on the JS thread — all processing runs fully native

This allows integration of virtually any kind of audio processing, whether:
	•	Custom-written code tailored to your app’s needs
	•	Third-party DSP libraries (e.g., for pitch/tempo manipulation, watermarking, audio detection, etc.)

All without modifying the core AudioNode logic — keeping the system clean, flexible, and decoupled.

✅ Checklist
	•	Added ExternalAudioProcessor interface
	•	Implemented singleton processor registry
	•	Injected processing safely into AudioNode::processAudio
	•	Provided TurboModule demo for runtime control (volume reducer)
	•	Documented and isolated logic for easy integration


jerryseigle:
I’m not sure where to include a working example, so I’ve attached a few sample files here. If approved, I’ll also add proper documentation. You can test the implementation using these files, and there’s also a demo video available to showcase it in action.
[shared.zip](https://github.com/user-attachments/files/20449060/shared.zip)

https://github.com/user-attachments/assets/adcb7325-1ed2-4c3a-8fb8-76a2d500df8c

jerryseigle:
> Hey, not sure where to respond, so will write everything here :)
> 
> Yes, each node is a separate AudioNode, which means that in order to do such external processing node and be able to hook it up anywhere in the graph, I think we have to iterate it over a bit :)
> 
> Instead of modifying the AudioNode directly, I would go for new separate node, that implements this, f.e. `CustomAudioNode`, it would be inheriting from `AudioNode` and overwrite the `processAudio` or `processNode` method.
> 
> Which could be then exposed to JS/RN as a new node type, that we can connect to, f.e.
> 
> ```ts
> const absn1 = audioContext.createBufferSource();
> const absn2 = audioContext.createBufferSource();
> const externalProcessor = audioContext.createCustomProcessor();
> 
> absn1.connect(externalProcessor);
> absn2.connect(externalProcessor);
> 
> externalProcessor.connect(audioContext.destination);
> ```
> 
> Regarding your question about timing, if we have:
> 
> ```ts
> absn1.start(now + 0.01);
> absn2.start(now + 0.01);
> ```
> 
> both nodes will start exactly at the same time (or at the exact sample frame to be precise), but as I stated above, they do not share the same AudioNode instance. It is a bit more complex and I will be happy to explain how the pull-graph works within audio api or web audio, if you're interested :)
> 
> But for this case, it might be not necessary.
> 
> But overall great job! 🔥 I've imagined that external c++ integrations would be much harder to implement :)

Just so I understand shouldn't be modeled similar to the GainNode and placed in the effect directory?

jerryseigle:
> > Just so I understand so am I to model it similar to the GainNode and placed in the effect directory?
> 
> Yes, exactly, if you are willing to wait, I will prep all the required interfaces for you over the weekend :)

Yes, I can wait if you prep the required interfaces. Thanks!

jerryseigle:
Hey! Hope you had a good weekend 🙂 Just checking in to see if you had a
chance to put together the interfaces. No rush at all—just excited to keep
moving forward when you’re ready. Thanks again!

michalsek:
Hey, hey, 

unfortunately hadn't a chance to look at it, will figure out something along the week :)

michalsek:
Hey, hey how it is going with the PR? :)

jerryseigle:
> Hey, hey how it is going with the PR? :)

Hey, it’s progressing well. I anticipate completing it by Tuesday. I’m currently testing and making a few adjustments.

jerryseigle:
@michalsek  PR is complete. Added support for CustomProcessorNode via TurboModule. Ready for review.

michalsek (approved):

packages/react-native-audio-api/src/core/CustomProcessorNode.ts
```diff
@@ -0,0 +1,36 @@
...
+import AudioNode from "./AudioNode";
+import AudioParam from "./AudioParam";
+import BaseAudioContext from "./BaseAudioContext";
+
+export default class CustomProcessorNode extends AudioNode {
+  readonly customProcessor: AudioParam;
+
+  constructor(
+    context: BaseAudioContext,
+    customProcessor: ICustomProcessorNode
```
mdydek, line 12: Is `identifier` necessary in each instance of `CustomProcessorNode`? If yes, why we enable user to create node without it, if not can you please explain how are you handling it?

jerryseigle: Yes, the identifier is necessary for each CustomProcessorNode, as it’s used to link the node to its matching processor and control handlers.

Would you prefer that I enforce this on the TypeScript side—either by requiring the identifier at creation like:
```
const processor = audioContext.createCustomProcessor("my-id");
```
or by continuing to allow:
```
const processor = audioContext.createCustomProcessor();
processor.identifier = "my-id";
```

Would you prefer that I enforce this on the TypeScript side (e.g., by requiring the identifier during creation or throwing if it’s missing), or would you like to handle that change? I’m happy to implement it if needed.

michalsek: We thought about just passing it in createNode function / constructor :)

jerryseigle: Yes, that makes perfect sense! Would you like me to handle that change and update the API to accept the identifier during creation, or will you be taking care of it?

michalsek: Depends entirely on you :) 

We can merge it at the current stage and interate over later or if you can still bear our whining you can update the PR :)

jerryseigle: Sounds good! I’ll go ahead and make the update should have it done in a few hours. 👍🏾

jerryseigle: Everything is now complete on my end.

- The `identifier` is no longer optional and must be passed into the `createNode` function/constructor.
```ts
audioContext.createCustomProcessor("abc");
```

I also applied Copilot recommended improvements
- Replaced `processorMode` strings with an enum for improved type safety
- Preallocated the buffer in `processThrough` to avoid heap allocations in the audio thread.

All functionality has been tested and is working as expected.
**Ready for your final review.**

vikalp-mightybyte:
@michalsek @maciejmakowski2003 @jerryseigle 
Can we please also update documentation to show how it? 
and, When will this be released, expectedly?

michalsek:
@vikalp-mightybyte definitely will add it to docs! :)

PR #474, opened 2025-05-28
fix: minor fixes

maciejmakowski2003:
Closes #

## Introduced changes

- Fixed small issues connected with lock screen on Android

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/LockScreenManager.kt
```diff
@@ -144,7 +144,7 @@ class LockScreenManager(
 
     elapsedTime =
       if (info.hasKey("elapsedTime")) {
-        info.getDouble("elapsedTime").toLong()
+        (info.getDouble("elapsedTime") * 1000).toLong()
```
mdydek, line 147: so now we always pass seconds in `LockScreenInfo`?

michalsek (approved):

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/LockScreenManager.kt
maciejmakowski2003: yes, just like on iOS

PR #480, opened 2025-05-30
Refactor/system audio settings and engines

maciejmakowski2003:
Closes #

## Introduced changes

- Added `@Synchronized` decorator to `AudioAPIModule` methods on Android
- Added new option `initSuspended` to `AudioContextOptions`, it allows user to create `AudioContext` that is `Suspended` -> `AVAudioEngine` has not started yet, better control over lock screen
- Refactor `AudioPlayer` and engine state management:
  - on iOS (similar on Android)
    1. If AC is init with `{ initSuspended: true }` then `AVAudioEngine` will not be started yet
    2. else AVAudioEngine will be started
    3. `audioContext.resume()` calls `start` -> starts `AVAudioEngine` and install `NativeAudioPlayer` if it was not started 
      or just starts/resume engine
    4. `audioContext.suspend()` pause `AVAudioEngine` instead of stopping -> does not release engine's resources
    5. `AVAudioSession` is activated in the first engine's start invocation
    6. `audioContext.close()` - > engine is stopped (releases resources)
- Refactored `AudioContext's` promises
  - `resume()` and `suspend()` resolves promise when context was `suspended` or `running` and rejects when context was `closed`
- Small fixes

## To consider
- [ ] Scheduling suspend after finishing processing `RENDER_QUANTUM`

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/AudioAPIModule.kt
+++ b/packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/AudioAPIModule.kt
@@ -53,10 +53,8 @@ class AudioAPIModule(
     return true
   }
 
-  @Synchronized
   override fun getDevicePreferredSampleRate(): Double = MediaSessionManager.getDevicePreferredSampleRate()
 
-  @Synchronized
   override fun setAudioSessionActivity(
     enabled: Boolean,
     promise: Promise?,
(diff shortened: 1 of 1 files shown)


packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/AudioAPIModule.kt
```diff
@@ -53,15 +53,18 @@ class AudioAPIModule(
     return true
   }
 
+  @Synchronized
```
michalsek: why? 👀

PR #488, opened 2025-06-04
fix: remove explicit base class destructor calls in AudioRecorder subclasses

tiramisuuuu:
Closes #

## Introduced changes

- Removed explicit calls to the base class destructor (AudioRecorder::~AudioRecorder()) in both IOSAudioRecorder and AndroidAudioRecorder destructors.
- This resolves a double destruction bug that caused crashes on iOS (and potentially Android) when reloading or backgrounding the app (in my development build using expo)
- The destructors now rely on C++'s automatic destruction order, ensuring safe and correct cleanup of resources.

- Disclaimer: I am not a C++ expert so I had to read myself in a bit. As far as I understood it is not necessary to call the base destructors ([Link](https://stackoverflow.com/questions/677620/do-i-need-to-explicitly-call-the-base-virtual-destructor))

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
@@ -26,8 +26,6 @@ AndroidAudioRecorder::AndroidAudioRecorder(
 }
 
 AndroidAudioRecorder::~AndroidAudioRecorder() {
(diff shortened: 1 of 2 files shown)


tiramisuuuu:
I was getting runtime exceptions in my development build due to the double destruction bug in AudioRecorder. This fix resolves it by removing explicit base class destructor calls. I am not a C++ expert, so feel free to correct me on this.

PR #495, opened 2025-06-09
feat: new type proposition

mdydek:
Closes #

## Introduced changes

- proposition of slight changes in settings types

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/src/specs/NativeAudioAPIModule.ts
+++ b/packages/react-native-audio-api/src/specs/NativeAudioAPIModule.ts
@@ -18,10 +18,16 @@ interface Spec extends TurboModule {
   ): void;
   getDevicePreferredSampleRate(): number;
   observeAudioInterruptions(enabled: boolean): void;
-  requestAudioFocus(
+  setAudioOptions(
     observeAudioInterruptions: boolean,
     options?: {
-      [key: string]: string | boolean | number | AudioAttributeType | undefined;
+      [key: string]:
+        | string
+        | string[]
+        | boolean
+        | number
+        | AudioAttributeType
+        | undefined;
     }
   ): void;
   abandonAudioFocus(): void;
--- a/packages/react-native-audio-api/src/system/AudioManager.ts
+++ b/packages/react-native-audio-api/src/system/AudioManager.ts
@@ -35,6 +35,13 @@ class AudioManager {
     NativeAudioAPIModule!.observeAudioInterruptions(enabled);
   }
 
+  /**
+   * Used to set various system options for android and iOS
+   *
+   * @param observeAudioInterruption - If true, the module will observe audio
+   *   interruptions and emit events accordingly (true by default).
+   * @param options - Additional audio options to configure the audio session.
+   */
   setAudioOptions(observeAudioInterruption = true, options?: AudioOptions) {
     NativeAudioAPIModule!.setAudioOptions(observeAudioInterruption, options);
   }
--- a/packages/react-native-audio-api/src/system/AudioManager.ts
+++ b/packages/react-native-audio-api/src/system/AudioManager.ts
@@ -1,9 +1,4 @@
-import {
-  SessionOptions,
-  LockScreenInfo,
-  PermissionStatus,
-  AudioFocusOptions,
-} from './types';
+import { AudioOptions, LockScreenInfo, PermissionStatus } from './types';
 import { SystemEventName, SystemEventCallback } from '../events/types';
 import { NativeAudioAPIModule } from '../specs';
 import { AudioEventEmitter, AudioEventSubscription } from '../events';
@@ -32,14 +27,6 @@ class AudioManager {
     NativeAudioAPIModule!.resetLockScreenInfo();
   }
 
-  setAudioSessionOptions(options: SessionOptions) {
-    NativeAudioAPIModule!.setAudioSessionOptions(
-      options.iosCategory ?? '',
-      options.iosMode ?? '',
-      options.iosOptions ?? []
-    );
-  }
-
   getDevicePreferredSampleRate(): number {
     return NativeAudioAPIModule!.getDevicePreferredSampleRate();
   }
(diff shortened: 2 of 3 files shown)


packages/react-native-audio-api/src/specs/NativeAudioAPIModule.ts
```diff
@@ -18,10 +18,16 @@ interface Spec extends TurboModule {
...
     observeAudioInterruptions: boolean,
     options?: {
-      [key: string]: string | boolean | number | AudioAttributeType | undefined;
+      [key: string]:
+        | string
+        | string[]
+        | boolean
+        | number
+        | AudioAttributeType
+        | undefined;
```
michalsek, line 30: `[key: string]` is a no go (but i guess it is a temp solution)

also `| string | string[] | boolean | number` is a problematic 
to put them together, you can match any option name with any option value.

packages/react-native-audio-api/src/system/types.ts
```diff
@@ -99,13 +93,22 @@ export interface LockScreenInfo extends BaseLockScreenInfo {
...
 }
 
-export interface AudioFocusOptions extends BaseAudioFocusOptions {
-  acceptsDelayedFocusGain?: boolean;
-  audioAttributes?: AudioAttributeType;
-  focusGain?: focusGainType;
-  pauseWhenDucked?: boolean;
+export interface AudioOptions extends BaseAudioOptions {
+  androidAcceptsDelayedFocusGain?: boolean;
+  androidAudioAttributes?: AudioAttributeType;
```
michalsek, line 107: shouldn't this be an array?

packages/react-native-audio-api/src/system/types.ts
```diff
@@ -99,13 +93,22 @@ export interface LockScreenInfo extends BaseLockScreenInfo {
...
+  androidFocusGain?: focusGainType;
```
michalsek, line 108: can't comment on the original type: we should have only one case-type when it comes to setting options, currently:
- ios settings are camel case
- android settings are snake case
- permission status (which is kind of related) is upper-case

michalsek: Also i'm wondering if we should include spatialization behaviour currently, it seems it is either `auto` or `off`. I think we could stick with whatever is default for now. 

Especially ~some-day~ we will have our on spatialization implementation thus we would like to have this off by defULT

michalsek: As we talked a bit today about this: could you add short JSDoc-like descriptions to the values (aka what is their meaning). Could be a great start for documenting the api  on code-level :)

michalsek (changes requested):

packages/react-native-audio-api/src/system/types.ts
mdydek: no,
```
export type AudioAttributeType = {
  allowedCapturePolicy?:
    | 'allow_capture_by_all'
    | 'allow_capture_by_system'
    | 'allow_capture_by_none';
  contentType?: audioAttributeContentType;
  flag?: 'flag_hw_av_sync' | 'flag_audibility_enforced';
  hapticChannelsMuted?: boolean;
  isContentSpatialized?: boolean;
  spatializationBehavior?:
    | 'spatialization_behavior_auto'
    | 'spatialization_behavior_never';
  usage?: audioAttributeUsageType;
};
```
this is a type definition if it, attributes refers to many options in it, but for sure not an array

packages/react-native-audio-api/src/system/types.ts
mdydek: with case-type I wanted to stick with original system values, therefore it is different for ios and android, but it can be narrowed to one easily

packages/react-native-audio-api/src/system/AudioManager.ts
```diff
@@ -48,11 +35,15 @@ class AudioManager {
...
-  ) {
-    NativeAudioAPIModule!.requestAudioFocus(observeAudioInterruption, options);
+  /**
+   * Used to set various system options for android and iOS
+   *
+   * @param observeAudioInterruption - If true, the module will observe audio
+   *   interruptions and emit events accordingly (true by default).
+   * @param options - Additional audio options to configure the audio session.
+   */
+  setAudioOptions(observeAudioInterruption = true, options?: AudioOptions) {
```
maciejmakowski2003, line 45: I think we have to call observe interruption every play invocation on Android so merging AudioSession settings into one method its not the best option in my opinion

packages/react-native-audio-api/src/system/types.ts
```diff
@@ -18,68 +20,59 @@ export type IOSMode =
...
+  | 'usageAssistanceAccessibility'
+  | 'usageAssistanceNavigationGuidance'
+  | 'usageAssistanceSonification'
+  | 'usageAssistant'
+  | 'usageGame'
+  | 'usageMedia'
+  | 'usageNotification'
+  | 'usageNotificationEvent'
+  | 'usageNotificationRingtone'
+  | 'usageNotificationCommunicationRequest'
```
maciejmakowski2003, line 44: do we need all of these usage types?

packages/react-native-audio-api/src/system/types.ts
```diff
@@ -18,68 +20,59 @@ export type IOSMode =
...
 export type audioAttributeContentType =
-  | 'content_type_movie'
-  | 'content_type_music'
-  | 'content_type_sonification'
-  | 'content_type_speech'
-  | 'content_type_unknown';
+  | 'contentTypeMovie'
+  | 'contentTypeMusic'
+  | 'contentTypeSonification'
+  | 'contentTypeSpeech'
```
maciejmakowski2003, line 56: same here

maciejmakowski2003 (changes requested):

packages/react-native-audio-api/src/system/AudioManager.ts
mdydek: so am I correct, if the audio interruption event will be fired, it won't be invoked 2nd time and we have to re-observe it?

packages/react-native-audio-api/src/system/types.ts
mdydek: Frankly, I'm not so sure, but probably there are use-cases where this could be useful and in my opinion it is better to write them all, rather than those needed and later on to add them, but decision is up to you

maciejmakowski2003: okey, sounds rational

packages/react-native-audio-api/src/system/AudioManager.ts
maciejmakowski2003: yeah

PR #499, opened 2025-06-10
fix: additional conditions to if that handled looping

### LINKED ISSUE #496: Updating `loopStart` causes audio to play at a wrong position
### Description

Hi,

First of all thank your for the lib, great work 👏
However I am having an issue. If I were to update the `loopStart` consequently, very fast, then the audio starts playing at a wrong position. This happens in both cases but with different severity:
1. audio is already playing and then I do the updates -> played part is very off
2. updates are done and then node is started -> played part is slightly off

Just to give some context, the UI I am building is something where a user can drag a start/end handle to update the start and end playing of audio. I am updating the `loopStart` as the user drags which generates a lot of update events, reproducing the behavior explained. I understand it may be wiser to throttle updates which I may later do but it is still an issue that I would love for it to be fixed.

### Steps to reproduce

1. Clone [repo](https://github.com/itsramiel/AudioAppRepro)
2. run app
3. notice that I am updating the `loopStart` multiple times in a short time with a `sleep` function in between and the last update is for a start time of 7.15 but the audio plays way past that almost at the end like ~10, 11

https://github.com/itsramiel/AudioAppRepro

0.6.1

### React Native version

0.79.3

### Platforms

iOS, Android

### JavaScript runtime

Hermes

### Workflow

React Native

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

mdydek:
Closes #496

## Introduced changes

- added several conditions that check also the direction of playing
- removed underflow error caused by two unsigned data types being subtracted from each other and result was below 0

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/src/core/AudioBufferSourceNode.ts
+++ b/packages/react-native-audio-api/src/core/AudioBufferSourceNode.ts
@@ -36,7 +36,7 @@ export default class AudioBufferSourceNode extends AudioScheduledSourceNode {
   public get loopSkip(): boolean {
     return (this.node as IAudioBufferSourceNode).loopSkip;
   }
-  
+
   public set loopSkip(value: boolean) {
     (this.node as IAudioBufferSourceNode).loopSkip = value;
   }


packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferSourceNode.cpp
```diff
@@ -247,8 +247,13 @@ void AudioBufferSourceNode::processWithoutInterpolation(
...
 
-  if (loop_ && (readIndex >= frameEnd || readIndex < frameStart)) {
-    readIndex = frameStart + (readIndex - frameStart) % frameDelta;
+  // if we are moving towards loop, we do nothing becuase we will achieve it
+  // otherwise, we wrap to the start of the loop if necessary
+  if (loop_ &&
+      ((readIndex >= frameEnd && direction == 1) ||
+       (readIndex < frameStart && direction == -1))) {
+    readIndex = frameStart +
+        ((long long int)readIndex - (long long int)frameStart) % frameDelta;
```
itsramiel: When `readIndex` < `frameStart`, readIndex must be set to frameStart. 

With this implementation, if the audio is playing and is second 0 and loopStart is set to 5, then the audio will not move to second 5 but instead continue playing from 0.

mdydek: I’m well aware of that, but check out this example: https://mdn.github.io/webaudio-examples/audio-buffer-source-node/loop/ . We try to be precise with web audio api implementation and this example shows that it is a desired behavior.

itsramiel: Hmm I cant tell where the progress is of the currently playing audio as in if I set loopStart ahead of current progress whether it moves or keeps playing

itsramiel: Assuming that is the spec and we need to go ahead with it, although I would love not to,h ere is my use case:

with patch of #500 
https://github.com/user-attachments/assets/054ce0c3-6d0b-4b7e-a448-852eedc9c636

I drag start thumg and expect the audio to move along. How would I implement that without `loopStart` moving the currently played audio.

mdydek: I’ve tested it this way: put start at 0 and end at 7, then after clicking play move start at 6, after moving there is no jumping in sound at all and you can clearly hear track looping from 6 to 7 after tht

mdydek: https://github.com/software-mansion/react-native-audio-api/blob/main/apps/common-app/src/examples/AudioFile/AudioFile.tsx I think you can refer to this example, here we utilize `offset` parameter in start function, which can start playing track from desired place and you can set `loopStart` also to that value.

itsramiel: but that means everytime the thumb is moved, I would have to disconnect the audio source node, recreate one and start it with an offset which would cause an audible delay i guess?

itsramiel: web api is limiting 😅

michalsek: Someone thought that "Creating audio nodes is cheap" and now everyone is cracking their heads why 😅 

@mdydek I think we could add a param to ABSN constructor that controls behaviour or loop parameters.

I think we are closer and closer to writing ABSN on web ourselves anyway 🙂

itsramiel: I am not exactly sure I understand that. Sorry was not aware that this is how web standards defines it. I will implement my functionality around the api and see if I hit any blockers and ask for assistance then.

appreciate the hard work 🙏

michalsek: Oh, sorry, I'm so used to this weird part of the spec, I tend to cut out all of the context.

So the one of the base principles of how web audio is implemented, is that all source nodes are one time use only (once it stops playing is not usable anymore). One of the arguments behind that, brought up by one of the spec contributors is that "creating nodes is cheap operation".

This single assumption creates a lot of weird cases, when dealing with web audio (and in most of the cases audio-api too). 

F.e. simple case of pausing a single buffer involves getting rid of the old AudioBufferSourceNode, creating new one and calculating start offset based on start/stop timing of the orginal one.

So i'm pretty sure, that in your use-case, the expected by the specification way of doing this would be to create new AudioBufferSource node, that will have new loop parameters (you can reuse the same AudioBuffer though 😄)


But the approach i've suggested to Michal is to introduce new constructor level parameter, that would steer the loopStart/End behaviour between default spec-compatible and the one more user-friendly :)

f.e.
```ts
const source = audioContext.createBufferSource({ skipToLoopStart: true });
```

this - for now - would be a mobile only options, but i'm pretty sure we will eventually write our own version of `AudioBufferSourceNode` with all the features we found lacking.

PR #504, opened 2025-06-12
Refactor/audio buffer base source node

maciejmakowski2003:
Closes #

## ⚠️ Breaking changes ⚠️

- Aligned `onended` event type with web. Currently it has empty body.
- Changed signature of `enqueueBuffer` to `enqueueBuffer: (audioBuffer: IAudioBuffer, isLastBuffer: boolean) => void`

## Introduced changes

- Implemented `AudioBufferBaseSourceNode` as a parent interface for all classes that play `AudioBuffers`

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferQueueSourceNode.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferQueueSourceNode.cpp
@@ -12,22 +12,9 @@ namespace audioapi {
 
 AudioBufferQueueSourceNode::AudioBufferQueueSourceNode(
     BaseAudioContext *context)
-    : AudioBufferBaseSourceNode(context), vReadIndex_(0.0) {
+    : AudioBufferBaseSourceNode(context) {
   buffers_ = {};
-
-  detuneParam_ = std::make_shared<AudioParam>(
-      0.0, MOST_NEGATIVE_SINGLE_FLOAT, MOST_POSITIVE_SINGLE_FLOAT, context);
-  playbackRateParam_ = std::make_shared<AudioParam>(
-      1.0, MOST_NEGATIVE_SINGLE_FLOAT, MOST_POSITIVE_SINGLE_FLOAT, context);
-
-  playbackRateBus_ = std::make_shared<AudioBus>(
-      RENDER_QUANTUM_SIZE * 3, channelCount_, context_->getSampleRate());
-
-  stretch_ =
-      std::make_shared<signalsmith::stretch::SignalsmithStretch<float>>();
   stretch_->presetDefault(channelCount_, context_->getSampleRate(), true);
-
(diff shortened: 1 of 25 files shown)


packages/react-native-audio-api/src/core/AudioBufferQueueSourceNode.ts
```diff
@@ -44,7 +29,7 @@ export default class AudioBufferQueueSourceNode extends AudioScheduledSourceNode
       }
     }
 
-    (this.node as IAudioBufferQueueSourceNode).start(when, offset);
```
mdydek, line 47: why starting with offset is removed?

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferQueueSourceNode.cpp
```diff
@@ -67,10 +37,9 @@ void AudioBufferQueueSourceNode::pause() {
 
 void AudioBufferQueueSourceNode::enqueueBuffer(
     const std::shared_ptr<AudioBuffer> &buffer,
-    int bufferId,
     bool isLastBuffer) {
   auto locker = Locker(getBufferLock());
-  buffers_.emplace(bufferId, buffer);
```
mdydek, line 73: are we getting rid of `bufferId` totally?

packages/react-native-audio-api/src/core/AudioBufferQueueSourceNode.ts
maciejmakowski2003: because it does not make sense to start with offset on the interface which behaves like streaming source

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferQueueSourceNode.cpp
maciejmakowski2003: yeah

packages/react-native-audio-api/src/core/AudioBufferQueueSourceNode.ts
mdydek: can you explain more what is the use-case for this interface?

maciejmakowski2003: playing audio that is consisted of many `AudioBuffers` delivering on fly

mdydek: so basically offset is useless because they last f.e. 1 second?

maciejmakowski2003: yeah and in general it imitates streaming so we want to start from the beginning and pause if needed

michalsek (approved):

PR #569, opened 2025-07-18
Feat/recorder adapter node

poneciak57:
Closes RNAA-176

## ⚠️ Breaking changes ⚠️

- implemented RecorderAdapterNode so now Recorder can be connected into the audio graph

## Introduced changes

- AudioRecorder can be connected to the audio graph through RecorderAdapterNode

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/apps/common-app/src/examples/Record/Record.tsx
+++ b/apps/common-app/src/examples/Record/Record.tsx
@@ -1,19 +1,22 @@
 import React, { useRef, FC, useEffect } from 'react';
 import {
-  AudioBuffer,
+  // AudioBuffer,
   AudioContext,
   AudioManager,
   AudioRecorder,
-  AudioBufferSourceNode,
+  // AudioBufferSourceNode,
+  RecorderAdapterNode
 } from 'react-native-audio-api';
 
 import { Container, Button } from '../../components';
 
 const Record: FC = () => {
   const recorderRef = useRef<AudioRecorder | null>(null);
-  const audioBuffersRef = useRef<AudioBuffer[]>([]);
-  const sourcesRef = useRef<AudioBufferSourceNode[]>([]);
+  // const audioBuffersRef = useRef<AudioBuffer[]>([]);
+  // const sourcesRef = useRef<AudioBufferSourceNode[]>([]);
   const aCtxRef = useRef<AudioContext | null>(null);
+  const recorderAdapterRef = useRef<RecorderAdapterNode | null>(null);
+
 
   useEffect(() => {
     AudioManager.setAudioSessionOptions({
@@ -22,78 +25,102 @@ const Record: FC = () => {
       iosOptions: ['allowBluetooth', 'defaultToSpeaker'],
     });
 
-    recorderRef.current = new AudioRecorder({
-      sampleRate: 16000,
-      bufferLengthInSamples: 16000,
-    });
   }, []);
 
-  const onReplay = () => {
-    const aCtx = new AudioContext({ sampleRate: 16000 });
-    aCtxRef.current = aCtx;
 
-    if (aCtx.state === 'suspended') {
-      aCtx.resume();
-    }
-
-    const tNow = aCtx.currentTime;
-    let nextStartAt = tNow + 1;
-    const buffers = audioBuffersRef.current;
-
-    console.log(tNow, nextStartAt, buffers.length);
-
-    for (let i = 0; i < buffers.length; i++) {
-      const source = aCtx.createBufferSource();
-      source.buffer = buffers[i];
-
-      source.connect(aCtx.destination);
-      source.onended = () => {
-        console.log('Audio buffer source ended');
-      };
-      sourcesRef.current.push(source);
-
-      source.start(nextStartAt);
-      nextStartAt += buffers[i].duration;
-    }
-
-    setTimeout(
-      () => {
-        console.log('clearing data');
-        audioBuffersRef.current = [];
-        sourcesRef.current = [];
-      },
-      (nextStartAt - tNow) * 1000
-    );
-  };
-
-  const onRecord = () => {
+  const start = () => {
+    aCtxRef.current = new AudioContext({ sampleRate: 16000 });
     if (!recorderRef.current) {
-      return;
+      recorderRef.current = new AudioRecorder({
+        sampleRate: 16000,
+        bufferLengthInSamples: 16000,
+      });
     }
-
+    recorderAdapterRef.current = aCtxRef.current.createRecorderAdapter();
+    recorderAdapterRef.current.setRecorder(recorderRef.current);
+    recorderAdapterRef.current.connect(aCtxRef.current.destination);
+    recorderAdapterRef.current.start();
     recorderRef.current.onAudioReady((event) => {
       const { buffer, numFrames, when } = event;
-
-      console.log(
-        'Audio recorder buffer ready:',
-        buffer.duration,
-        numFrames,
-        when
-      );
-      audioBuffersRef.current.push(buffer);
+      console.log('Audio recorder buffer ready:', buffer.duration, numFrames, when);
     });
-
     recorderRef.current.start();
+    console.log('Recording started');
+    console.log('Audio context state:', aCtxRef.current.state);
+    if (aCtxRef.current.state === 'suspended') {
+      console.log('Resuming audio context');
+      aCtxRef.current.resume();
+    }
+  }
 
-    setTimeout(() => {
-      recorderRef.current?.stop();
-    }, 3000);
-  };
+  // const onReplay = () => {
+  //   const aCtx = new AudioContext({ sampleRate: 16000 });
+  //   aCtxRef.current = aCtx;
+  //   recorderAdapterRef.current = aCtx.createRecorderAdapter();
+
+  //   if (aCtx.state === 'suspended') {
+  //     aCtx.resume();
+  //   }
+
+  //   const tNow = aCtx.currentTime;
+  //   let nextStartAt = tNow + 1;
+  //   const buffers = audioBuffersRef.current;
+
+  //   console.log(tNow, nextStartAt, buffers.length);
+
+  //   for (let i = 0; i < buffers.length; i++) {
+  //     const source = aCtx.createBufferSource();
+  //     source.buffer = buffers[i];
+
+  //     source.connect(aCtx.destination);
+  //     source.onEnded = () => {
+  //       console.log('Audio buffer source ended');
+  //     };
+  //     sourcesRef.current.push(source);
+
+  //     source.start(nextStartAt);
+  //     nextStartAt += buffers[i].duration;
+  //   }
+
+  //   setTimeout(
+  //     () => {
+  //       console.log('clearing data');
+  //       audioBuffersRef.current = [];
+  //       sourcesRef.current = [];
+  //     },
+  //     (nextStartAt - tNow) * 1000
+  //   );
+  // };
+
+  // const onRecord = () => {
+  //   if (!recorderRef.current) {
+  //     return;
+  //   }
+
+  //   recorderRef.current.onAudioReady((event) => {
+  //     const { buffer, numFrames, when } = event;
+
+  //     console.log(
+  //       'Audio recorder buffer ready:',
+  //       buffer.duration,
+  //       numFrames,
+  //       when
+  //     );
+  //     audioBuffersRef.current.push(buffer);
+  //   });
+
+  //   recorderRef.current.start();
+
+  //   setTimeout(() => {
+  //     recorderRef.current?.stop();
+  //   }, 3000);
+  // };
 
   return (
     <Container centered>
-      <Button title="Record" onPress={onRecord} />
-      <Button title="Replay" onPress={onReplay} />
+      <Button title="Start Recording" onPress={start} />
+      {/* <Button title="Record" onPress={onRecord} />
+      <Button title="Replay" onPress={onReplay} /> */}
     </Container>
   );
 };
--- a/apps/common-app/src/examples/Record/Record.tsx
+++ b/apps/common-app/src/examples/Record/Record.tsx
@@ -20,24 +20,16 @@ const Record: FC = () => {
       iosMode: 'spokenAudio',
       iosOptions: ['allowBluetooth', 'defaultToSpeaker'],
     });
-    AudioManager.requestRecordingPermissions()
-      .then(() => {
-
-    if (!aCtxRef.current) {
-      aCtxRef.current = new AudioContext({ sampleRate: 48000 });
-    }
-    if (!recorderRef.current) {
-      recorderRef.current = new AudioRecorder({
-        sampleRate: 48000,
-        bufferLengthInSamples: 48000,
-      });
-    }
     
-    /// Create the recorder adapter and connect it to the audio context
-    if (!recorderAdapterRef.current) {
-      recorderAdapterRef.current = aCtxRef.current.createRecorderAdapter();
-    }
+    recorderRef.current = new AudioRecorder({
+      sampleRate: 16000,
+      bufferLengthInSamples: 16000,
+    });
+    aCtxRef.current = new AudioContext({ sampleRate: 16000 });
+    recorderAdapterRef.current = aCtxRef.current.createRecorderAdapter();
+    
     recorderAdapterRef.current.setRecorder(recorderRef.current);
+    // recorderRef.current.connect(recorderAdapterRef.current); // TODO this would be desired flow
     recorderAdapterRef.current.connect(aCtxRef.current.destination);
     console.log('Recorder adapter connected to audio context');
     
@@ -45,8 +37,6 @@ const Record: FC = () => {
       console.log('Resuming audio context');
       aCtxRef.current.resume();
     }
-
-      });
   }, []);
 
 
--- a/apps/common-app/src/examples/Record/Record.tsx
+++ b/apps/common-app/src/examples/Record/Record.tsx
@@ -1,10 +1,8 @@
-import React, { useRef, FC, useEffect } from 'react';
+import React, { useRef, FC, useEffect, act } from 'react';
 import {
-  // AudioBuffer,
   AudioContext,
   AudioManager,
   AudioRecorder,
-  // AudioBufferSourceNode,
   RecorderAdapterNode
 } from 'react-native-audio-api';
 
@@ -12,8 +10,6 @@ import { Container, Button } from '../../components';
 
 const Record: FC = () => {
   const recorderRef = useRef<AudioRecorder | null>(null);
-  // const audioBuffersRef = useRef<AudioBuffer[]>([]);
-  // const sourcesRef = useRef<AudioBufferSourceNode[]>([]);
   const aCtxRef = useRef<AudioContext | null>(null);
   const recorderAdapterRef = useRef<RecorderAdapterNode | null>(null);
 
@@ -29,23 +25,29 @@ const Record: FC = () => {
 
 
   const start = () => {
-    aCtxRef.current = new AudioContext({ sampleRate: 16000 });
+    if (!aCtxRef.current) {
+      aCtxRef.current = new AudioContext({ sampleRate: 16000 });
+    }
     if (!recorderRef.current) {
       recorderRef.current = new AudioRecorder({
         sampleRate: 16000,
         bufferLengthInSamples: 16000,
       });
     }
+
+    /// Create the recorder adapter and connect it to the audio context
     recorderAdapterRef.current = aCtxRef.current.createRecorderAdapter();
     recorderAdapterRef.current.setRecorder(recorderRef.current);
     recorderAdapterRef.current.connect(aCtxRef.current.destination);
     recorderAdapterRef.current.start();
-    recorderRef.current.onAudioReady((event) => {
-      const { buffer, numFrames, when } = event;
-      console.log('Audio recorder buffer ready:', buffer.duration, numFrames, when);
-    });
+
+    // recorderRef.current.onAudioReady((event) => {
+    //   const { buffer, numFrames, when } = event;
+    //   console.log('Audio recorder buffer ready:', buffer.duration, numFrames, when);
+    // });
+    
     recorderRef.current.start();
-    console.log('Recording started');
+    console.log('Recording started'); 
     console.log('Audio context state:', aCtxRef.current.state);
     if (aCtxRef.current.state === 'suspended') {
       console.log('Resuming audio context');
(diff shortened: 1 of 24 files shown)


packages/react-native-audio-api/common/cpp/audioapi/core/inputs/AudioRecorder.cpp
```diff
@@ -16,6 +17,8 @@ AudioRecorder::AudioRecorder(
       audioEventHandlerRegistry_(audioEventHandlerRegistry) {
   circularBuffer_ =
       std::make_shared<CircularAudioArray>(std::max(2 * bufferLength, 2048));
+  adapterBuffer_ = std::make_shared<CircularOverflowableAudioArray>(
+      std::max(2 * bufferLength, 2048));
```
msluszniak: Remove magic numbers and repetitions, replace my variable names with something even better ;p
```suggestion
  constexpr auto veryGoodNameForConstant = 2048;
  const auto alsoPrettyDecentName = 2 * bufferLength;
  const auto superName = std::max(veryGoodNameForConstant, alsoPrettyDecentName);
  circularBuffer_ = std::make_shared<CircularAudioArray>(superName);
  adapterBuffer_ = std::make_shared<CircularOverflowableAudioArray>(superName);
```

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/AudioRecorder.cpp
```diff
@@ -55,4 +58,13 @@ void AudioRecorder::sendRemainingData() {
   invokeOnAudioReadyCallback(bus, availableFrames, 0);
 }
 
+void AudioRecorder::readFrames(float *output, size_t framesToRead) {
+  size_t readFrames = adapterBuffer_->read(output, framesToRead);
+
+  if (readFrames < framesToRead) {
+    // Fill the rest with silence
+    memset(output + readFrames, 0, (framesToRead - readFrames) * sizeof(float));
```
msluszniak: prefer std::memset

packages/react-native-audio-api/common/cpp/audioapi/core/sources/RecorderAdapterNode.cpp
```diff
@@ -0,0 +1,43 @@
...
+    int framesToProcess) {
+  size_t startOffset = 0;
+  size_t offsetLength = 0;
+  updatePlaybackInfo(processingBus, framesToProcess, startOffset, offsetLength);
+  if (!isPlaying() && !isStopScheduled()) {
+    processingBus->zero();
+    return;
+  }
+
+  if (!recorder_) {
```
msluszniak: [readability-implicit-bool-conversion](https://clang.llvm.org/extra/clang-tidy/checks/readability/implicit-bool-conversion.html)
```suggestion
  if (recorder_ == nullptr) {
```

packages/react-native-audio-api/common/cpp/audioapi/utils/CircularOverflowableAudioArray.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+CircularOverflowableAudioArray::CircularOverflowableAudioArray(size_t size)
+    : AudioArray(size) {}
+
+void CircularOverflowableAudioArray::write(const float *data, size_t size) {
+  size_t writeIndex = vWriteIndex_.load(std::memory_order_relaxed);
+
+  /// Advances the read index if there is not enough space
+  readLock_.lock();
+  size_t availableSpace = (vReadIndex_ - writeIndex - 1 + size_) % size_;
```
msluszniak: You must be precise here. If `vReadIndex_ - writeIndex` might be negative you will have negative intermediate value for unsigned type. Move size_ to prevent this. If it cannot be negative remove size_ completely since it is neutral element in adding element modulo size_
```suggestion
  size_t availableSpace = (size_ + vReadIndex_ - writeIndex - 1) % size_;
```

packages/react-native-audio-api/common/cpp/audioapi/utils/CircularOverflowableAudioArray.cpp
```diff
@@ -0,0 +1,51 @@
...
+    vReadIndex_ = (writeIndex + size + 1) % size_;
+  }
+  readLock_.unlock();
+
+  size_t partSize = size_ - writeIndex;
+  if (size > partSize) {
+    memcpy(data_ + writeIndex, data, partSize * sizeof(float));
+    memcpy(data_, data + partSize, (size - partSize) * sizeof(float));
+  } else {
+    memcpy(data_ + writeIndex, data, size * sizeof(float));
```
msluszniak: std::memcpy

packages/react-native-audio-api/common/cpp/audioapi/utils/CircularOverflowableAudioArray.cpp
```diff
@@ -0,0 +1,51 @@
...
+  size_t availableSpace = getAvailableSpace();
+  size_t readSize = std::min(size, availableSpace);
+
+  size_t partSize = size_ - vReadIndex_;
+  if (readSize > partSize) {
+    memcpy(output, data_ + vReadIndex_, partSize * sizeof(float));
+    memcpy(output + partSize, data_, (readSize - partSize) * sizeof(float));
+  } else {
+    memcpy(output, data_ + vReadIndex_, readSize * sizeof(float));
+  }
```
msluszniak: Very similar code here as above, refactor it.

packages/react-native-audio-api/common/cpp/audioapi/utils/CircularOverflowableAudioArray.cpp
```diff
@@ -0,0 +1,51 @@
...
+  readLock_.unlock();
+  return readSize;
+}
+
+size_t CircularOverflowableAudioArray::getAvailableSpace() const {
+  return (vWriteIndex_.load(std::memory_order_relaxed) - vReadIndex_ + size_) %
+      size_;
+}
+
+} // namespace audioapi
```
msluszniak: It is recommended to add ending newline to make file POSIX compliant
```suggestion
} // namespace audioapi

```

packages/react-native-audio-api/common/cpp/audioapi/core/sources/RecorderAdapterNode.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include <audioapi/core/sources/RecorderAdapterNode.h>
+#include <audioapi/utils/AudioArray.h>
+#include <audioapi/utils/AudioBus.h>
+
+namespace audioapi {
+
+RecorderAdapterNode::RecorderAdapterNode(BaseAudioContext *context)
+    : AudioScheduledSourceNode(context), recorder_(nullptr) {}
+
+void RecorderAdapterNode::setRecorder(
```
msluszniak: You can mark this `noexcept` and generally make sure that you placed all `const` `noexcept`, `final` etc.

packages/react-native-audio-api/common/cpp/audioapi/utils/CircularOverflowableAudioArray.h
```diff
@@ -0,0 +1,42 @@
...
+  /// @brief Reads data from the circular buffer.
+  /// @param output Pointer to the output buffer.
+  /// @param size Number of frames to read.
+  /// @return The number of frames actually read.
+  size_t read(float *output, size_t size);
+
+ private:
+  std::atomic<size_t> vWriteIndex_ = { 0 };
+  size_t vReadIndex_ = 0; // it is only used after acquiring readLock_
+  std::mutex readLock_;
```
msluszniak: Check M&M rule for c++, you should mark this mutex mutable, and probably for `std::atomic<size_t>` as well but this needs to be individually checked because not all atomic var needs to be marked with this qualifier ;)

poneciak57:
It looks like AudioRecorder doesn't support web platforms, so unfortunately my adapter won't be compatible with the web either.
Please let me know if the implementation looks good to you—I'll add the documentation once I get your feedback.

packages/react-native-audio-api/common/cpp/audioapi/utils/CircularOverflowableAudioArray.cpp
```diff
@@ -0,0 +1,57 @@
...
+namespace audioapi {
+
+CircularOverflowableAudioArray::CircularOverflowableAudioArray(
+    size_t
+        size) noexcept(std::is_nothrow_constructible<AudioArray, size_t>::value)
+    : AudioArray(size) {}
+
+void CircularOverflowableAudioArray::write(
+    const float *data,
+    const size_t size) {
```
msluszniak, line 13: const next to size_t triggers this clang tidy check: https://clang.llvm.org/extra/clang-tidy/checks/readability/avoid-const-params-in-decls.html

Please remove const from `const size_t size` both in function implementation and definition ;)

packages/react-native-audio-api/common/cpp/audioapi/utils/CircularOverflowableAudioArray.h
```diff
@@ -0,0 +1,42 @@
...
+
+  /// @brief Reads data from the circular buffer.
+  /// @param output Pointer to the output buffer.
+  /// @param size Number of frames to read.
+  /// @return The number of frames actually read.
+  size_t read(float *output, size_t size) const;
+
+ private:
+  std::atomic<size_t> vWriteIndex_ = { 0 };
+  mutable size_t vReadIndex_ = 0; // it is only used after acquiring readLock_
```
msluszniak, line 36: I believe that this mutable next to non-atomic variable is not needed, maybe more for an atomic vWriteIndex_ ;)

apps/common-app/src/examples/Record/Record.tsx
```diff
@@ -1,101 +1,181 @@
...
+      iosOptions: ['defaultToSpeaker', 'allowBluetoothA2DP'],
     });
-
+    
     recorderRef.current = new AudioRecorder({
-      sampleRate: 16000,
-      bufferLengthInSamples: 16000,
+      sampleRate: 48000,
+      bufferLengthInSamples: 48000,
     });
```
michalsek: Can you also check if we properly handle recording sampleRate of f.e. 16k while audio context works in 48k (or 44.1k)?

apps/common-app/src/examples/Record/Record.tsx
```diff
@@ -1,101 +1,181 @@
...
-      <Button title="Replay" onPress={onReplay} />
+      <Button title="Start Recording" onPress={start} />
+      <Button title="Stop Recording" onPress={stopRecorder} />
     </Container>
   );
 };
 
 export default Record;
+
+// import React, { useRef, FC, useEffect } from 'react';
```
michalsek: I think we can remove this code, optionaly it would be great to have both versions in this example screen (so we could easier search for regressions :))

packages/audiodocs/docs/sources/recorder-adapter-node.mdx
```diff
@@ -0,0 +1,42 @@
...
+    sampleRate: 48000,
+    bufferLengthInSamples: 48000,
+});
+aCtxRef.current = new AudioContext({ sampleRate: 48000 });
+recorderAdapterRef.current = aCtxRef.current.createRecorderAdapter();
+
+// Connecting adapter to recorder
+recorderRef.current.connect(recorderAdapterRef.current);
+// Connecting adapter to destination
+recorderAdapterRef.current.connect(aCtxRef.current.destination)
```
michalsek: Can you make this example not relying on refs? Just:
```tsx
const recorder = new AudioRecorder({ sampleRate: 48000, bufferLengthInSamples: 4800 });

const audioContext = new AudioContext({ sampleRate: 48000 });

// and so on...
```

packages/react-native-audio-api/ios/audioapi/ios/core/NativeAudioRecorder.m
```diff
@@ -15,6 +15,13 @@ - (instancetype)initWithReceiverBlock:(AudioReceiverBlock)receiverBlock
     self.receiverBlock = [receiverBlock copy];
 
     float devicePrefferedSampleRate = [[AVAudioSession sharedInstance] sampleRate];
+    NSLog(@"Device sample rate: %f, requested: %f", devicePrefferedSampleRate, sampleRate);
```
michalsek:
```suggestion
```

packages/react-native-audio-api/ios/audioapi/ios/core/IOSAudioRecorder.mm
```diff
@@ -19,8 +20,9 @@
 {
   AudioReceiverBlock audioReceiverBlock = ^(const AudioBufferList *inputBuffer, int numFrames, AVAudioTime *when) {
     if (isRunning_.load()) {
+      // NSLog(@"Received audio data with %d frames at time: %@", numFrames, when);
```
michalsek: cleanup :)

packages/react-native-audio-api/common/cpp/audioapi/utils/CircularOverflowableAudioArray.cpp
```diff
@@ -0,0 +1,57 @@
...
+  size_t writeIndex = vWriteIndex_.load(std::memory_order_relaxed);
```
michalsek, line 14: can we use different name than `vWriteIndex_` / `vReadIndex_`

in AudioBufferSourceNode (base!) we are using `v[]Index` as a float index that can land in between array elements when we are interpolating the signal. Same naming here would suggest that something similar happens here too

apps/common-app/src/examples/Record/Record.tsx
```diff
@@ -1,101 +1,181 @@
...
     recorderRef.current = new AudioRecorder({
-      sampleRate: 16000,
-      bufferLengthInSamples: 16000,
+      sampleRate: 48000,
+      bufferLengthInSamples: 48000,
     });
+    aCtxRef.current = new AudioContext({ sampleRate: 48000 });
+    recorderAdapterRef.current = aCtxRef.current.createRecorderAdapter();
+    
+    recorderRef.current.connect(recorderAdapterRef.current); // TODO this would be desired flow
```
michalsek:
```suggestion
    recorderRef.current.connect(recorderAdapterRef.current);
```

packages/audiodocs/docs/sources/recorder-adapter-node.mdx
```diff
@@ -0,0 +1,42 @@
...
+## Properties
+None
+
+## Methods
+None
+
+## Remarks
+- Adapter without the recorder will produce silence
+- Adapter that is connected only to the recorder will work fine and keep small buffer of recorded Data
+- Adapter that is connected only to the recorder will not be garbage collected
```
michalsek: Even if audio recorder is garbage collected? This seems like something that should be taken care of

packages/react-native-audio-api/common/cpp/audioapi/utils/CircularOverflowableAudioArray.cpp
poneciak57: same name appears in `CircularAudioArray`, should i change it too ?

packages/audiodocs/docs/sources/recorder-adapter-node.mdx
poneciak57: no if audio recorder is destroyed adapter node should be garbage collected i can point that out in documentation

apps/common-app/src/examples/Record/Record.tsx
poneciak57: checked :)

apps/common-app/src/examples/Record/Record.tsx
```diff
@@ -1,35 +1,102 @@
...
-  const onReplay = () => {
-    const aCtx = new AudioContext({ sampleRate: 16000 });
+
+  const startEcho = () => {
+    if (!recorderRef.current) {
+      console.error('AudioContext or AudioRecorder is not initialized');
+      return;
+    }
+
+    aCtxRef.current = new AudioContext({ sampleRate: SAMPLE_RATE });
```
maciejmakowski2003, line 45: I think it would be better to create AudioContext only once, not at every `startEcho` invocation

packages/audiodocs/docs/inputs/audio-recorder.mdx
```diff
@@ -51,6 +51,16 @@ The above stops recording.
 
 #### Returns `undefined`.
 
+### `connect`
+
+The above method connects recorder to the adapter.
+
+| Parameters | Type | Description |
+| :---: | :---: | :---- |
+| `node` | [RecorderAdapterNode](/sources/recorder-adapter-node) | adapter that we want to connect to |
```
maciejmakowski2003:
```suggestion
| `node` | [RecorderAdapterNode](/sources/recorder-adapter-node) | Adapter that we want to connect to. |
```

packages/audiodocs/docs/sources/recorder-adapter-node.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+```tsx
+const recorder = new AudioRecorder({
+    sampleRate: 48000,
+    bufferLengthInSamples: 48000,
+});
+const audioContext = new AudioContext({ sampleRate: 48000 });
+const recorderAdapterNode = aCtxRef.current.createRecorderAdapter();
+
+// Connecting adapter to recorder
```
maciejmakowski2003: imo unnecessary comments

packages/audiodocs/docs/sources/recorder-adapter-node.mdx
```diff
@@ -0,0 +1,42 @@
...
+recorder.connect(recorderAdapterNode);
+// Connecting adapter to destination
+recorderAdapterNode.connect(audioContext.destination)
+```
+
+## Properties
+None
+
+## Methods
+None
```
maciejmakowski2003: If we do not have any methods or properties we can just skip these sections

packages/audiodocs/docs/sources/recorder-adapter-node.mdx
```diff
@@ -0,0 +1,42 @@
...
+
+## Remarks
+- Adapter without the recorder will produce silence
```
maciejmakowski2003, line 32: I am not sure if that is the best place for these remarks. They are related with all `AudioNodes` (garbage collecting and producing silence without source connected). I would consider adding section to `AudioNode` page about it.

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioRecorderHostObject.h
```diff
@@ -45,7 +46,15 @@ class AudioRecorderHostObject : public JsiHostObject {
 
     addFunctions(
       JSI_EXPORT_FUNCTION(AudioRecorderHostObject, start),
-      JSI_EXPORT_FUNCTION(AudioRecorderHostObject, stop));
+      JSI_EXPORT_FUNCTION(AudioRecorderHostObject, stop),
+      JSI_EXPORT_FUNCTION(AudioRecorderHostObject, connect));
```
maciejmakowski2003: could you add `disconnect` method too?

packages/react-native-audio-api/common/cpp/audioapi/core/sources/RecorderAdapterNode.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include <type_traits>
+
+namespace audioapi {
+
+RecorderAdapterNode::RecorderAdapterNode(BaseAudioContext *context) noexcept(
+    std::is_nothrow_constructible<AudioNode, BaseAudioContext *>::value)
+    : AudioNode(context) {
+  // It should be marked as initialized only after it is connected to the
+  // recorder. Internall buffer size is based on the recorder's buffer length.
+  isInitialized_ = false;
```
maciejmakowski2003, line 14: I am not sure if it is a great approach to that issue. In general `isInitialized` property give us information if node can be processed by the graph and in this case it should be possible without `AudioRecorder` connected (it should produce silence).

packages/react-native-audio-api/common/cpp/audioapi/utils/CircularOverflowableAudioArray.h
maciejmakowski2003, line 36: why we need `mutable` here and below?

packages/react-native-audio-api/src/core/AudioRecorder.ts
```diff
@@ -3,9 +3,11 @@ import { AudioRecorderOptions } from '../types';
 import AudioBuffer from './AudioBuffer';
 import { OnAudioReadyEventType } from '../events/types';
 import { AudioEventEmitter } from '../events';
+import RecorderAdapterNode from './RecorderAdapterNode';
 
 export default class AudioRecorder {
-  protected readonly recorder: IAudioRecorder;
+  public readonly recorder: IAudioRecorder;
```
maciejmakowski2003: why it is public? it should not be visible for the user.

if you want to make it visible only internally, you can do sth like below.
```
/** @internal */
  public readonly buffer: IAudioBuffer;
```

packages/react-native-audio-api/src/core/RecorderAdapterNode.ts
```diff
@@ -0,0 +1,8 @@
+import { IRecorderAdapterNode } from '../interfaces';
+import AudioNode from './AudioNode';
+
+export default class RecorderAdapterNode extends AudioNode {
+  public getNode(): IRecorderAdapterNode {
```
maciejmakowski2003, line 9: it should also be internal method.

packages/audiodocs/docs/sources/recorder-adapter-node.mdx
poneciak57: but these only aply to this node

packages/react-native-audio-api/common/cpp/audioapi/utils/CircularOverflowableAudioArray.h
poneciak57: so we can mark read method as const

apps/common-app/src/examples/Record/Record.tsx
poneciak57: i think it does not matter that much and creating the context once will cause the user to hear the silence noise before even starting anything

packages/react-native-audio-api/common/cpp/audioapi/core/sources/RecorderAdapterNode.cpp
poneciak57: but if it was never connected to the recorder it does not make sense to process it i guess ?

apps/common-app/src/examples/Record/Record.tsx
maciejmakowski2003: okey, lets leave it. it is only example

packages/audiodocs/docs/sources/recorder-adapter-node.mdx
maciejmakowski2003: hmm, okey lets leave it.

packages/react-native-audio-api/common/cpp/audioapi/core/sources/RecorderAdapterNode.cpp
maciejmakowski2003: it is the same question that we can ask about every effect node. imo it should return silent buffer or sth like it if there is no connection.

packages/react-native-audio-api/common/cpp/audioapi/utils/CircularOverflowableAudioArray.h
maciejmakowski2003: ohh thats right. my bad

packages/react-native-audio-api/src/core/RecorderAdapterNode.ts
```diff
@@ -0,0 +1,12 @@
+import { IRecorderAdapterNode } from '../interfaces';
+import AudioNode from './AudioNode';
+
+export default class RecorderAdapterNode extends AudioNode {
+  /** @internal */
+  public isInitialized: boolean = false;
```
maciejmakowski2003:
```suggestion
  public isConnected: boolean = false;
```

packages/react-native-audio-api/src/core/AudioRecorder.ts
```diff
@@ -22,6 +24,16 @@ export default class AudioRecorder {
     this.recorder.stop();
   }
 
+  public connect(node: RecorderAdapterNode): void {
```
maciejmakowski2003, line 27: what about disconnect method?

poneciak57: oh XD

maciejmakowski2003 (approved):
great work!!!

michalsek:
Lets merge this! 🔥

PR #578, opened 2025-07-29
Audio Context/Recorder lifecycle improvements

michalsek:
closes RNAA-192

## Introduced changes

- adds experimental function to track if other apps are playing and calling own interruption began and ended events. Usefull for when we want to interrupt locally even when using mixable category and options. It is not ideal, but lets test it until we find better solution for playback and record category with coexistence with f.e. navigation apps.
- properly track and react to interruption and route change events. Previous implementation did track route changes during interruption, but did nothing with it. Currently it will properly mark engine and session as "needs reconfig", this is usefull when category change happened while engine was suspended
- properly track for "is the AVAudioEngine running?", previously during system interruption engine might get paused or tore down, resulting in audio context state returning wrong state ("running") which prevented the user of the lib from restarting the engine properly, additionally to context state engine (and all layers up to context) have its own `isPlaying` state or methods. 
- cleanup of some unnecessary/redundant code

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioDecoder.cpp
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioDecoder.cpp
@@ -38,7 +38,8 @@ std::shared_ptr<AudioBus> AudioDecoder::decodeWithFilePath(
       &decoder, buffer.data(), totalFrameCount, &framesDecoded);
 
   if (framesDecoded == 0) {
-    // __android_log_print(ANDROID_LOG_ERROR, "AudioDecoder", "Failed to decode");
+    // __android_log_print(ANDROID_LOG_ERROR, "AudioDecoder", "Failed to
+    // decode");
 
     ma_decoder_uninit(&decoder);
     return nullptr;
@@ -88,7 +89,8 @@ std::shared_ptr<AudioBus> AudioDecoder::decodeWithMemoryBlock(
       &decoder, buffer.data(), totalFrameCount, &framesDecoded);
 
   if (framesDecoded == 0) {
-    // __android_log_print(ANDROID_LOG_ERROR, "AudioDecoder", "Failed to decode");
+    // __android_log_print(ANDROID_LOG_ERROR, "AudioDecoder", "Failed to
+    // decode");
 
     ma_decoder_uninit(&decoder);
     return nullptr;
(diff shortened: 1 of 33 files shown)


packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioDecoder.cpp
```diff
@@ -38,7 +38,8 @@ std::shared_ptr<AudioBus> AudioDecoder::decodeWithFilePath(
       &decoder, buffer.data(), totalFrameCount, &framesDecoded);
 
   if (framesDecoded == 0) {
-    // __android_log_print(ANDROID_LOG_ERROR, "AudioDecoder", "Failed to decode");
+    // __android_log_print(ANDROID_LOG_ERROR, "AudioDecoder", "Failed to
```
michalsek, line 41: fu

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioDecoder.cpp
```diff
@@ -88,7 +89,8 @@ std::shared_ptr<AudioBus> AudioDecoder::decodeWithMemoryBlock(
...
```
michalsek, line 92: do we really need those logs?

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
```diff
@@ -45,10 +45,13 @@ bool AudioPlayer::openAudioStream() {
   return true;
 }
 
-void AudioPlayer::start() {
+bool AudioPlayer::start() {
   if (mStream_) {
     mStream_->requestStart();
```
michalsek: ideally this should check if requestStart is successful

michalsek: this is audio, don't assume anything

packages/react-native-audio-api/ios/audioapi/ios/core/NativeAudioPlayer.m
```diff
@@ -32,13 +32,15 @@ - (instancetype)initWithRenderAudio:(RenderAudioBlock)renderAudio
   return self;
 }
 
-- (void)start
+- (BOOL)start
```
michalsek: BOOl -> bool bool

maciejmakowski2003 (approved):
amazing work! thanks for fighting with this MONSTER!

PR #605, opened 2025-08-14
fix: disconnecting all nodes in manager

### LINKED ISSUE #603: Crash when trying to stop `AudioBufferSourceNode`
### Description

When I do something like this to stop `AudioBufferSourceNode`
```
const handlePlay = async () => {
  const audioContext = new AudioContext();

  const audioBuffer = await fetch(
    "https://software-mansion.github.io/react-native-audio-api/audio/music/example-music-01.mp3"
  )
    .then((response) => response.arrayBuffer())
    .then((arrayBuffer) => audioContext.decodeAudioData(arrayBuffer));

  const playerNode = audioContext.createBufferSource();
  playerNode.buffer = audioBuffer;

  playerNode.connect(audioContext.destination);
  playerNode.start(audioContext.currentTime);

  // do something...

  playerNode.disconnect();
};
```

This causes a crash on iOS (Android not tested).

From Xcode, I get the following error:
```
AURemoteIO::IOThread (37): EXC_BAD_ACCESS (code=1, address=0xbeadde8fc150)
A bad access to memory terminated the process.
```

It works correctly on version `0.7.0`, but crashes on `0.7.1-nightly-b6583fe-20250813`.
Not sure if this is related to the following fix: [(link if applicable).](https://github.com/software-mansion/react-native-audio-api/issues/598)

Please let me know if the provided example is not the best practice for using AudioBufferSourceNode.
Thanks again for the great library!```

### Steps to reproduce

Click Play Audio -> Click Stop Audio (check the minimum reproduce repo)

https://github.com/shinenic/react-native-audio-api-test/blob/main/App.js

0.7.1-nightly-b6583fe-20250813

### React Native version

0.79.5

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo Dev Client

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Device model

ios simulator

### Acknowledgements

Yes

mdydek:
Closes #603

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- added copying outputnodes map, because disconnecting nodes remove elements from map under the hood

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
@@ -72,7 +72,8 @@ void AudioNodeManager::settlePendingConnections() {
       assert(to != nullptr);
       from->disconnectNode(to);
     } else {
-      for (auto &node : from->outputNodes_) {
+      auto outputNodesCopy = from->outputNodes_;
+      for (auto &node : outputNodesCopy) {
         from->disconnectNode(node);
       }
     }
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
@@ -72,9 +72,10 @@ void AudioNodeManager::settlePendingConnections() {
       assert(to != nullptr);
       from->disconnectNode(to);
     } else {
-      auto outputNodesCopy = from->outputNodes_;
-      for (auto &node : outputNodesCopy) {
-        from->disconnectNode(node);
+      for (auto it = from->outputNodes_.begin(); it != from->outputNodes_.end();) {
+        auto next = std::next(it);
+        from->disconnectNode(*it);
+        it = next;
       }
     }
   }


packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
```diff
@@ -72,7 +72,8 @@ void AudioNodeManager::settlePendingConnections() {
       assert(to != nullptr);
       from->disconnectNode(to);
     } else {
-      for (auto &node : from->outputNodes_) {
+      auto outputNodesCopy = from->outputNodes_;
+      for (auto &node : outputNodesCopy) {
```
maciejmakowski2003: can we switch to iterator based loop? it is highly irresponsible to use for-range loop that invokes function that removes under the hood

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
```diff
@@ -72,8 +72,10 @@ void AudioNodeManager::settlePendingConnections() {
       assert(to != nullptr);
       from->disconnectNode(to);
     } else {
-      for (auto &node : from->outputNodes_) {
-        from->disconnectNode(node);
+      for (auto it = from->outputNodes_.begin(); it != from->outputNodes_.end();) {
```
maciejmakowski2003, line 75: it would be more efficient to do it like below

for (auto it = sourceNodes_.begin(), end = sourceNodes_.end(); it != end;
       ++it) {
    //sth
  }

maciejmakowski2003 (changes requested):

PR #616, opened 2025-08-18
Feat/improved performance and stability of AudioNodeManager

poneciak57:
# Description
Implemented blazingly fast single-producer single-consumer channel.
Performed benchmarks of the channel itself
Implemented AudioNodeManager and NodeDestructor that uses performant channel impl
Performed benchmarks on mobile devices, emulators and simulators

# Method
The benchmark was conducted by calculating the average function execution time over 5 minutes.
- Used drums mode
- Song groove 1
- Duration 5 minutes

# SPSC benchmarks
Other spsc channel benchmarks can be found [here](https://github.com/poneciak57/channels-cpp)

> [!NOTE]
> For comparison, for a typical sample rate (44.1 kHz), the time between two render quantum averaging 128 frames is 2.9 ms = 2_900_000 ns

# iOS Simulator Results
> Device: iPhone 16 Pro
- Main branch - 7k nanoseconds
- Improved SPSC - 4k nanoseconds
- Improvement% - 42.86%
- Improvementx - 1.75x

# Android Emulator Results
> Device: Android 16 (baklava) Arm64
- Main branch - 650k nanoseconds
- Improved SPSC - 6k nanoseconds
- Improvement% - 99.08%
- Improvementx - 108.33x

# iOS Device Results
> Device: Szczepan Złoty (old ios)
- Main branch - 24k nanoseconds
- Improved SPSC - 17k nanoseconds
- Improvement% - 29.17%
- Improvementx - 1.41x (newer devices tend to have better atomic operations)

# Android Device Results
> Device: Samsung A50
- Main branch - 20k ns with very high spikes even up to 200k - 500k
- Improved SPSC - stable 6-7k
- Improvement% - 65%
- Improvementx - 2.86x up to 90x

# Summary
The main improvement on Android is stability and performance.
On iOS, smaller but still significant time improvement.

Closes #

## ⚠️ Breaking changes ⚠️

- significantly improved android performance and stability

## Introduced changes

- improved overall performance and stability

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.h
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.h
@@ -102,11 +102,11 @@ class AudioNodeManager {
 
   /// @brief Initial capacity for various node types for deletion
   /// @note Higher capacity decreases number of reallocations at runtime (can be easily adjusted to 128 if needed)
-  const size_t kInitialCapacity = 32;
+  static constexpr size_t kInitialCapacity = 32;
 
   /// @brief Initial capacity for event passing channel
   /// @note High value reduces wait time for sender (JavaScript/HostObjects thread here)
-  const size_t kChannelCapacity = 1024;
+  static constexpr size_t kChannelCapacity = 1024;
 
   std::vector<std::shared_ptr<AudioScheduledSourceNode>> sourceNodes_;
   std::vector<std::shared_ptr<AudioNode>> processingNodes_;
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeDestructor.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeDestructor.cpp
@@ -9,9 +9,9 @@ AudioNodeDestructor::AudioNodeDestructor() {
   auto [sender, receiver] = channels::spsc::channel<
       std::shared_ptr<AudioNode>,
       channels::spsc::OverflowStrategy::WAIT_ON_FULL,
-      channels::spsc::WaitStrategy::ATOMIC_WAIT>(1024);
+      channels::spsc::WaitStrategy::ATOMIC_WAIT>(kChannelCapacity);
   sender_ = std::move(sender);
-  thread_ =
+  workerHandle_ =
       std::thread(&AudioNodeDestructor::process, this, std::move(receiver));
 }
 
@@ -20,8 +20,8 @@ AudioNodeDestructor::~AudioNodeDestructor() {
 
   // We need to send a nullptr to unblock the receiver
   sender_.send(nullptr);
-  if (thread_.joinable()) {
-    thread_.join();
+  if (workerHandle_.joinable()) {
+    workerHandle_.join();
   }
 }
 
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeDestructor.h
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeDestructor.h
@@ -27,7 +27,9 @@ class AudioNodeDestructor {
   bool tryAddNodeForDeconstruction(std::shared_ptr<AudioNode> &&node);
 
  private:
-  std::thread thread_;
+  static constexpr size_t kChannelCapacity = 1024;
+
+  std::thread workerHandle_;
   std::atomic<bool> isExiting_;
 
   channels::spsc::Sender<
--- a/packages/react-native-audio-api/common/cpp/test/CMakeLists.txt
+++ b/packages/react-native-audio-api/common/cpp/test/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.14)
 project(rnaudioapi_test)
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(ROOT ${CMAKE_SOURCE_DIR}/../../../../..)
 set(REACT_NATIVE_DIR "${ROOT}/node_modules/react-native")


packages/react-native-audio-api/common/cpp/audioapi/utils/SpscChannel.hpp
```diff
@@ -0,0 +1,396 @@
+#pragma once
+
+#include <atomic>
+#include <memory>
+#include <algorithm>
+#include <thread>
+#include <type_traits>
+
+
+namespace channels::spsc {
```
mdydek: this namespace would suggest that it is something outside of our library, even though it isn't let's follow the same convention like in f.e. fft. My proposition is audioapi::spsc.

packages/react-native-audio-api/common/cpp/audioapi/utils/SpscChannel.hpp
```diff
@@ -0,0 +1,396 @@
...
+        if constexpr (Strategy == OverflowStrategy::WAIT_ON_FULL) {
+            return try_send_wait_on_full(std::forward<U>(value));
+        } else {
+            return try_send_overwrite_on_full(std::forward<U>(value));
+        }
+    }    
+    
+    /// @brief Try to receive a value from the channel
+    /// @param value The variable to store the received value
+    /// @return True if the value was received successfully, false if the channel is empty
```
mdydek: it returns `ResponseStatus` not `bool`

packages/react-native-audio-api/common/cpp/audioapi/utils/SpscChannel.hpp
```diff
@@ -0,0 +1,396 @@
...
+    /// @note This function is lock-free and wait-free
+    ResponseStatus try_receive(T& value) noexcept(std::is_nothrow_move_assignable_v<T> && std::is_nothrow_destructible_v<T>) {
+        if constexpr (Strategy == OverflowStrategy::OVERWRITE_ON_FULL) {
+            // Set reader active flag to prevent overwrites during read
+            bool isOccupied = oldestOccupied_.exchange(true, std::memory_order_acq_rel);
```
mdydek, line 252: from higher level it looks like `isOccupied` is set to true and then checked in if, can you elaborate more why is it like that?

packages/react-native-audio-api/common/cpp/audioapi/utils/SpscChannel.hpp
```diff
@@ -0,0 +1,396 @@
...
+
+template<typename T, OverflowStrategy Strategy, WaitStrategy Wait>
+class Sender;
+template<typename T, OverflowStrategy Strategy, WaitStrategy Wait>
+class Receiver;
+template<typename T, OverflowStrategy Strategy, WaitStrategy Wait>
+class InnerChannel;
+
+/// @brief Create a bounded single-producer, single-consumer channel
+/// @param capacity The minimum capacity of the channel, real capacity will be equal to the closest higher or equal power of two - 1. So for example, if capacity = 12 then channel will hold 15 elements.
```
mdydek, line 59: why real would be 15 not 16, you use bit mask under the hood set to next power of two - 1, which would cause valid indexes from 0, 2^n - 1 and it would have 2^n elements obviously that way

packages/react-native-audio-api/common/cpp/audioapi/utils/SpscChannel.hpp
```diff
@@ -0,0 +1,396 @@
...
+    inline size_t next_index(const size_t val) const noexcept {
+        return (val + 1) & capacity_mask_;
+    }
+
+    const size_t capacity_;
+    const size_t capacity_mask_; // mask for bitwise next_index
+    T* buffer_;
+
+    /// Producer-side data (accessed by sender thread)
+    alignas(64) std::atomic<size_t> sendCursor_{0};
```
mdydek, line 382: elaborate more here also use case for `alignas(64)`

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.h
```diff
@@ -16,53 +18,105 @@ class AudioParam;
...
 
-  std::vector<std::tuple<
-      std::shared_ptr<AudioNode>,
-      std::shared_ptr<AudioParam>,
-      ConnectionType>>
-      audioParamToConnect_;
+  channels::spsc::Sender<
+    std::unique_ptr<Event>,
+    channels::spsc::OverflowStrategy::WAIT_ON_FULL,
+    channels::spsc::WaitStrategy::BUSY_LOOP> sender_;
```
mdydek: can we have different overflow strategies and wait strategies? If that is no desired behavior, some protection against mis-use would be valuable

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
```diff
@@ -14,132 +84,155 @@ void AudioNodeManager::addPendingNodeConnection(
...
-  for (auto it = audioParamToConnect_.begin(), end = audioParamToConnect_.end();
-       it != end;
-       ++it) {
-    std::shared_ptr<AudioNode> from = std::get<0>(*it);
-    std::shared_ptr<AudioParam> to = std::get<1>(*it);
-    ConnectionType type = std::get<2>(*it);
+void AudioNodeManager::handleConnectEvent(std::unique_ptr<Event> event) {
+  assert(event->payloadType == EventPayloadType::NODES);
+  event->payload.nodes.from->connectNode(event->payload.nodes.to);
+}
```
mdydek: this drops possibility of connecting params, cause assert would not pass it

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
```diff
@@ -14,132 +84,155 @@ void AudioNodeManager::addPendingNodeConnection(
...
 
-    assert(from != nullptr);
-    assert(to != nullptr);
+void AudioNodeManager::handleDisconnectEvent(std::unique_ptr<Event> event) {
```
mdydek, line 179: same issue with params here

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
```diff
@@ -14,132 +84,155 @@ void AudioNodeManager::addPendingNodeConnection(
...
+  assert(event->payloadType == EventPayloadType::NODES);
+  event->payload.nodes.from->disconnectNode(event->payload.nodes.to);
+}
 
-    if (type == ConnectionType::CONNECT) {
-      from->connectParam(to);
-    } else {
-      from->disconnectParam(to);
-    }
+void AudioNodeManager::handleDisconnectAllEvent(std::unique_ptr<Event> event) {
```
mdydek, line 189: and here (params)

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
```diff
@@ -14,132 +84,155 @@ void AudioNodeManager::addPendingNodeConnection(
...
+        (sNodesIt->get()->isUnscheduled() || sNodesIt->get()->isFinished())) {
+      (*sNodesIt)->cleanup();
+      if (nodeDeconstructor_.tryAddNodeForDeconstruction(*sNodesIt)) {
         sNodesIt = sourceNodes_.erase(sNodesIt);
       } else {
         ++sNodesIt;
       }
+    } else {
+      ++sNodesIt;
     }
```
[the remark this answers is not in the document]
poneciak57: not exacly

packages/react-native-audio-api/common/cpp/audioapi/utils/SpscChannel.hpp
poneciak57: ok will change that. I work on this in different folder and just copy a file after testing so forgot to change namespace

packages/react-native-audio-api/common/cpp/audioapi/utils/SpscChannel.hpp
poneciak57: yeah it has 16 elements but the senderCursor needs to be one behind to easily mark channel as full. So yeah it takes space of 16 elements but it can store up to one less

packages/react-native-audio-api/common/cpp/audioapi/utils/SpscChannel.hpp
poneciak57: `alignas(64)` aligns an object to a 64-byte boundary (not 64 bits). Since modern CPUs usually have 64-byte cache lines, this prevents two objects from accidentally sharing the same cache line. If two CPUs modify data in the same cache line, even in different variables, this causes "false sharing," leading to extra cache misses and slower performance.

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.h
poneciak57: yeah we can but they are specifically choosen here for performance reasons. we do not want override because we do not want to lose data here and it reducs performance a little bit. It is hard to protect agains missues all strategies are well described and default ones are most performant but explicit usage is better in my view

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
poneciak57: good catch. I must have missed that.

packages/react-native-audio-api/common/cpp/audioapi/utils/SpscChannel.hpp
poneciak57: exhange returns previous value held by atomic

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
poneciak57: Disconnect all, disconnects only nodes.

packages/react-native-audio-api/common/cpp/audioapi/utils/SpscChannel.hpp
poneciak57: This is main reason why my implemenation outperforms boost's in benchmarks

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.cpp
```diff
@@ -14,132 +84,165 @@ void AudioNodeManager::addPendingNodeConnection(
...
+    auto next = std::next(it);
+    event->payload.nodes.from->disconnectNode(*it);
+    it = next;
+  }
 }
 
-void AudioNodeManager::cleanupNode(const std::shared_ptr<AudioNode> &node) {
-  nodeDeconstructor_.addNodeForDeconstruction(node);
-  node->cleanup();
+void AudioNodeManager::handleAddEvent(std::unique_ptr<Event> event) {
```
maciejmakowski2003:
```suggestion
void AudioNodeManager::handleAddToDeconstructionEvent(std::unique_ptr<Event> event) {
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.h
```diff
@@ -16,53 +18,105 @@ class AudioParam;
...
   void addAudioParam(const std::shared_ptr<AudioParam> &param);
 
   void cleanup();
 
  private:
-  std::mutex graphLock_;
   AudioNodeDestructor nodeDeconstructor_;
 
   // all nodes created in the context
   std::unordered_set<std::shared_ptr<AudioScheduledSourceNode>> sourceNodes_;
```
maciejmakowski2003: let's switch to std::vector

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioNodeManager.h
mdydek: but do we really need to explicitly set them for receiver and sender, can't we put overflow strategy and wait strategy in class and at the end call
`  channels::spsc::Receiver<
    std::unique_ptr<Event>,
    overflowStrategy,
    waitStrategy> receiver_;
`

maciejmakowski2003 (approved):
I delved into the implementation and I am a big fan of this approach. great work!

PR #644, opened 2025-09-03
Fix/android foreground service

maciejmakowski2003:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- Fixed android Foreground Service management

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.h
```diff
@@ -5,12 +5,15 @@
...
 
+#include <audioapi/android/core/NativeAudioPlayer.hpp>
+
 namespace audioapi {
 
 using namespace oboe;
 
 class AudioContext;
 class AudioBus;
+class NativeAudioPlayer;
```
mdydek: I don't think that you need forward declaration if you include header for `NativeAudioPlayer` class

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.h
```diff
@@ -44,6 +52,8 @@ class AudioPlayer : public AudioStreamDataCallback, AudioStreamErrorCallback {
   int channelCount_;
 
   bool openAudioStream();
+
+  jni::global_ref<NativeAudioPlayer> nativeAudioPlayer_;
```
mdydek: this is the way to have reference in c++ for java/kotlin class?

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/NativeAudioPlayer.hpp
```diff
@@ -0,0 +1,36 @@
...
+#include <unordered_map>
+
+namespace audioapi {
+
+using namespace facebook;
+using namespace react;
+
+class NativeAudioPlayer : public jni::JavaClass<NativeAudioPlayer> {
+ public:
+  static auto constexpr kJavaDescriptor =
```
mdydek, line 18: is this some kind of magic variable needed by jni, because I don't see any usage of this outside of this file?

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.h
```diff
@@ -19,13 +22,18 @@ class AudioPlayer : public AudioStreamDataCallback, AudioStreamErrorCallback {
       float sampleRate,
       int channelCount);
 
+  ~AudioPlayer() override {
+    nativeAudioPlayer_.release();
```
mdydek, line 25: this makes disconnection between cpp and java class, right?

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/MediaNotificationManager.kt
```diff
@@ -182,45 +168,30 @@ class MediaNotificationManager(
...
-      super.onCreate()
-      try {
-        notification =
-          MediaSessionManager.mediaNotificationManager
-            .prepareNotification(NotificationCompat.Builder(this, MediaSessionManager.CHANNEL_ID), false)
-        startForeground(MediaSessionManager.NOTIFICATION_ID, notification)
-      } catch (ex: Exception) {
-        Log.w("AudioManagerModule", "Error starting service: ${ex.message}")
+    private var isServiceStarted = false
+    private val serviceLock = Any()
```
mdydek, line 185: why any?

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/MediaSessionManager.kt
```diff
@@ -106,17 +82,59 @@ object MediaSessionManager {
...
+    val uuid = UUID.randomUUID().toString()
+    nativeAudioPlayers[uuid] = player
 
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-      try {
-        this.reactContext.get()?.bindService(myIntent, connection, Context.BIND_AUTO_CREATE)
-      } catch (ignored: Exception) {
-        ContextCompat.startForegroundService(this.reactContext.get()!!, myIntent)
+    return uuid
+  }
```
mdydek: Naming of this function is kind of confusing, `attachSourceNode` and parameter is named `player`

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/NativeAudioPlayer.hpp
maciejmakowski2003: yeah, fbjni uses it

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.h
maciejmakowski2003: yeah

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.h
maciejmakowski2003: thats right

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/MediaSessionManager.kt
maciejmakowski2003: I am trying to be stick to ios convention

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/MediaNotificationManager.kt
maciejmakowski2003: ```The synchronized block doesn't care if the object is a complex class or a simple one. It only needs a reference to a single, shared object in memory to act as a "key." Only one thread can "hold the key" (i.e., be inside the synchronized block for that object) at a time.```

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/MediaSessionManager.kt
maciejmakowski2003: will fix it

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioContextHostObject.h
poneciak57, line 1: Here i am unsure about thread safety of this. I am unsure how does the promise get resolved. If it is resolved instantly on the javascript site it is completly fine but if it can resolve after any of these methods call we might have a problem. 
So it is wrong if the the promise is not resolved instantly. (I checked it is scheduled asynchronously so it might resolve later on)
Setup:
User calls close(); 
and for example creates new context and overwrites old one.
Anyway old handle to context is somehow lost (pretty possible)

What bad happens?
There is very little possibility of `use after free`. If the promise will not get executed before the garbage collection of the ContextHostObject, `this` will point to invalid memory.

Correct me if i am wrong this will rarelly happen but the outcome would be crash and not working close for example.

Possible solutions:
1. If you can execute it in sync you can just return some promise on typescript side and here just write blocking function
2. If you want to keep the promise (javscript promise) capture shared_ptr of audio context in lambda

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/MediaSessionManager.kt
```diff
@@ -106,17 +82,59 @@ object MediaSessionManager {
...
+  private fun startForegroundService() {
+    synchronized(serviceStateLock) {
+      if (isServiceRunning || reactContext.get() == null) {
+        return
       }
-    } else {
-      this.reactContext.get()?.startService(myIntent)
+
+      val intent = Intent(reactContext.get(), MediaNotificationManager.AudioForegroundService::class.java)
+      intent.action = "START_FOREGROUND"
```
poneciak57: maybe use an enum for that. If you need a string representation you can create enum with string representation in java. (should be possible in kotlin too)

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/MediaSessionManager.kt
```diff
@@ -106,17 +82,59 @@ object MediaSessionManager {
...
+  }
+
+  private fun stopForegroundService() {
+    synchronized(serviceStateLock) {
+      if (!isServiceRunning || reactContext.get() == null) {
+        return
+      }
+
+      val intent = Intent(reactContext.get(), MediaNotificationManager.AudioForegroundService::class.java)
+      intent.action = "STOP_FOREGROUND"
```
poneciak57: same maybe use enum

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/MediaNotificationManager.kt
```diff
@@ -229,26 +200,40 @@ class MediaNotificationManager(
...
+        "START_FOREGROUND" -> startForegroundService()
+        "STOP_FOREGROUND" -> stopForegroundService()
+        else -> startForegroundService()
+      }
+
       return START_NOT_STICKY
     }
 
+    private fun stopForegroundService() {
+      synchronized(serviceLock) {
```
poneciak57, line 238: is this synchronization necessery? Can you call stop and start at the same time?

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.h
poneciak57: +1

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/MediaNotificationManager.kt
```diff
@@ -182,45 +168,30 @@ class MediaNotificationManager(
...
+            notification =
+              MediaSessionManager.mediaNotificationManager
+                .prepareNotification(
+                  NotificationCompat.Builder(this, MediaSessionManager.CHANNEL_ID),
+                  false,
+                )
+            startForeground(MediaSessionManager.NOTIFICATION_ID, notification)
+            isServiceStarted = true
+          } catch (ex: Exception) {
+            Log.w("AudioManagerModule", "Error starting foreground service: ${ex.message}")
```
poneciak57: if thats an error use `Log.e` instead
```suggestion
            Log.e("AudioManagerModule", "Error starting foreground service: ${ex.message}")
```

poneciak57 (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioContextHostObject.h
maciejmakowski2003: I will go with the second one

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/MediaNotificationManager.kt
maciejmakowski2003: yeah, cause it can be invoked by system thread and control thread

vltkv (approved):
LGTM

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.h
poneciak57: ++1

maciejmakowski2003: done

poneciak57 (approved):

PR #652, opened 2025-09-05
fix: number of channels

### LINKED ISSUE #650: .m4a audio file played at twice the expected speed
### Description

When playing back an audio file recorded as a .m4a, the audio is played at twice the expected speed. When played on another player (e.g. VLC on a computer), the file is played at normal speed.

It looks like the problem is linked to how the file is decoded/loaded into the AudioBuffer: the AudioBuffer duration says that the audio is 13 sec long, while it is actually 26 sec long.

The audio file was recorded with a sample rate of 48000 and a pcm-16 encoding, and compressed using AAC with a bitrate of 128000.
The AudioContext used for playback uses default settings.

### Steps to reproduce

1. Download the file: https://drive.google.com/uc?export=download&id=136ohXTHKr00mURjHHXXbj_-Jt-2KVR25
2. Load it and play it with react-native-audio-api, and compare with how it's played with another player (like VLC).

https://snack.expo.dev/@dominique.lebrun/hot-green-ramen

0.8.0

### React Native version

0.79.6

### Platforms

Android

### JavaScript runtime

None

### Workflow

Expo Dev Client

### Architecture

None

### Build type

Release app & dev bundle

### Device

Real device

### Device model

One Plus Nord 2T

### Acknowledgements

Yes

mdydek:
Closes #650 

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- now passing channel number from decoder to decoding pipeline in ffmpeg to always align with desired number of channels

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/libs/ffmpeg/FFmpegDecoding.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/libs/ffmpeg/FFmpegDecoding.cpp
@@ -208,7 +208,7 @@ std::vector<int16_t> readAllPcmFrames(
   return buffer;
 }
 
-std::vector<int16_t> decodeWithMemoryBlock(const void *data, size_t size, int sample_rate) {
+std::vector<int16_t> decodeWithMemoryBlock(const void *data, size_t size, const int channel_count, int sample_rate) {
     if (data == nullptr || size == 0) {
         return {};
     }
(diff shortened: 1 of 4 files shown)


packages/react-native-audio-api/common/cpp/audioapi/libs/ffmpeg/FFmpegDecoding.cpp
```diff
@@ -329,7 +329,7 @@ std::vector<int16_t> decodeWithMemoryBlock(const void *data, size_t size, int sa
     return decoded_buffer;
 }
 
-std::vector<int16_t> decodeWithFilePath(const std::string &path, int sample_rate) {
+std::vector<int16_t> decodeWithFilePath(const std::string &path, const int channel_count, int sample_rate) {
```
maciejmakowski2003, line 332: so currently we are forcing ffmpeg to return decoded data with given amount of channels, right?

mdydek: yes, in the previous implementation, channels number was same as internal audio, but right now eveyrthing is forced to channel count from `AudioDecoder.h`

PR #678, opened 2025-09-18
fix: better memory handling

mdydek:
Closes RNAA-251
Because of behavior of objective-c blocks some buffers was destroyed only after stopping recording even though they were not needed, by anything. Second issue was that we were passing audio buffer host object from c++ without setting its memory pressure, which made it not garbage collected by js in reasonable moment.

## ⚠️ Breaking changes ⚠️

- removed time from microphone callback

## Introduced changes

- added `@autoreleasepool` which marks scope for better handling of garbage collecting
- added memory pressure to audio buffer host object that is moved in event

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/ios/audioapi/ios/core/NativeAudioRecorder.m
+++ b/packages/react-native-audio-api/ios/audioapi/ios/core/NativeAudioRecorder.m
@@ -55,37 +55,40 @@
   float outputSampleRate = self.outputFormat.sampleRate;
 
   if (inputSampleRate != outputSampleRate) {
+    AVAudioPCMBuffer *inputBuffer = [[AVAudioPCMBuffer alloc] initWithPCMFormat:self.inputFormat
+                                                                  frameCapacity:frameCount];
+    memcpy(
+        inputBuffer.mutableAudioBufferList->mBuffers[0].mData,
+        inputData->mBuffers[0].mData,
+        inputData->mBuffers[0].mDataByteSize);
+    inputBuffer.frameLength = frameCount;
+
+    int outputFrameCount = frameCount * outputSampleRate / inputSampleRate;
+
+    AVAudioPCMBuffer *outputBuffer = [[AVAudioPCMBuffer alloc] initWithPCMFormat:self.audioConverter.outputFormat
+                                                                   frameCapacity:outputFrameCount];
+
+    NSError *error = nil;
+    AVAudioConverterInputBlock inputBlock =
+        ^AVAudioBuffer *_Nullable(AVAudioPacketCount inNumberOfPackets, AVAudioConverterInputStatus *outStatus)
+    {
+      *outStatus = AVAudioConverterInputStatus_HaveData;
+      return inputBuffer;
+    };
+
+    /// IMPORTANT: AVAudioConverter leaks memory without autorelease pool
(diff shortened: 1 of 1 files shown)


mdydek:
Be aware that this fixes some part of problematic behavior, memory still rises during long recording session, but the rise is significantly less steep. Probably needs more insight if the issue is still from our side

packages/react-native-audio-api/ios/audioapi/ios/core/NativeAudioRecorder.m
```diff
@@ -54,44 +54,43 @@ - (OSStatus)processAudioInput:(const AudioBufferList *)inputData
...
-      return inputBuffer;
-    };
-
-    [self.audioConverter convertToBuffer:outputBuffer error:&error withInputFromBlock:inputBlock];
-
-    if (error) {
-      NSLog(@"Error during audio conversion: %@", error.localizedDescription);
-      return kAudioServicesBadSpecifierSizeError;
+    @autoreleasepool {
+      AVAudioPCMBuffer *inputBuffer = [[AVAudioPCMBuffer alloc] initWithPCMFormat:self.inputFormat
```
poneciak57: This is not an optimal or correct solution.

Some objects in objective-C are autorelase means that they will be garbage collected. But they might not be garbage collected on an audio thread created by AVAudioEngine which caused an issue here. 

What happens when you wrap a block of code in `@autoreleasepool`? Using this pool makes sure that we forcefully release that memory.

## Why is it bad?
According to [this](https://developer.apple.com/library/archive/documentation/MusicAudio/Conceptual/AudioUnitProgrammingGuide/TheAudioUnit/TheAudioUnit.html) apple documentation

> Processing is the most computationally expensive part of an audio unit’s life cycle. Within the 
> processing loop, avoid the following actions:
> - Mutually exclusive (mutex) resource locking
> - Memory or resource allocation

So it might be the case that the thread that processes this code block avoids dealocation at all cost.

## How it can be fixed other way around?
You can try to remove all allocations out of the processing loop by for example allocating them during initialization/preparation and making them persistent across multiple calls to the callback.

## Addition
This solution should be completly safe as it ensures proper dealocation but it can cause more frequent audio dropouts as it does heavy tasks on processing callback.

poneciak57 (changes requested):

packages/react-native-audio-api/ios/audioapi/ios/core/NativeAudioRecorder.m
mdydek: `So it might be the case that the thread that processes this code block avoids dealocation at all cost.` It is not true, dealocations of `outputBuffer` are as soon as it exits the scope, the issue is that `inputBuffer` ones is not deallocated before stopping recording. Also prealocation is not possible, because we need `frameCount`, which we get from the system itself inside the processing loop.

poneciak57: Dont we know any upper bound? To maybe prealocate bigger buffer.

mdydek: We could be “estimating” this value, but I think that you agree that it is not a good idea

PR #683, opened 2025-09-22
fix: removed cppFlags from build.gradle

maciejmakowski2003:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- Removed cppFlags from `build.gradle` becasue:
image

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


poneciak57 (changes requested):
Have you checked if `-O2` flag is being set when building in release mode after that change? If so it looks good to me, but it is important as we do not have it elsewhere.

PR #685, opened 2025-09-22
Docs improvements

vltkv:
Closes #

## ⚠️ Breaking changes ⚠️

- 

## Introduced changes

- unify constructor descriptions
- move playgrounds to the beginning of the page

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/audiodocs/docs/core/audio-context.mdx
```diff
@@ -11,7 +11,7 @@ It is responsible for supervising and managing audio-processing graph.
 
 `new AudioContext()`
 
-[`new AudioContext(options: AudioContextOptions)`](/docs/core/audio-context#audiocontextoptions)
+[`new AudioContext(sampleRate)`](/docs/core/audio-context#audiocontextoptions)
```
poneciak57: Thats not true, might be misleading

packages/audiodocs/docs/core/offline-audio-context.mdx
```diff
@@ -10,7 +10,7 @@ Instead, it processes the audio as quickly as possible and outputs the result to
 
 
 ## Constructor
-[`OfflineAudioContext(options)`](/docs/core/offline-audio-context#offlineaudiocontextoptions)
+[`OfflineAudioContext(numberOfChannels, length, sampleRate)`](/docs/core/offline-audio-context#offlineaudiocontextoptions)
```
poneciak57: Thats not true, might be misleading

packages/audiodocs/docs/inputs/audio-recorder.mdx
```diff
@@ -12,7 +12,7 @@ It connects to the audio graph through a [`RecorderAdapterNode`](/docs/sources/r
 
 ## Constructor
 
-[`AudioRecorder(options)`](/docs/inputs/audio-recorder#audiorecorderoptions)
+[`AudioRecorder(sampleRate, bufferLengthInSamples)`](/docs/inputs/audio-recorder#audiorecorderoptions)
```
poneciak57: Also not true

packages/audiodocs/docs/sources/audio-buffer-source-node.mdx
```diff
@@ -10,22 +10,30 @@ import { useAudioBufferSourcePlayground } from '@site/src/components/Interactive
...
 #### [`AudioNode`](/docs/core/audio-node#properties) properties
 
 <AudioNodePropsTable numberOfInputs={0} numberOfOutputs={1} channelCount={"defined by associated buffer"} channelCountMode={"max"} channelInterpretation={"speakers"} />
 
 ## Constructor
 
-[`BaseAudioContext.createBufferSource(options)`](/docs/core/base-audio-context#createbuffersource)
+`BaseAudioContext.createBufferSource()`
+
+[`BaseAudioContext.createBufferSource(pitchCorrection)`](/docs/core/base-audio-context#createbuffersource)
```
poneciak57: Not exacly true, there is no info about the options tho

poneciak57 (changes requested):
I don't feel like these changes are apropriate. I understand that having a link to input options type might be something creating concerns about docs usability. Hovewer correctness is more important. Maybe instead of writing it like this:
```
AudioContext(sampleRate) // incorrect
```
```
AudioContext({sampleRate}) // implies that it is an object with property of sampleRate
```
But i still think having reference to options class would be better or we can change our api interface and keep it like this. (i do not recommend changing api)

packages/audiodocs/docs/core/audio-context.mdx
poneciak57: here we can add link only around word `AudioContextOptions`

vltkv: I based my approach on how some constructors are currently documented (e.g. AudioBuffer), so I thought it was fine. My main concern is usability - right now users need to click through multiple places to figure out constructor options (some of them, like pitchCorrection, are especially hard to find).
I’d be in favor of either:
•  using your suggestion AudioContext({ sampleRate }), or at least
•  moving the options object definitions directly under the constructor, so all necessary information will be in one place

michalsek: Also there is aditional param to init the audio context as suspended:
```ts
interface AudioContextOptions {
  sampleRate?: number;
  initSuspended?: boolean;
}
```

poneciak57 (approved):

michalsek (approved):

PR #691, opened 2025-09-25
feat(android): support background recording

yzlin:
## Introduced changes

When launching audio recorder on Android, the foreground service didn't start. This PR leverages the same foreground service `AudioPlayer` used for `AudioRecorder` so that it can support background recording

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/core/NativeAudioRecorder.kt
```diff
@@ -0,0 +1,24 @@
+package com.swmansion.audioapi.core
+
+import com.facebook.common.internal.DoNotStrip
+import com.swmansion.audioapi.system.MediaSessionManager
+
+@DoNotStrip
+class NativeAudioRecorder {
+  private var sourceNodeId: String? = null
```
maciejmakowski2003:
```suggestion
  private var inputNodeId: String? = null
```

PR #693, opened 2025-09-26
Feat: Three.js Vinyl player example

miloszwielgus:
## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- created a 3d vinyl record example for the `BiquadFilterNode` in the docs

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/audiodocs/src/components/RecordPlayerExample/VinylAnimation.tsx
+++ b/packages/audiodocs/src/components/RecordPlayerExample/VinylAnimation.tsx
@@ -1,17 +1,10 @@
-import React, { useRef, useEffect } from 'react';
+import React, { useRef, useEffect, useMemo, useState } from 'react';
 import * as THREE from 'three';
 import { OrbitControls } from 'three/examples/jsm/controls/OrbitControls.js';
 import { RoundedBoxGeometry } from 'three/examples/jsm/geometries/RoundedBoxGeometry.js';
 import { LineSegments2 } from 'three/examples/jsm/lines/LineSegments2.js';
 import { LineSegmentsGeometry } from 'three/examples/jsm/lines/LineSegmentsGeometry.js';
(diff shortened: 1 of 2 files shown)


packages/audiodocs/src/components/RecordPlayerExample/VinylAnimation.tsx
```diff
@@ -0,0 +1,583 @@
...
+    coverHeight: 0,
+    wasOnGround: true,
+    fallVelocity: 0,
+  });
+
+  // Scene object refs
+  const sceneRefsRef = useRef<SceneRefs | null>(null);
+  
+  // Audio refs
+  const audioRefsRef = useRef<AudioRefs | null>(null);
```
michalsek: Lets start removing the useRef madness from audio-api slowly :) 

`AudioContext` can be stored globaly outside of the component or through useMemo hook

```ts
const context = useMemo(() => new AudioContext()); 
```

which makes it easier to use, as well as it is a more react-way

michalsek: AudioBuffer can be safely stored in useState, and I'm pretty sure we can do the same for Nodes or alternatively we can create simple AudioService outside react, to manage the state without keeping everything in refs.

michalsek (changes requested):

packages/audiodocs/src/components/RecordPlayerExample/VinylAnimation.tsx
miloszwielgus: I used a useMemo to create the `AudioContext` as well as the effect nodes, the AudioBuffer is now stored in a useState and the bufferSource is still used through a ref, is the=is the correct approach?

PR #699, opened 2025-09-29
feat: async promise resolving

poneciak57:
Implemented method that simplifies creating promises that are heavy and should be resolved on separate thread. It also creates a room for adding thread pool in a future.

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


maciejmakowski2003 (approved):

maciejmakowski2003:
lets use it in all places creating another thread inside promise

packages/react-native-audio-api/common/cpp/audioapi/jsi/JsiPromise.cpp
mdydek, line 1: wouldn't it be good idea to also refactor synchronous promise creation to move semantics and maybe refactor it slightly so approach is very similar in both of these functions?

PR #710, opened 2025-10-01
feat: added onLoopEnded callback

maciejmakowski2003:
Closes RNAA-226

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- Added `onLoopEnded` event to `AudioBufferSourceNode`

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferSourceNode.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferSourceNode.cpp
@@ -29,6 +29,8 @@ AudioBufferSourceNode::~AudioBufferSourceNode() {
 
   buffer_.reset();
   alignedBus_.reset();
+
+  clearOnLoopEndedCallback();
 }
 
 bool AudioBufferSourceNode::getLoop() const {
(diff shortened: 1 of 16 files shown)


packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferSourceNode.cpp
```diff
@@ -150,6 +167,13 @@ double AudioBufferSourceNode::getCurrentPosition() const {
       static_cast<int>(vReadIndex_), buffer_->getSampleRate());
 }
 
+void AudioBufferSourceNode::sendOnLoopEndedEvent() {
+  if (onLoopEndedCallbackId_ != 0) {
+    context_->audioEventHandlerRegistry_->invokeHandlerWithEventBody(
+        "loopEnded", onLoopEndedCallbackId_, {});
+  }
+}
+
```
poneciak57, line 178:
```suggestion
void AudioBufferSourceNode::sendOnLoopEndedEvent() {
  auto onLoopEndedCallbackId = onLoopEndedCallbackId_.load(std::memory_order_acquire);
  if (onLoopEndedCallbackId != 0) {
    context_->audioEventHandlerRegistry_->invokeHandlerWithEventBody(
        "loopEnded", onLoopEndedCallbackId, {});
  }
}
```

I still feel it might be unsafe. If we call `sendOnLoopEndedEvent` and `clearOnLoopEndedCallback` we might pass the if but call unregistered callback. I think the best way to make it secure is to move all responsibility of callback lifecycle to eventRegistry so it should eventually return info if we try to invoke unexisting callback etc

maciejmakowski2003: it works just like that. it wont break in that case, cause `EventHandlerRegistry` checks it by itself

michalsek (approved):

PR #712, opened 2025-10-02
Feat/thread pool

poneciak57:
Closes RNAA-263
Closes RNAA-259

Implemented simple thread pool which allows for blazingly fast task dispatch. It has some limitations and allow dispatch only from one thread. It is not thread safe. To make it safer while mantaining the speed multi-producer multi-consumer queue/channel would be required. But for now it seems sufficient cuz we want to mostly offload audio thread and still we can create multiple PromiseVendors.

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/utils/ThreadPool.hpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/utils/ThreadPool.hpp
@@ -1,0 +1,100 @@
+#pragma once
+#include <thread>
+#include <vector>
+#include <functional>
+#include <variant>
+#include <audioapi/utils/SpscChannel.hpp>
+
+namespace audioapi {
+using namespace audioapi::channels::spsc;
+
+/// @brief A simple thread pool implementation using lock-free SPSC channels for task scheduling and execution.
+/// @note The thread pool consists of a load balancer thread and multiple worker threads.
+/// @note The load balancer receives tasks and distributes them to worker threads in a round-robin fashion.
+/// @note Each worker thread has its own SPSC channel to receive tasks from the load balancer.
+/// @note The thread pool can be shut down gracefully by sending a stop event to the load balancer, which then propagates the stop event to all worker threads.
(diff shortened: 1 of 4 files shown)


maciejmakowski2003 (approved):

maciejmakowski2003:
LGTM

packages/react-native-audio-api/common/cpp/audioapi/utils/ThreadPool.hpp
```diff
@@ -0,0 +1,98 @@
...
+/// @note Each worker thread has its own SPSC channel to receive tasks from the load balancer.
+/// @note The thread pool can be shut down gracefully by sending a stop event to the load balancer, which then propagates the stop event to all worker threads.
+/// @note IMPORTANT: ThreadPool is not thread-safe and events should be scheduled from a single thread only.
+class ThreadPool {
+  struct StopEvent {};
+  struct TaskEvent { std::function<void()> task; };
+  using Event = std::variant<TaskEvent, StopEvent>;
+
+  using Sender = audioapi::channels::spsc::Sender<Event, audioapi::channels::spsc::OverflowStrategy::WAIT_ON_FULL, audioapi::channels::spsc::WaitStrategy::ATOMIC_WAIT>;
+  using Receiver = audioapi::channels::spsc::Receiver<Event, audioapi::channels::spsc::OverflowStrategy::WAIT_ON_FULL, audioapi::channels::spsc::WaitStrategy::ATOMIC_WAIT>;
```
michalsek: the `audioapi::` shouldn't be required here and removing it might sligtly declutter the code :)

michalsek (commented):
one nit, good job! :)

packages/react-native-audio-api/common/cpp/audioapi/utils/ThreadPool.hpp
poneciak57: good catch

PR #725, opened 2025-10-08
fix: tryin to fix audio array buffer

maciejmakowski2003:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/core/effects/WorkletNode.cpp
```diff
@@ -63,16 +48,16 @@ std::shared_ptr<AudioBus> WorkletNode::processNode(
...
+              audioArray->copy(bus_->getChannel(ch));
               auto sharedAudioArray =
-                  std::make_shared<AudioArrayBuffer>(buffPtr, buffRealLength_);
+                  std::make_shared<AudioArrayBuffer>(audioArray);
               auto arrayBuffer =
                   jsi::ArrayBuffer(uiRuntimeRaw, std::move(sharedAudioArray));
               jsArray.setValueAtIndex(uiRuntimeRaw, ch, std::move(arrayBuffer));
             }
-            jsArray.setExternalMemoryPressure(
-                uiRuntimeRaw, channelCount_ * buffRealLength_);
```
michalsek, line 75: I don't see memory pressure set anywhere else (host objects or something), won't it cause JS GC to hoard our buffers?

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBuffer.h
```diff
@@ -10,6 +10,7 @@
 namespace audioapi {
 
 class AudioBus;
+class AudioBufferHostObject;
```
mdydek: since there is no usage of `AudioBufferHostObject` anywhere in header I don't see a reason for forward declaration here of this class

packages/react-native-audio-api/common/cpp/audioapi/core/effects/WorkletNode.cpp
maciejmakowski2003: I revert it just in case

PR #732, opened 2025-10-10
Fix/streamer fails on stream end

### LINKED ISSUE #731: [iOS] Expo App Crash when using StreamerNode when reaching end of stream
> Hi @mdydek! I just started using the library today, and I was looking for this exact thing. I'm currently seeing on `0.9.0` my Expo app crash (Simulator - iPhone 16 - iOS 18.6) when I allow a song to play all the way through, and the `onEnded` callback fires specifically in the `StreamerNode::cleanup()` function when the audio finishes playing. The crash occurs in `std::__1::thread::join()`, which imo suggests the library is trying to join a thread that's already been joined or destroyed. Not an expert but from the Apple crash report, I gather:
> ```
> audioapi::StreamerNode::cleanup() (StreamerNode.cpp:276)
>     ↓
>   std::__1::thread::join()
>     ↓
>   SIGABRT
> ```
> I hope #718 addresses this for iOS as well, since it's fixing callback clearing for `AudioScheduledSourceNode`, which StreamerNode derives from.
> 
> Will look out for `0.10.0` and then test, and if not, open a new bug.
> 
> Thanks! 

 _Originally posted by @jlm0 in [#717](https://github.com/software-mansion/react-native-audio-api/issues/717#issuecomment-3383779080)_

mdydek:
Closes #731 

## Introduced changes

- removed possibility of invoking cleanup from streaming thread
- recompiled ffmpeg with http support

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
@@ -125,25 +125,20 @@ void StreamerNode::streamAudio() {
   while (streamFlag.load()) {
     if (pendingFrame_ != nullptr) {
       if (!processFrameWithResampler(pendingFrame_)) {
-        cleanup();
         return;
       }
     } else {
       if (av_read_frame(fmtCtx_, pkt_) < 0) {
-        cleanup();
         return;
       }
       if (pkt_->stream_index == audio_stream_index_) {
         if (avcodec_send_packet(codecCtx_, pkt_) != 0) {
-          cleanup();
           return;
         }
         if (avcodec_receive_frame(codecCtx_, frame_) != 0) {
-          cleanup();
           return;
         }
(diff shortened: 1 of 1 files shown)


packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
```diff
@@ -44,10 +44,12 @@ bool StreamerNode::initialize(const std::string &input_url) {
   }
 
   if (!openInput(input_url)) {
+    printf("Failed to open input\n");
```
michalsek: I would either cleanup those printfs or move popup them to JS layer

michalsek: Or hide behind some dev flag

PR #733, opened 2025-10-10
Fix/async promise dispatch

poneciak57:
## Introduced changes
- decoding is performed on separate thread in background

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioDecoderHostObject.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioDecoderHostObject.cpp
@@ -32,29 +32,30 @@ JSI_HOST_FUNCTION_IMPL(AudioDecoderHostObject, decodeWithMemoryBlock) {
 
   auto sampleRate = args[1].getNumber();
 
-  auto promise = promiseVendor_->createPromise(
-      [data, size, sampleRate](std::shared_ptr<Promise> promise) {
-        std::thread([data, size, sampleRate, promise = std::move(promise)]() {
-          auto result =
-              AudioDecoder::decodeWithMemoryBlock(data, size, sampleRate);
+  auto promise = promiseVendor_->createAsyncPromise(
+      [data, size, sampleRate]() -> PromiseResolver {
+        auto result =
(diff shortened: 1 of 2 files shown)


packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioDecoderHostObject.cpp
```diff
@@ -32,29 +32,30 @@ JSI_HOST_FUNCTION_IMPL(AudioDecoderHostObject, decodeWithMemoryBlock) {
...
-        }).detach();
+  auto promise = promiseVendor_->createAsyncPromise(
+      [data, size, sampleRate]() -> PromiseResolver {
+        auto result =
+            AudioDecoder::decodeWithMemoryBlock(data, size, sampleRate);
+
+        if (!result) {
+          return [](jsi::Runtime &runtime)
+                     -> std::variant<jsi::Value, std::string> {
+            return std::string("Failed to decode audio data.");
```
msluszniak, line 43: This is a matter of style, but I like having `using std::literals::string_literals::operator""s` at the top, and then use `return "Failed to decode audio data."s;` instead. It's more elegant and shorter, but less explicit, so decide which one you prefer and feel free to ignore ;)

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioDecoderHostObject.cpp
```diff
@@ -63,29 +64,29 @@ JSI_HOST_FUNCTION_IMPL(AudioDecoderHostObject, decodeWithFilePath) {
...
-                runtime, audioBufferHostObject->getSizeInBytes());
-            return jsiObject;
-          });
-        }).detach();
+  auto promise = promiseVendor_->createAsyncPromise(
+      [sourcePath, sampleRate]() -> PromiseResolver {
+        auto result = AudioDecoder::decodeWithFilePath(sourcePath, sampleRate);
+
+        if (!result) {
+          return [](jsi::Runtime &runtime)
```
msluszniak, line 72: `runtime` is not used here, comment out the name to prevent some warnings about unused param

PR #741, opened 2025-10-15
Fix/expo54 worklets build error

poneciak57:
Closes RNAA-294

## Introduced changes
- fixed an issue with `react-native-worklets` library compatibility

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/src/core/AudioContext.ts
+++ b/packages/react-native-audio-api/src/core/AudioContext.ts
@@ -3,10 +3,7 @@ import BaseAudioContext from './BaseAudioContext';
 import AudioManager from '../system';
 import { AudioContextOptions } from '../types';
 import { NotSupportedError } from '../errors';
-import {
-  isWorkletsVersionSupported,
-  workletsModule,
-} from '../utils';
+import { isWorkletsVersionSupported, workletsModule } from '../utils';
 
 export default class AudioContext extends BaseAudioContext {
   constructor(options?: AudioContextOptions) {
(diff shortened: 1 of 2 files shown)


maciejmakowski2003 (approved):

packages/react-native-audio-api/src/core/AudioContext.ts
```diff
@@ -17,7 +21,8 @@ export default class AudioContext extends BaseAudioContext {
       );
     }
     let audioRuntime = null;
-    if (isWorkletsAvailable) {
+
+    if (isWorkletsAvailable && isWorkletsVersionSupported) {
```
mdydek: imo there is no reason to export both `isWorkletsAvailable` and `isWorkletsVersionSupported`, because in this setup only latter is needed. They should be left internally for more informative assert, but there is no need to check them both, because they are tighly coupled

packages/react-native-audio-api/src/core/OfflineAudioContext.ts
```diff
@@ -18,7 +22,7 @@ export default class OfflineAudioContext extends BaseAudioContext {
     arg2?: number
   ) {
     let audioRuntime = null;
-    if (isWorkletsAvailable) {
+    if (isWorkletsAvailable && isWorkletsVersionSupported) {
```
mdydek: same here

packages/react-native-audio-api/src/core/AudioContext.ts
poneciak57: But there is an overhead of creating a runtime so if we know that there is no feature depending on it available we can skip it

PR #749, opened 2025-10-20
Refactor/remove binaries from repo

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- moved binaries outside of our main repository

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/scripts/download-prebuilt-binaries.sh
+++ b/packages/react-native-audio-api/scripts/download-prebuilt-binaries.sh
@@ -4,15 +4,15 @@
 MAIN_DOWNLOAD_URL="https://github.com/software-mansion-labs/rn-audio-libs/releases/download"
 TAG="v1.0.0"
 DOWNLOAD_NAMES=(
-    "armeabi-v7a.zip" "arm64-v8a.zip" "x86.zip" "x86_64.zip" 
+    "armeabi-v7a.zip" "arm64-v8a.zip" "x86.zip" "x86_64.zip"
     "ffmpeg_ios.zip" "iphoneos.zip" "iphonesimulator.zip" "jniLibs.zip"
 )
 
 # Use a temporary directory for downloads, ensuring it exists
-TEMP_DOWNLOAD_DIR="/tmp/audioapi-binaries-temp"
+TEMP_DOWNLOAD_DIR="$(pwd)/audioapi-binaries-temp"
 mkdir -p "$TEMP_DOWNLOAD_DIR"
 if [ $1 == "android" ]
-then 
+then
     PROJECT_ROOT="$(pwd)/.."
 else
     PROJECT_ROOT="$(pwd)"
@@ -21,7 +21,7 @@ fi
 for name in "${DOWNLOAD_NAMES[@]}"; do
     ARCH_URL="${MAIN_DOWNLOAD_URL}/${TAG}/${name}"
     ZIP_FILE_PATH="${TEMP_DOWNLOAD_DIR}/${name}"
-    
+
     if [[ "$name" == "jniLibs.zip" ]]; then
         OUTPUT_DIR="${PROJECT_ROOT}/android/src/main"
     else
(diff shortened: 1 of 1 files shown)


packages/react-native-audio-api/android/build.gradle
```diff
@@ -277,10 +276,16 @@ def assertMinimalReactNativeVersion = task assertMinimalReactNativeVersionTask {
     }
 }
 
+task downloadPrebuiltBinaries(type: Exec) {
+  commandLine 'chmod', '+x', '../scripts/download-prebuilt-binaries.sh'
+  commandLine 'bash', '../scripts/download-prebuilt-binaries.sh'
+  args 'android'
+}
+
```
poneciak57, line 285: i guess it will fail on windows 💀

poneciak57 (changes requested):
In my view the best way to handle that would be to make it optional similarly to worklets when user can for example download zip with prebuild binaries or use script i dont think that it will work on windows.

packages/react-native-audio-api/android/build.gradle
mdydek: so your suggestion is that invoking this script should be optional? what will be responsible for deciding?

poneciak57: No my main statement is that there is no bash on windows and it will always fail when trying to build on windows. The thing with decission is just suggestion. Best way would be to create node script tbh as many libraries use them such as reanimated or our lib.

mdydek: after discussion with a team, we agreed to not support non-unix os :) @poneciak57

packages/react-native-audio-api/scripts/download-prebuilt-binaries.sh
```diff
@@ -0,0 +1,54 @@
...
+
+MAIN_DOWNLOAD_URL="https://github.com/software-mansion-labs/rn-audio-libs/releases/download"
+TAG="v1.0.0"
+DOWNLOAD_NAMES=(
+    "armeabi-v7a.zip" "arm64-v8a.zip" "x86.zip" "x86_64.zip" 
+    "ffmpeg_ios.zip" "iphoneos.zip" "iphonesimulator.zip" "jniLibs.zip"
+)
+
+# Use a temporary directory for downloads, ensuring it exists
+TEMP_DOWNLOAD_DIR="/tmp/audioapi-binaries-temp"
```
poneciak57: as we talked you should not use `/tmp` as it might be inacessible in sanboxes like EAS builds

packages/react-native-audio-api/scripts/download-prebuilt-binaries.sh
```diff
@@ -0,0 +1,54 @@
...
+            continue 
+        fi
+    fi
+
+    unzip -o "$ZIP_FILE_PATH" -d "$OUTPUT_DIR"
+    
+    MACOSX_DIR="${OUTPUT_DIR}/__MACOSX"
+    if [ -d "$MACOSX_DIR" ]; then
+        rm -rf "$MACOSX_DIR"
+    fi
```
poneciak57:
```suggestion
    rm -rf "$MACOSX_DIR"
```

packages/react-native-audio-api/scripts/download-prebuilt-binaries.sh
```diff
@@ -0,0 +1,54 @@
...
+    
+    if [[ "$name" == "jniLibs.zip" ]]; then
+        OUTPUT_DIR="${PROJECT_ROOT}/android/src/main"
+    else
+        OUTPUT_DIR="${PROJECT_ROOT}/common/cpp/audioapi/external"
+    fi
+
+    mkdir -p "$OUTPUT_DIR"
+
+    if [ -f "$ZIP_FILE_PATH" ]; then
```
poneciak57, line 33: won't this be a problem when in the future those zips get updated? Assume that user has zips version A and rnaa version A and we update api to version B that requires newer zip files, in this scenario older projects might face an issue while bumping api version

packages/react-native-audio-api/scripts/download-prebuilt-binaries.sh
```diff
@@ -0,0 +1,54 @@
...
+        echo "Zip file already exists locally. Skipping download."
+    else
+        echo "Downloading from: $ARCH_URL"
+        curl -fsSL "$ARCH_URL" -o "$ZIP_FILE_PATH"
+        
+        if [ $? -ne 0 ]; then
+            rm -f "$ZIP_FILE_PATH"
+            continue 
+        fi
+    fi
```
poneciak57, line 36: here you can continue in if statement
```suggestion
        echo "Zip file already exists locally. Skipping download."
        continue
    fi
    echo "Downloading from: $ARCH_URL"
    curl -fsSL "$ARCH_URL" -o "$ZIP_FILE_PATH"
        
    if [ $? -ne 0 ]; then
         rm -f "$ZIP_FILE_PATH"
         continue 
    fi
```

packages/react-native-audio-api/scripts/download-prebuilt-binaries.sh
mdydek: `rm -rf "$TEMP_DOWNLOAD_DIR"` it won't be a problem, because in the end this directory is deleted, it is a double check if someone or something would interrupt in between, so f.e some zips were downloaded but script did not finished

poneciak57: ah alright

poneciak57 (approved):

packages/react-native-audio-api/android/build.gradle
fax1ty: @mdydek @poneciak57 
Please, add some information to the documentation. Currently, it is impossible to predict whether a build will fail on Windows or not, and why. Also, Bash is available through Git Bash.

image

PR #759, opened 2025-10-23
refactor: added condition variable to streaming thread

mdydek:
Closes RNAA-222

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- logic of buffering had some issues, when streaming thread was reaching upper bound of buffer, some artefacts were incorporated in signal, this approach refactors this case by introducing conditional variable for synchronization between threads

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
@@ -235,7 +235,7 @@ bool StreamerNode::processFrameWithResampler(AVFrame *frame) {
     float *dst = bus.getChannel(ch)->getData();
     memcpy(dst, src, converted_samples * sizeof(float));
   }
-  StreamingData data{bus, static_cast<size_t>(converted_samples)};
+  StreamingData data{std::move(bus), static_cast<size_t>(converted_samples)};
   sender_.send(std::move(data));
   return true;
 }
--- a/packages/react-native-audio-api/common/cpp/audioapi/utils/AudioBus.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/utils/AudioBus.cpp
@@ -39,7 +39,6 @@ AudioBus::AudioBus(AudioBus &&other) noexcept
       numberOfChannels_(other.numberOfChannels_),
       sampleRate_(other.sampleRate_),
       size_(other.size_) {
-  other.channels_.clear();
   other.numberOfChannels_ = 0;
   other.sampleRate_ = 0.0f;
   other.size_ = 0;


mdydek:
you can hear a difference by using link provided in example, compare main version to this branch

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
```diff
@@ -160,30 +153,33 @@ std::shared_ptr<AudioBus> StreamerNode::processNode(
...
-          processingBus->getChannel(ch)->getData(),
-          bufferedBus_->getChannel(ch)->getData(),
-          offsetLength * sizeof(float));
-
-      memmove(
-          bufferedBus_->getChannel(ch)->getData(),
-          bufferedBus_->getChannel(ch)->getData() + offsetLength,
-          (maxBufferSize_ - offsetLength) * sizeof(float));
+  {
+    Locker locker(bufferMutex_);
```
poneciak57: use SPSC try_receive here with local buffering

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
```diff
@@ -220,16 +216,17 @@ bool StreamerNode::processFrameWithResampler(AVFrame *frame) {
     return false;
   }
 
-  // Check if converted data fits in buffer
-  if (bufferedBusIndex_ + converted_samples > maxBufferSize_) {
-    pendingFrame_ = frame;
+  std::unique_lock<std::mutex> lock(bufferMutex_);
```
poneciak57: use SPSC send here with ATOMIC_WAIT

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
```diff
@@ -220,16 +216,17 @@ bool StreamerNode::processFrameWithResampler(AVFrame *frame) {
...
+  busCv_.wait(lock, [&] {
+    // Check if converted data fits in buffer or if we would like to finish
+    return (bufferedBusIndex_ + converted_samples) <= maxBufferSize_ ||
+        !streamFlag.load();
+  });
+  // if we would like to finish dont copy anything
+  if (!streamFlag.load()) {
```
poneciak57: use explicit memory ordering
https://en.cppreference.com/w/cpp/atomic/memory_order.html

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
```diff
@@ -281,6 +278,7 @@ bool StreamerNode::setupDecoder() {
 
 void StreamerNode::cleanup() {
   streamFlag.store(false);
```
poneciak57: use explicit memory ordering
https://en.cppreference.com/w/cpp/atomic/memory_order.html

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
```diff
@@ -84,6 +83,7 @@ bool StreamerNode::initialize(const std::string &input_url) {
 void StreamerNode::stop(double when) {
   AudioScheduledSourceNode::stop(when);
   streamFlag.store(false);
+  busCv_.notify_one();
 }
```
poneciak57, line 91: after first call to stop streaming thread will finish and be lost is it intended behaviour ?

mdydek: what do you mean, by "be lost"?

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
```diff
@@ -220,22 +223,20 @@ bool StreamerNode::processFrameWithResampler(AVFrame *frame) {
...
+      codecCtx_->ch_layout.nb_channels,
+      context_->getSampleRate());
   for (int ch = 0; ch < codecCtx_->ch_layout.nb_channels; ch++) {
     auto *src = reinterpret_cast<float *>(resampledData_[ch]);
-    float *dst = bufferedBus_->getChannel(ch)->getData() + bufferedBusIndex_;
+    float *dst = bus->getChannel(ch)->getData();
     memcpy(dst, src, converted_samples * sizeof(float));
   }
-  bufferedBusIndex_ += converted_samples;
+  sender_.send({bus, static_cast<size_t>(converted_samples)});
```
poneciak57: you don't need shared_ptr here create an AudioBus and move it in send

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
poneciak57: In `initialize` you create a thread
```c++
streamingThread_ = std::thread(&StreamerNode::streamAudio, this);
```
```c++
void StreamerNode::streamAudio() {
  while (streamFlag.load(std::memory_order_acquire)) {
     ...
```
If you call stop it will stop streaming forever as the while loop will quit and thread will automatically join.
So like it will "be lost" in some sense like you will not be able to "reviwe" it. If thats intended and i am lacking context thats alright just lmk.

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
poneciak57: make sure it has move constructor if not you can implement it

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
```diff
@@ -220,22 +223,20 @@ bool StreamerNode::processFrameWithResampler(AVFrame *frame) {
     return false;
   }
 
-  // Check if converted data fits in buffer
-  if (bufferedBusIndex_ + converted_samples > maxBufferSize_) {
-    pendingFrame_ = frame;
+  // if we would like to finish dont copy anything
+  if (!streamFlag.load(std::memory_order_acquire)) {
```
poneciak57, line 226: this condition is already checked in streamAudio

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
```diff
@@ -160,28 +158,33 @@ std::shared_ptr<AudioBus> StreamerNode::processNode(
...
+        memcpy(
+            processingBus->getChannel(ch)->getData(),
+            bufferedBus_->getChannel(ch)->getData() + processedSamples_,
+            bufferRemaining * sizeof(float));
+      }
+      framesToProcess -= bufferRemaining;
+      alreadyProcessed += bufferRemaining;
+    }
+    StreamingData data;
+    receiver_.try_receive(data);
```
poneciak57, line 174: it may fail if channel is empty

poneciak57 (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
mdydek: according to docs `Similar to all of AudioScheduledSourceNodes, it can be started only once. If you want to play the same sound again you have to create a new one.`

mdydek: so it is intended behavior that calling stop does not allow for "revival"

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
mdydek: it certainly will, but it would mean that streaming thread is not fast enough, same example when you play YouTube with great quality and you don't have fast connection, video will lag, same goes for audio, but there is no way to improve that we have to assume existence

poneciak57: oh i see that you handle this case below when checking for `nullptr`

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.h
```diff
@@ -56,19 +71,21 @@ class StreamerNode : public AudioScheduledSourceNode {
   AVCodecParameters* codecpar_;
   AVPacket* pkt_;
   AVFrame* frame_; // Frame that is currently being processed
-  AVFrame* pendingFrame_; // Frame that is saved if bufferedBus is full
+  SwrContext* swrCtx_;
+  uint8_t** resampledData_; // weird ffmpeg way of using raw byte pointers for resampled data
+
   std::shared_ptr<AudioBus> bufferedBus_; // audio bus for buffering hls frames
-  size_t bufferedBusIndex_; // index in the buffered bus where we write the next frame
+  size_t bufferedBusSize_; // index in the buffered bus where we write the next frame
```
maciejmakowski2003: I pretty sure that naming is wrong - `// index in the buffered bus where we write the next frame`

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
mdydek: this flag can be changed in between those calls, I think it's good to double check here

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.cpp
```diff
@@ -227,16 +226,17 @@ bool StreamerNode::processFrameWithResampler(AVFrame *frame) {
...
       codecCtx_->ch_layout.nb_channels,
       context_->getSampleRate());
   for (int ch = 0; ch < codecCtx_->ch_layout.nb_channels; ch++) {
     auto *src = reinterpret_cast<float *>(resampledData_[ch]);
-    float *dst = bus->getChannel(ch)->getData();
+    float *dst = bus.getChannel(ch)->getData();
     memcpy(dst, src, converted_samples * sizeof(float));
   }
-  sender_.send({bus, static_cast<size_t>(converted_samples)});
+  StreamingData data{bus, static_cast<size_t>(converted_samples)};
```
poneciak57: bus gets copied here you should move

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.h
```diff
@@ -40,11 +42,22 @@ static constexpr bool VERBOSE = false;
...
+  StreamingData(StreamingData&& data) noexcept : bus(std::move(data.bus)), size(data.size) {}
+  StreamingData& operator=(const StreamingData& data) {
+    if (this == &data) {
+      return *this;
+    }
+    bus = data.bus;
+    size = data.size;
+    return *this;
+  }
 };
```
poneciak57, line 59: It is usually better to create non-copyable types if you don't need to copy it. (This is just a comment if copy constructor is somehow required keep it)

packages/react-native-audio-api/common/cpp/audioapi/utils/AudioBus.cpp
```diff
@@ -34,6 +34,35 @@ AudioBus::AudioBus(const AudioBus &other) {
   }
 }
 
+AudioBus::AudioBus(AudioBus &&other) noexcept
+    : channels_(std::move(other.channels_)),
+      numberOfChannels_(other.numberOfChannels_),
+      sampleRate_(other.sampleRate_),
+      size_(other.size_) {
+  other.channels_.clear();
```
poneciak57: no need to call clear as you moved it

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.h
mdydek: I double checked, looks like channel inside requires it, it didn't compile without it

PR #776, opened 2025-10-31
feat: expo example app

poneciak57:
## Introduced changes
- created expo example app to showcase and test how audio api works in expo projects

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


michalsek (approved):

apps/expo-example/app.json
```diff
@@ -0,0 +1,31 @@
...
+    "orientation": "portrait",
+    "icon": "./assets/icon.png",
+    "userInterfaceStyle": "light",
+    "newArchEnabled": true,
+    "splash": {
+      "image": "./assets/splash-icon.png",
+      "resizeMode": "contain",
+      "backgroundColor": "#ffffff"
+    },
+    "ios": {
```
maciejmakowski2003, line 15: lets add audio api expo plugin, as an example usage if anyone clones it

apps/common-app/package.json
```diff
@@ -5,42 +5,5 @@
   "peerDependencies": {
     "react": "*",
     "react-native": "*"
-  },
-  "dependencies": {
```
maciejmakowski2003, line 9: is it necessary? it was done just like that in order to bump up packages only in one place. maybe we can leave all packages that have a common version for both expo and fabric?

maciejmakowski2003 (changes requested):

apps/common-app/package.json
poneciak57: i can try to revert that particular thing but some random packages were problematic in this setup so i removed all of them

apps/expo-example/app.json
poneciak57: sure

PR #783, opened 2025-11-04
Fix/pause on audio buffer queue source node

miloszwielgus:
Closes RNAA-151

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
@@ -14,6 +14,7 @@ AudioScheduledSourceNode::AudioScheduledSourceNode(BaseAudioContext *context)
       stopTime_(-1.0),
       playbackState_(PlaybackState::UNSCHEDULED) {
   numberOfInputs_ = 0;
+  audioEventHandlerRegistry_ = context_->audioEventHandlerRegistry_;
 }
 
 void AudioScheduledSourceNode::start(double when) {
@@ -45,19 +46,13 @@ bool AudioScheduledSourceNode::isStopScheduled() {
   return playbackState_ == PlaybackState::STOP_SCHEDULED;
 }
 
-void AudioScheduledSourceNode::clearOnEndedCallback() {
-  if (onEndedCallbackId_ == 0 || context_ == nullptr ||
-      context_->audioEventHandlerRegistry_ == nullptr) {
-    return;
-  }
-
-  context_->audioEventHandlerRegistry_->unregisterHandler(
-      "ended", onEndedCallbackId_);
-  onEndedCallbackId_ = 0;
-}
-
 void AudioScheduledSourceNode::setOnEndedCallbackId(const uint64_t callbackId) {
-  onEndedCallbackId_ = callbackId;
+  auto oldCallbackId =
+      onEndedCallbackId_.exchange(callbackId, std::memory_order_acq_rel);
+
+  if (oldCallbackId != 0) {
+    audioEventHandlerRegistry_->unregisterHandler("ended", oldCallbackId);
+  }
 }
 
 void AudioScheduledSourceNode::updatePlaybackInfo(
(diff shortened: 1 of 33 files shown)


apps/common-app/src/examples/PauseResumeBug/PauseResumeBug.tsx
```diff
@@ -0,0 +1,193 @@
+import React, { useState, useEffect, FC, useRef } from 'react';
```
maciejmakowski2003: its is not a great idea to add it to example app, we can remove it just before merge

apps/fabric-example/ios/FabricExample.xcodeproj/project.pbxproj
```diff
@@ -260,7 +260,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CURRENT_PROJECT_VERSION = 1;
-				DEVELOPMENT_TEAM = WC59N2X3FV;
+				DEVELOPMENT_TEAM = ARNJX758MC;
```
maciejmakowski2003: dont push development team changes

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferQueueSourceNode.cpp
```diff
@@ -31,6 +31,11 @@ void AudioBufferQueueSourceNode::stop(double when) {
   isPaused_ = false;
 }
 
+void AudioBufferQueueSourceNode::start(double when) {
+  isPaused_ = false;
```
maciejmakowski2003, line 35: it is not necessary, if we add `isPaused_ = false;` to disable it will perfectly fix like below:

```
void AudioBufferQueueSourceNode::disable() {
  if (isPaused_) {
    playbackState_ = PlaybackState::UNSCHEDULED;
    startTime_ = -1.0;
    stopTime_ = -1.0;
    isPaused_ = false;

    return;
  }

  AudioScheduledSourceNode::disable();
  buffers_ = {};
}
```

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
```diff
@@ -19,6 +19,7 @@ AudioScheduledSourceNode::AudioScheduledSourceNode(BaseAudioContext *context)
 void AudioScheduledSourceNode::start(double when) {
   playbackState_ = PlaybackState::SCHEDULED;
   startTime_ = when;
+  stopTime_ = -1.0;
```
maciejmakowski2003: same here, it is unnecessary with above change to disable

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferQueueSourceNode.cpp
miloszwielgus: Hmm, I tried implementing this but it didn't fix the bug for me. If I traced it, correctly the issue stems from a race condition between the main and audio threads: when we quickly go `start -> pause -> start` the second start runs before the `disable()`, so that it still has `stopTime_ = 0.0` and the node does not play. Thats why in my example adding even a 1ms delay after `pause()` allows `disable()` to run before the second `start` and fixes the issue.

maciejmakowski2003: hmm, so lets add this line `isPaused_ = false;` to disable() as well

apps/common-app/src/examples/index.ts
```diff
@@ -103,5 +105,11 @@ export const Examples: Example[] = [
...
     subtitle: 'Stream audio from a URL',
     screen: Streaming,
-  }
+  },
+  {
+    key: 'PauseResumeBug',
+    title: 'Pause/Resume Bug',
+    subtitle: 'Reproduce AudioBufferQueueSourceNode pause/resume bug',
+    screen: PauseResumeBug,
+  },
```
mdydek: 👀 some leftover probably

michalsek (approved):

maciejmakowski2003 (approved):

PR #796, opened 2025-11-09
Refactor/context wise

maciejmakowski2003:
Closes #

## ⚠️ Breaking changes ⚠️

- Removed `initSuspended` option from `AudioContext` ctor
- Changed default state of context to `suspended`, however if user starts some source node, context will be automatically started. it wont work if context was suspended by the user

## Introduced changes

- 

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.cpp
@@ -51,16 +51,12 @@ bool AudioContext::resume() {
     return true;
   }
 
-  if (!playerHasBeenStarted_ && audioPlayer_->start()) {
-    playerHasBeenStarted_ = true;
-    state_ = ContextState::RUNNING;
-    return true;
-  } else if (playerHasBeenStarted_ && audioPlayer_->resume()) {
+  if (playerHasBeenStarted_ && audioPlayer_->resume()) {
     state_ = ContextState::RUNNING;
     return true;
   }
 
-  return false;
+  return start();
 }
 
 bool AudioContext::suspend() {
@@ -78,11 +74,19 @@ bool AudioContext::suspend() {
   return true;
 }
 
-void AudioContext::start() {
+bool AudioContext::start() {
+  if (isClosed()) {
+    return false;
+  }
+
   if (!playerHasBeenStarted_ && audioPlayer_->start()) {
     playerHasBeenStarted_ = true;
     state_ = ContextState::RUNNING;
+
+    return true;
   }
+
+  return false;
 }
 
 std::function<void(std::shared_ptr<AudioBus>, int)>
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.h
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.h
@@ -21,7 +21,7 @@ class AudioContext : public BaseAudioContext {
   void close();
   bool resume();
   bool suspend();
-  void start();
+  bool start();
 
  private:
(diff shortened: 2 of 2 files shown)


packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.cpp
```diff
@@ -67,13 +57,11 @@ bool AudioContext::resume() {
...
-  }
-
-  if (audioPlayer_->resume()) {
-    state_ = ContextState::RUNNING;
-    return true;
+  } else {
+    if (audioPlayer_->resume()) {
+      state_ = ContextState::RUNNING;
+      return true;
+    }
```
vltkv:
```suggestion
  } else if (audioPlayer_->resume()) {
      state_ = ContextState::RUNNING;
      return true;
```

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
```diff
@@ -18,6 +22,12 @@ AudioScheduledSourceNode::AudioScheduledSourceNode(BaseAudioContext *context)
 }
 
 void AudioScheduledSourceNode::start(double when) {
+#if !RN_AUDIO_API_TEST
+  if (auto context = dynamic_cast<AudioContext *>(context_)) {
+    context->start();
+  }
+#endif
+
```
poneciak57, line 30: what if you try to use AudioScheduledSourceNode with OfflineAudioContext? It will be UB for sure

poneciak57 (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.cpp
```diff
@@ -94,6 +78,13 @@ bool AudioContext::suspend() {
   return true;
 }
 
+void AudioContext::start() {
+  if (!playerHasBeenStarted_ && audioPlayer_->start()) {
+    playerHasBeenStarted_ = true;
+    state_ = ContextState::RUNNING;
```
mdydek, line 84: code repetition here

packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.cpp
```diff
@@ -61,17 +51,11 @@ bool AudioContext::resume() {
...
-    }
-
-    return false;
-  }
-
-  if (audioPlayer_->resume()) {
+  if (!playerHasBeenStarted_ && audioPlayer_->start()) {
+    playerHasBeenStarted_ = true;
+    state_ = ContextState::RUNNING;
+    return true;
```
mdydek: and here

packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.cpp
```diff
@@ -61,17 +51,11 @@ bool AudioContext::resume() {
...
+  } else if (playerHasBeenStarted_ && audioPlayer_->resume()) {
     state_ = ContextState::RUNNING;
     return true;
   }
```
mdydek:
```suggestion
if (playerHasBeenStarted_ && audioPlayer_->resume()) {
state_ = ContextState::RUNNING;
return true;
}
start();
```
something like that should be equivalent

maciejmakowski2003: actually it is not. we wont get a boolean indicating if operation is successful

PR #807, opened 2025-11-13
feat: audio node options in ctor

### LINKED ISSUE #574: Inconsistent return type of createBufferSource() between platforms (Promise on Web, AudioBufferSourceNode on iOS)
### Description

While following [this example](https://docs.swmansion.com/react-native-audio-api/guides/lets-make-some-noise/#loading-an-audio-file), I encountered an issue on Expo Web due to CORS protection on the remote audio file. To work around it, I used a local audio file via Expo assets as suggested in the [example for decodeAudioDataSource()](https://docs.swmansion.com/react-native-audio-api/core/base-audio-context/#decodeaudiodatasource).

Then I used the following code:
`const playerNode = audioContext.createBufferSource();`

This worked as expected on iOS, but on Web I received the following error:
`Uncaught (in promise) TypeError: playerNode.connect is not a function`

After some investigation, I found the root cause:
- On **iOS**, `createBufferSource()` returns an `AudioBufferSourceNode`.
- On **Web**, it returns a `Promise<AudioBufferSourceNode>`.

Changing the code to:
`const playerNode = await audioContext.createBufferSource();`
resolved the issue on both platforms.

📌 Suggestion
If this behavior is intended, it would be very helpful to document the platform-specific return type of `createBufferSource()`—especially since it's unexpected and can cause runtime issues.

If this is not intended, I recommend unifying the behavior so that createBufferSource() returns a consistent type across platforms.

Thanks again for the great work on this library—just wanted to flag this inconsistency to help improve the dev experience

### Steps to reproduce

        const audioContext = new AudioContext();

        const buffer = await
            Asset.fromModule(require('@/assets/audio/example-music-01.mp3'))
                .downloadAsync()
                .then((asset) => {
                    if (!asset.localUri) {
                        throw new Error('Failed to load audio asset');
                    }
                    return audioContext.decodeAudioDataSource(asset.localUri);
                })

        const playerNode = audioContext.createBufferSource();
        playerNode.buffer = buffer;
        playerNode.connect(audioContext.destination);
        playerNode.start(audioContext.currentTime);
        playerNode.stop(audioContext.currentTime + 10);

https://snack.expo.dev/@niklasdahlheimer/createbuffersource-example

0.6.5

### React Native version

0.79.5

### Platforms

iOS

### JavaScript runtime

None

### Workflow

Expo Dev Client

### Architecture

Fabric (New Architecture)

### Build type

Debug app & dev bundle

### Device

iOS simulator

### Device model

_No response_

### Acknowledgements

Yes

mdydek:
Closes RNAA-297, RNAA-293
Closes #574 

## ⚠️ Breaking changes ⚠️

- createIIRFilter method of BaseAudioContext now takes two parameters `feedforward: number[], feedback: number[]` instead of one object that encapsulated them
- createConvolver method of BaseAudioContext now takes no parameters
- createBuffer method of BaseAudioContext now has first parameter named `numberOfChannels`

## Introduced changes

- now nodes can be constructed also by `new` keyword with some possible option to manipulate its values

LIST:
- [x] AudioBuffer
- [x] AudioNode
- [x] AnalyserNode
- [x] AudioBufferBaseSourceNode
- [x] AudioBufferQueueSourceNode
- [x] AudioBufferSourceNode
- [x] BiquadFilterNode
- [x] ConstantSourceNode
- [x] ConvolverNode
- [x] DelayNode
- [x] GainNode
- [x] IIRFilterNode
- [x] OscillatorNode
- [x] PeriodicWave
- [x] RecorderAdapterNode
- [x] StereoPannerNode
- [x] StreamerNode 
- [x] WorkletNode
- [x] WorkletSourceNode
- [x] WorkletProcessingNode

## Checklist

- [x] Linked relevant issue
- [x] Added support for web

### DIFF
--- a/apps/common-app/src/demos/Record/Record.tsx
+++ b/apps/common-app/src/demos/Record/Record.tsx
@@ -47,7 +47,7 @@ const Record: FC = () => {
       resumeIconResourceName: 'resume',
       color: 0xff6200,
     });
-  }
+  };
 
   const onStartRecording = useCallback(async () => {
     if (state !== RecordingState.Idle) {
@@ -67,14 +67,27 @@ const Record: FC = () => {
       setHasPermissions(true);
     }
 
-    const success = await AudioManager.setAudioSessionActivity(true);
+    let success = false;
 
-    if (!success) {
+    try {
+      success = await AudioManager.setAudioSessionActivity(true);
+    } catch (error) {
+      console.error(error);
       Alert.alert('Error', 'Failed to activate audio session for recording.');
+      setState(RecordingState.Idle);
       return;
     }
 
-    const result = Recorder.start();
+    if (!success) {
+      Alert.alert('Error', 'Failed to activate audio session for recording.');
+      setState(RecordingState.Idle);
+      return;
+    }
+
+    const result = Recorder.start({
+      fileNameOverride: `overridden_name_${Date.now()}`,
+    });
+
     setupNotification(false);
 
     if (result.status === 'success') {
--- a/apps/fabric-example/ios/Podfile.lock
+++ b/apps/fabric-example/ios/Podfile.lock
@@ -3291,7 +3291,7 @@ SPEC CHECKSUMS:
   FBLazyVector: 309703e71d3f2f1ed7dc7889d58309c9d77a95a4
   fmt: a40bb5bd0294ea969aaaba240a927bd33d878cdd
   glog: 5683914934d5b6e4240e497e0f4a3b42d1854183
-  hermes-engine: f93b5009d8ccd9429fe2a772351980df8a22a413
+  hermes-engine: 42d6f09ee6ede2feb220e2fb772e8bebb42ca403
   RCT-Folly: 846fda9475e61ec7bcbf8a3fe81edfcaeb090669
   RCTDeprecation: a41bbdd9af30bf2e5715796b313e44ec43eefff1
   RCTRequired: 7be34aabb0b77c3cefe644528df0fa0afad4e4d0
--- a/packages/audiodocs/docs/inputs/audio-recorder.mdx
+++ b/packages/audiodocs/docs/inputs/audio-recorder.mdx
@@ -86,7 +86,7 @@ Additionally to be able to record audio while application is in the background, 
       <uses-permission android:name="android.permission.RECORD_AUDIO"/>
 
       <!-- Paste this inside <application> tag -->
-        <service android:stopWithTask="true" android:name="com.swmansion.audioapi.system.MediaNotificationManager$AudioForegroundService" android:foregroundServiceType="microphone" />
+        <service android:stopWithTask="true" android:name="com.swmansion.audioapi.system.CentralizedForegroundService" android:foregroundServiceType="microphone" />
     </manifest>
 
     ```
--- a/packages/audiodocs/docs/other/audio-api-plugin.mdx
+++ b/packages/audiodocs/docs/other/audio-api-plugin.mdx
@@ -112,6 +112,8 @@ Allows to specify certain android app permissions to apply.
 
 ##### Permissions:
 
+- `android.permission.POST_NOTIFICATIONS` - Required by Foreground Services on Android 13+ to post notifications.
+
 - `android.permission.FOREGROUND_SERVICE` - Allows an app to run a Foreground Service
 
 - `android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK` - Allows an app to run a Foreground Service specifically for continues audio or video playback.
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
@@ -102,7 +102,7 @@ Result<NoneType, std::string> AndroidAudioRecorder::openAudioStream() {
 /// RN side requires their "file://" prefix, but sometimes it returned raw path.
 /// Most likely this was due to alpha version mistakes, but in case of problems leaving this here. (ㆆ _ ㆆ)
 /// @returns On success, returns the file URI where the recording is being saved (if file output is enabled).
-Result<std::string, std::string> AndroidAudioRecorder::start() {
+Result<std::string, std::string> AndroidAudioRecorder::start(const std::string &fileNameOverride) {
   std::scoped_lock startLock(callbackMutex_, fileWriterMutex_, adapterNodeMutex_);
 
   if (!isIdle()) {
@@ -120,9 +120,12 @@ Result<std::string, std::string> AndroidAudioRecorder::start() {
   }
 
   if (usesFileOutput()) {
-    auto fileResult =
-        std::static_pointer_cast<AndroidFileWriterBackend>(fileWriter_)
-            ->openFile(streamSampleRate_, streamChannelCount_, streamMaxBufferSizeInFrames_);
+    auto fileResult = std::static_pointer_cast<AndroidFileWriterBackend>(fileWriter_)
+                          ->openFile(
+                              streamSampleRate_,
+                              streamChannelCount_,
+                              streamMaxBufferSizeInFrames_,
+                              fileNameOverride);
 
     if (!fileResult.is_ok()) {
       return Result<std::string, std::string>::Err(
@@ -228,7 +231,7 @@ Result<std::string, std::string> AndroidAudioRecorder::enableFileOutput(
   if (!isIdle()) {
     auto fileResult =
         std::static_pointer_cast<AndroidFileWriterBackend>(fileWriter_)
-            ->openFile(streamSampleRate_, streamChannelCount_, streamMaxBufferSizeInFrames_);
+            ->openFile(streamSampleRate_, streamChannelCount_, streamMaxBufferSizeInFrames_, "");
 
     if (!fileResult.is_ok()) {
       return Result<std::string, std::string>::Err(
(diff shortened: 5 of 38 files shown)


maciejmakowski2003 (commented):
looks promising, @michalsek wdyt?

packages/react-native-audio-api/common/cpp/audioapi/core/effects/IIRFilterNode.cpp
```diff
@@ -34,13 +35,9 @@
...
 namespace audioapi {
 
-IIRFilterNode::IIRFilterNode(
-    BaseAudioContext *context,
-    const std::vector<float> &feedforward,
-    const std::vector<float> &feedback)
-    : AudioNode(context), feedforward_(feedforward), feedback_(feedback) {
+IIRFilterNode::IIRFilterNode(BaseAudioContext *context, IIRFilterOptions options)
+    : AudioNode(context, options), feedforward_(options.feedforward), feedback_(options.feedback) {
   isInitialized_ = true;
```
poneciak57: here you copy possibly large vector

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/NodeOptions.h
```diff
@@ -0,0 +1,104 @@
...
+namespace audioapi {
+struct AudioNodeOptions {
+  int channelCount = 2;
+  ChannelCountMode channelCountMode = ChannelCountMode::MAX;
+  ChannelInterpretation channelInterpretation = ChannelInterpretation::SPEAKERS;
+};
+
+struct GainOptions : AudioNodeOptions {
+  float gain = 1.0f;
+};
```
poneciak57, line 26: That is not a change per se but just a quick reminder.
Remember when you have method
```c++
void method(AudioNodeOptions opts) {
    //
}

// and you call it with type extending the node options like so
GainOptions opts;
method(opts);
```
This type is sliced so it is copied here and it copies only part of `AudioNodeOptions`. Everything is fine with it but it is worth noting as it has some limitations and in future someone might not know this. So like leave some explanation here on how it works or in `AudioNode`.

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/NodeOptions.h
```diff
@@ -0,0 +1,104 @@
...
+
+#include <audioapi/core/effects/PeriodicWave.h>
+#include <audioapi/core/sources/AudioBuffer.h>
+#include <audioapi/core/types/BiquadFilterType.h>
+#include <audioapi/core/types/ChannelCountMode.h>
+#include <audioapi/core/types/ChannelInterpretation.h>
+#include <audioapi/core/types/OscillatorType.h>
+
+namespace audioapi {
+struct AudioNodeOptions {
```
poneciak57, line 18: you can make all types extend some empty base like
```c++
struct BaseOption {}
```
it does not cost anything compilers optimize this and might be helpfull in the future if all options types can be virtualized

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferBaseSourceNode.h
```diff
@@ -11,10 +11,13 @@ namespace audioapi {
...
 class AudioBus;
 class AudioParam;
+class BaseAudioBufferSourceOptions;
 
 class AudioBufferBaseSourceNode : public AudioScheduledSourceNode {
  public:
-  explicit AudioBufferBaseSourceNode(BaseAudioContext *context, bool pitchCorrection);
+  explicit AudioBufferBaseSourceNode(
+      BaseAudioContext *context,
+      BaseAudioBufferSourceOptions options);
```
poneciak57: here type gets sliced as explained before maybe const & ?

packages/react-native-audio-api/src/core/AnalyserNode.ts
```diff
@@ -1,13 +1,33 @@
...
+
+    if (!AnalyserNode.allowedFFTSize.includes(finalOptions.fftSize!)) {
+      throw new IndexSizeError(
+        `fftSize must be one of the following values: ${AnalyserNode.allowedFFTSize.join(
+          ', '
+        )}`
+      );
+    }
+    const analyserNode: IAnalyserNode =
+      context.context.createAnalyser(finalOptions);
```
poneciak57: when you have base option interface might be a good idea to add some virtual method validate so option type would be responsible for its validation that might be helpfull if you need to validate like a base type or something and is a good principle

packages/react-native-audio-api/src/core/ConvolverNode.ts
```diff
@@ -1,12 +1,33 @@
...
+        numberOfChannels !== 2 &&
+        numberOfChannels !== 4
+      ) {
+        throw new NotSupportedError(
+          `The number of channels provided (${numberOfChannels}) in impulse response for ConvolverNode buffer must be 1 or 2 or 4.`
+        );
+      }
+    }
+    const convolverNode: IConvolverNode =
+      context.context.createConvolver(finalOptions);
```
poneciak57: same here validation should be implemented for option type

packages/react-native-audio-api/src/core/OscillatorNode.ts
```diff
@@ -1,17 +1,34 @@
...
+      throw new InvalidStateError(
+        "'type' cannot be set to 'custom' without providing a 'periodicWave'."
+      );
+    }
+
+    if (finalOptions.periodicWave) {
+      finalOptions.type = 'custom';
+    }
+
+    const node = context.context.createOscillator(finalOptions);
```
poneciak57, line 25: same here validation should be implemented for option type

packages/react-native-audio-api/src/core/PeriodicWave.ts
```diff
@@ -1,10 +1,61 @@
...
+      imag = new Float32Array(2048);
+    } else if (sampleRate < 88200) {
+      real = new Float32Array(4096);
+      imag = new Float32Array(4096);
+    } else {
+      real = new Float32Array(16384);
+      imag = new Float32Array(16384);
+    }
+    imag[1] = 1;
+  } else {
```
poneciak57, line 16: same here validation should be implemented for option type

poneciak57 (changes requested):
Great job overall but i see some possible improvements

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/NodeOptions.h
mdydek: I don't like writing features "for the future". Adding some complexity because "maybe someday something will change and it will be helpful" for me should not be done.

poneciak57: in my view it depends but in this particular case it is not a big of a deal so ill resolve

packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.h
```diff
@@ -16,10 +16,13 @@ namespace audioapi {
 class AudioBus;
 class BaseAudioContext;
 class AudioParam;
+class AudioNodeOptions;
 
 class AudioNode : public std::enable_shared_from_this<AudioNode> {
  public:
   explicit AudioNode(BaseAudioContext *context);
+  // usually options are passed as derived class, keep in mind that object passed as options will be sliced
```
poneciak57: that is not entirely true, `const AudioNodeOptions&` is a reference type and it will not be sliced or copied or anything. But if someones defines 
```cpp
struct Base { int a; };
struct Derived : Base { int b; };

void fun1(Base b); // bad for bigger types
void fun2(const Base &b);

Derived d;
fun1(d); // it gets sliced and copied
fun2(d); // it does not get sliced or copied
```

maciejmakowski2003: @mdydek why do not we pass all options by reference?

packages/react-native-audio-api/src/defaults.ts
```diff
@@ -0,0 +1,94 @@
...
+  TAnalyserOptions,
+  TBiquadFilterOptions,
+  TOscillatorOptions,
+  TBaseAudioBufferSourceOptions,
+  TAudioBufferSourceOptions,
+  TAudioBufferOptions,
+  TDelayOptions,
+} from './types';
+
+export const AudioNodeOptions: TAudioNodeOptions = {
```
maciejmakowski2003, line 18: this file stores the same data as `NodeOptions.h`, is it necessary? I would like to avoid maintaining defaults on both sides.

maciejmakowski2003 (commented):
Overall it looks great. I have one remarks apart from comments that I left. It would be great to extract type conversions to HO layer. e.g. `BiquadFilterNode` does not have to convert string to filter type and vice versa. It has to be the responsibility of parser used by HO. Nodes should only process primitives, audio objects or types.

apps/common-app/src/examples/AudioFile/AudioPlayer.ts
```diff
@@ -43,9 +43,7 @@ class AudioPlayer {
       await this.audioContext.resume();
     }
 
-    this.sourceNode = this.audioContext.createBufferSource({
-      pitchCorrection: true,
-    });
```
maciejmakowski2003, line 48: why? I prefer:
```ts
this.audioContext.createBufferSource({
    pitchCorrection: true,
  });
```

imo there is no need top create that kind of breaking change. could be pretty annoying for users

PR #826, opened 2025-12-01
Verify and access react-native-worklets before module installation

michalsek:
Closes #RNAA-366

## Introduced changes

- adds `react-native-worklets` verification as a api installation step
- adds missing `semver` dependency (used by build-time script)
- adds `AudioAPIModule` class as a singleton layer to interact with the native module

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/src/AudioAPIModule/AudioAPIModule.ts
+++ b/packages/react-native-audio-api/src/AudioAPIModule/AudioAPIModule.ts
@@ -28,7 +28,7 @@ class AudioAPIModule {
   constructor() {
     // Important! Verify and import worklets first
     // otherwise the native module installation might crash
-    // if reanimated is not imported before audio-api
+    // if react-native-worklets is not imported before audio-api
     this.#verifyWorklets();
 
     if (this.#verifyInstallation()) {


apps/common-app/src/App.tsx
```diff
@@ -1,28 +1,23 @@
...
 
 const Stack = createStackNavigator();
 
-// Our slider component uses the text prop to display shared value
-// We need it whitelisted in order to have it "animated".
-Animated.addWhitelistedNativeProps({ text: true });
-
 const ItemSeparatorComponent = () => <Spacer.Vertical size={16} />;
 
 const HomeScreen: FC = () => {
```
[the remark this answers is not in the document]
michalsek: k

packages/react-native-audio-api/common/cpp/audioapi/core/analysis/AnalyserNode.h
```diff
@@ -62,7 +62,8 @@ class AnalyserNode : public AudioNode {
   std::unique_ptr<AudioArray> magnitudeBuffer_;
   bool shouldDoFFTAnalysis_{true};
 
-  static WindowType fromString(const std::string &type) {
+  static WindowType fromString(const std::string &type)
+  {
```
maciejmakowski2003: I do not like this formatting

packages/react-native-audio-api/src/AudioAPIModule/AudioAPIModule.ts
```diff
@@ -0,0 +1,131 @@
...
+  #workletsVersion = 'unknown';
+  #workletsAvailable_ = false;
+  #audioRuntime: WorkletRuntime | null = null;
+
+  public supportedWorkletsVersion = ['0.6.0', '0.6.1'];
+
+  constructor() {
+    // Important! Verify and import worklets first
+    // otherwise the native module installation might crash
+    // if reanimated is not imported before audio-api
```
poneciak57:
```suggestion
    // if react-native-worklets is not imported before audio-api
```

packages/react-native-audio-api/src/core/AudioContext.ts
```diff
@@ -16,11 +16,8 @@ export default class AudioContext extends BaseAudioContext {
         `The provided sampleRate is not supported: ${options.sampleRate}`
       );
     }
-    let audioRuntime = null;
 
-    if (isWorkletsVersionSupported) {
-      audioRuntime = workletsModule.createWorkletRuntime('AudioWorkletRuntime');
-    }
+    const audioRuntime = AudioAPIModule.getOrCreateAudioRuntime();
```
poneciak57, line 20: i am not sure if it is correct. C++ part assumes that runtime during node processing is accessed synchronously. If its possible that two context play at the same time asynchronously for example on two different threads it is not correct.

michalsek: won't using same name for every runtime cause similar problems? By briefly checking event loop might be shared

poneciak57: hmm i do not think so the name is some kind of key but ok if event loop is shared that should be alright

PR #847, opened 2025-12-16
Fix/audio engine fixes

michalsek:
## Introduced changes

- return same error message on iOS and android when stopping inactive recorder
- recorder stop: detach the node before calling stopIfPossible (otherwise it will skip)
- Run all engine modifications triggered by notification manager on the same thread (might be possible that in some scenarios apple with trigger multiple different ones at the same time)

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/ios/audioapi/ios/system/NotificationManager.mm
+++ b/packages/react-native-audio-api/ios/audioapi/ios/system/NotificationManager.mm
@@ -136,9 +136,7 @@ static NSString *NotificationManagerContext = @"NotificationManagerContext";
     NSDictionary *body = @{@"type" : @"ended", @"shouldResume" : @(shouldResume)};
     [self.audioAPIModule invokeHandlerWithEventName:@"interruption" eventBody:body];
   } else {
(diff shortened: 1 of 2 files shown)


packages/react-native-audio-api/ios/audioapi/ios/system/NotificationManager.mm
```diff
@@ -146,8 +148,10 @@ - (void)handleSecondaryAudio:(NSNotification *)notification
       [notification.userInfo[AVAudioSessionSilenceSecondaryAudioHintTypeKey] integerValue];
 
   if (secondaryAudioType == AVAudioSessionSilenceSecondaryAudioHintTypeBegin) {
-    [sessionManager markInactive];
-    [audioEngine onInterruptionBegin];
+    dispatch_async(dispatch_get_main_queue(), ^{
+      [sessionManager markInactive];
+      [audioEngine onInterruptionBegin];
```
mdydek, line 153: i think we could move sessionManager call outside of this queue, to not use it more than it is needed

PR #850, opened 2025-12-16
docs: added section about file loading using react image component

mdydek:
Closes RNAA-63

## Introduced changes

- refactored decoding slightly, there is more hidden from user and his api becomes more friendly

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/audiodocs/docs/utils/decoding.mdx
+++ b/packages/audiodocs/docs/utils/decoding.mdx
@@ -9,19 +9,25 @@ import { Optional, MobileOnly } from '@site/src/components/Badges';
 You can decode audio data independently, without creating an AudioContext, using the exported functions [`decodeAudioData`](/docs/utils/decoding#decodeaudiodata) and
 [`decodePCMInBase64`](/docs/utils/decoding#decodepcminbase64).
 
+:::warning
+Decoding on the web has to be done via `AudioContext` only.
+:::
+
 If you already have an audio context, you can decode audio data directly using its [`decodeAudioData`](/docs/core/base-audio-context#decodeaudiodata) function;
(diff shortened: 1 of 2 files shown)


packages/audiodocs/docs/utils/decoding.mdx
```diff
@@ -57,36 +57,76 @@ if not provided, the original sample rate from the file is used.
...
+import { decodeAudioData } from 'react-native-audio-api';
+
+const uri = Image.resolveAssetSource(require('path/to/file')).uri;
+
+const buffer = await decodeAudioData(uri);
+```
+</details>
+
 <details>
 <summary>Example decoding with memory block</summary>
```
maciejmakowski2003: do we need that section? isn't covered by first section?

mdydek: first one refers to static asset in your app, second one when you have remote url, for me, these are separate things

packages/audiodocs/docs/utils/decoding.mdx
```diff
@@ -39,54 +39,76 @@ if not provided, the original sample rate from the file is used.
...
-      console.error('Error decoding audio data source:', error);
-      return null;
-    });
+const buffer = await decodeAudioData(url);
 ```
 </details>
 
+:::warning
+Decoding local assets on web will only work via expo-asset library.
+:::
```
michalsek: hint: not true - `require/import` resolves to domain relative path which will work in most cases. The only problem there might be that if website root is different than domain/subdomain root it might require some additional work.

But this is just a suspicion - havent tested that case

maciejmakowski2003 (approved):

PR #864, opened 2025-12-29
Refactor/communication between audio and js threads

maciejmakowski2003:
Closes RNAA-250

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- Enhanced communication between JS thread and Audio thread. All node's modifications from JS thread are scheduled on `CrossThreadEventScheduler` wrapper to be executed on Audio thread. `HostObject` plays a `ShadowAudioNode` role - it stores a copy of node's state.

Read: Simply read `HostObject` state
Write: Schedule modification and modify `Shadow State` 

Lock free approach!!!

TODO:
- AudioNode creation guide
- test AudioPlayer on Android
- lambda dtor
- buffer copy constraints

Nodes:
- [ ] OscillatorNode
- [ ] AnalyserNode
- [ ] BiquadFilterNode
- [ ] ConvolverNode
- [ ] IIRFilterNode
- [ ] PeriodicWave
- [ ] WaveShaper
- [ ] AudioNode
- [ ] AudioContext
- [ ] BaseAudioContext
- [ ] OfflineAudioContext

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferSourceNodeHostObject.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferSourceNodeHostObject.cpp
@@ -3,6 +3,7 @@
 #include <audioapi/HostObjects/utils/NodeOptions.h>
 #include <audioapi/core/BaseAudioContext.h>
 #include <audioapi/core/sources/AudioBufferSourceNode.h>
+#include <audioapi/utils/AudioBus.h>
 
 #include <memory>
 #include <utility>
@@ -17,6 +18,8 @@ AudioBufferSourceNodeHostObject::AudioBufferSourceNodeHostObject(
     loopStart_(options.loopStart),
     loopEnd_(options.loopEnd),
     buffer_(std::make_shared<AudioBufferHostObject>(options.buffer)) {
+   pitchCorrection_ = options.pitchCorrection;
+
   addGetters(
       JSI_EXPORT_PROPERTY_GETTER(AudioBufferSourceNodeHostObject, loop),
       JSI_EXPORT_PROPERTY_GETTER(AudioBufferSourceNodeHostObject, loopSkip),
@@ -149,20 +152,40 @@ JSI_HOST_FUNCTION_IMPL(AudioBufferSourceNodeHostObject, setBuffer) {
                         args[0].getObject(runtime).asHostObject<AudioBufferHostObject>(runtime);
 
 
-  std::shared_ptr<AudioBuffer> bufferCopy = nullptr;
+  std::shared_ptr<AudioBus> alignedBus = nullptr;
+  std::shared_ptr<AudioBus> audioBus = nullptr;
+  std::shared_ptr<AudioBus> playbackRateBus = nullptr;
 
   if (bufferHostObject != nullptr) {
     thisValue.asObject(runtime).setExternalMemoryPressure(
-            runtime, bufferHostObject->getSizeInBytes() + 16);
+              runtime, bufferHostObject->getSizeInBytes() + 16);
 
-    bufferCopy = std::make_shared<AudioBuffer>(*bufferHostObject->audioBuffer_);
+    auto buffer = bufferHostObject->audioBuffer_;
+
+    channelCount_ = buffer->getNumberOfChannels();
+    audioBus = std::make_shared<AudioBus>(RENDER_QUANTUM_SIZE, channelCount_, 48000.0f);
+    playbackRateBus = std::make_shared<AudioBus>(RENDER_QUANTUM_SIZE * 3, channelCount_, 48000.0f);
+
+      if (pitchCorrection_) {
+        int extraTailFrames =
+            static_cast<int>((inputLatency_ + outputLatency_) * buffer->getSampleRate());
+        size_t totalSize = buffer->getLength() + extraTailFrames;
+
+        alignedBus = std::make_shared<AudioBus>(totalSize, channelCount_, buffer->getSampleRate());
+        alignedBus->copy(buffer->bus_.get(), 0, 0, buffer->getLength());
+        alignedBus->zero(buffer->getLength(), extraTailFrames);
+      } else {
+        alignedBus = std::make_shared<AudioBus>(*buffer->bus_);
+      }
   }
 
   auto event = [
         audioBufferSourceNode,
-        buffer = bufferCopy
+        alignedBus,
+        audioBus,
+        playbackRateBus
   ](BaseAudioContext &) {
-    audioBufferSourceNode->setBuffer(buffer);
+    audioBufferSourceNode->setBuffer(alignedBus, audioBus, playbackRateBus);
   };
 
   audioBufferSourceNode->scheduleAudioEvent(std::move(event));
--- a/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferSourceNodeHostObject.h
+++ b/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferSourceNodeHostObject.h
@@ -35,7 +35,7 @@ class AudioBufferSourceNodeHostObject : public AudioBufferBaseSourceNodeHostObje
   JSI_HOST_FUNCTION_DECL(start);
   JSI_HOST_FUNCTION_DECL(setBuffer);
 
- private:
+ protected:
   bool loop_;
   bool loopSkip_;
   double loopStart_;
--- a/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferBaseSourceNodeHostObject.h
+++ b/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferBaseSourceNodeHostObject.h
@@ -28,13 +28,14 @@ class AudioBufferBaseSourceNodeHostObject : public AudioScheduledSourceNodeHostO
   JSI_HOST_FUNCTION_DECL(getInputLatency);
   JSI_HOST_FUNCTION_DECL(getOutputLatency);
 
- private:
+ protected:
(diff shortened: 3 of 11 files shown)


packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioEventScheduler.h
```diff
@@ -0,0 +1,39 @@
...
+ public:
+  /// @brief Constructs an AudioEventScheduler with the given context and capacity.
+  /// @param context The audio context associated with this scheduler.
+  /// @param capacity The maximum number of events that can be scheduled. Default is 1024.
+  explicit AudioEventScheduler(
+      const std::shared_ptr<BaseAudioContext> &context,
+      size_t capacity = 1024);
+  ~AudioEventScheduler() = default;
+
+  /// @brief Schedules an event to be processed.
```
poneciak57: schedules an event or processess it instantly if context is not running

packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.cpp
```diff
@@ -111,6 +111,14 @@ void AudioNode::disable() {
   }
 }
 
+bool AudioNode::scheduleAudioEvent(std::function<void(BaseAudioContext &)> &&event) noexcept {
```
poneciak57: it is better not to use `std::function` here it expects lambda to be copyable but we might sometimes want to pass move only types as for example jsi::Function is move only. Instead go with template<typename F> ... F&& and like add some type trait magic to check if its invokable from BaseAudioContext ref

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferSourceNodeHostObject.h
```diff
@@ -30,6 +31,13 @@ class AudioBufferSourceNodeHostObject : public AudioBufferBaseSourceNodeHostObje
 
   JSI_HOST_FUNCTION_DECL(start);
   JSI_HOST_FUNCTION_DECL(setBuffer);
+
+ private:
+  bool loop_;
+  bool loopSkip_;
+  double loopStart_;
+  double loopEnd_;
+  std::shared_ptr<AudioBufferHostObject> buffer_;
```
poneciak57: the thing with shadow node seems alright but it is not applicable to all nodes for example it is impossible to make same assumption for Params or GainNode as its value is modified in processing loop so you can only keep local cache of values updated by events idk if its a proper design i would personally stick to that we read from original node and value will be updated asynchronously

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioEventScheduler.cpp
```diff
@@ -0,0 +1,26 @@
...
+                                         size_t capacity): context_(context), eventScheduler_(capacity) {}
+
+bool AudioEventScheduler::scheduleEvent(std::function<void(BaseAudioContext &)> &&event) noexcept {
+    if (context_->isRunning()) {
+        return eventScheduler_.scheduleEvent(std::move(event));
+    } else {
+      event(*context_);
+      return true;
+    }
+}
```
poneciak57: It should be fine but one thing is missing. We drain the queue at the start of renderAudio callback so here look at this sequence
TA - audio thread
TJ - JS thread

1. TA - drains the queue
2. TJ - pushes E1 to queue (ctx running so pushes to queue)
3. TJ - stops context
4. TA - processes last frame
5. TJ - pushes E2 to queue (ctx not running imediately executes)

We get inconsistent state because our operations are not executed in proper order that depends on the order of their schedule. So E1 should be executed before E2 for example these two events might depend on each other like connect disconnect or something like allowing these two things is not alright or even worse we can get race condition if last frames is processed after the flag is set to true

And draining queue on stop can cause a race cond on the other hand. So this is major problem that needs to be solved.

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferSourceNodeHostObject.h
maciejmakowski2003: I think only value getter in `AudioParam` could be more complicated

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
```diff
@@ -41,15 +44,16 @@ bool AudioPlayer::openAudioStream() {
     return false;
   }
 
-  mBus_ = std::make_shared<AudioBus>(RENDER_QUANTUM_SIZE, channelCount_, sampleRate_);
+  audioBus_ = std::make_shared<AudioBus>(RENDER_QUANTUM_SIZE, channelCount_, sampleRate_);
   return true;
 }
 
 bool AudioPlayer::start() {
   if (mStream_) {
```
mdydek: we can get rid of all implicit stream to bool casts in this pr

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
```diff
@@ -99,17 +111,19 @@ AudioPlayer::onAudioReady(AudioStream *oboeStream, void *audioData, int32_t numF
...
 
-    // TODO: optimize this with SIMD?
+    if (isRunning_.load(std::memory_order_acquire)) {
+      renderAudio_(audioBus_, framesToProcess);
+    } else {
+      audioBus_->zero();
+    }
+
     for (int i = 0; i < framesToProcess; i++) {
       for (int channel = 0; channel < channelCount_; channel += 1) {
```
mdydek:
```suggestion
    for (int i = 0; i < framesToProcess; i++) {
      for (int channel = 0; channel < channelCount_; channel++) {
```
consistency

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferSourceNodeHostObject.cpp
```diff
@@ -32,110 +35,143 @@ AudioBufferSourceNodeHostObject::AudioBufferSourceNodeHostObject(
...
-  return {loopEnd};
+  return {loopEnd_};
 }
 
 JSI_PROPERTY_SETTER_IMPL(AudioBufferSourceNodeHostObject, loop) {
   auto audioBufferSourceNode = std::static_pointer_cast<AudioBufferSourceNode>(node_);
-  audioBufferSourceNode->setLoop(value.getBool());
+  auto loop = value.getBool();
+
+  auto event = [audioBufferSourceNode, loop](BaseAudioContext &context) {
```
mdydek: do we need reference to context here and in all of the lambdas that they do not use context inside it? if not change sfinae definition of a function so more types can be passed in event scheduler

packages/react-native-audio-api/common/cpp/audioapi/core/BaseAudioContext.cpp
```diff
@@ -35,9 +35,12 @@
 namespace audioapi {
 
 BaseAudioContext::BaseAudioContext(
+    float sampleRate,
     const std::shared_ptr<IAudioEventHandlerRegistry> &audioEventHandlerRegistry,
     const RuntimeRegistry &runtimeRegistry)
-    : nodeManager_(std::make_shared<AudioNodeManager>()),
+    : sampleRate_ {sampleRate},
+      nodeManager_(std::make_shared<AudioNodeManager>()),
+      audioEventScheduler_(std::make_unique<CrossThreadEventScheduler<BaseAudioContext>>(1024)),
```
mdydek: magic number, put it in some constant

packages/react-native-audio-api/common/cpp/audioapi/utils/FatFunction.hpp
```diff
@@ -0,0 +1,145 @@
...
+          other.mover_(storage_.data(), other.storage_.data());
+        } else {
+          std::memcpy(storage_.data(), other.storage_.data(), N);
+        }
+        invoker_ = other.invoker_;
+        deleter_ = other.deleter_;
+        mover_ = other.mover_;
+        other.reset();
+      }
+    }
```
mdydek, line 102:
```suggestion
  /// @brief Move constructor
  /// @param other
  FatFunction(FatFunction &&other) noexcept {
    if (other.invoker_) {
      if (other.mover_) {
        other.mover_(storage_.data(), other.storage_.data());
      } else {
        std::memcpy(storage_.data(), other.storage_.data(), N);
      }
      invoker_ = other.invoker_;
      deleter_ = other.deleter_;
      mover_ = other.mover_;
      other.reset();
    }
  }

  /// @brief Move assignment operator
  /// @param other
  FatFunction &operator=(FatFunction &&other) noexcept {
    if (this != &other) {
      reset();
      new (this) FatFunction(std::move(other));
      }
    }
```
code duplication

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferSourceNodeHostObject.cpp
maciejmakowski2003: actually we are processing all events by passing the some argument for all of them, so it wont be possible in that approach

packages/react-native-audio-api/common/cpp/audioapi/events/AudioEventHandlerRegistry.h
poneciak57, line 1: I would mark this as something that we need to take a look at. It would be very good to be 100% sure that it is thread safe.

packages/react-native-audio-api/common/cpp/audioapi/utils/FatFunction.hpp
poneciak57: Uuuuu 🆒

PR #885, opened 2026-01-15
refactor: last improvements of notifications

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- recording notification persists state better
- better docs
- better js docs

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/audiodocs/docs/system/audio-manager.mdx
+++ b/packages/audiodocs/docs/system/audio-manager.mdx
@@ -242,7 +242,7 @@ interface EventTypeWithValue {
 }
 
 interface OnInterruptionEventType {
-  type: 'ended' | 'began'; //if interruption event has started or ended
+  type: 'ended' | 'began'; // if interruption event has started or ended
   shouldResume: boolean; // if the interruption was temporary and we can resume the playback/recording
 }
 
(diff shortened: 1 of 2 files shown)


packages/audiodocs/docs/system/audio-manager.mdx
```diff
@@ -242,7 +242,7 @@ interface EventTypeWithValue {
 }
 
 interface OnInterruptionEventType {
-  type: 'ended' | 'began'; / /if interruption event has started or ended
+  type: 'ended' | 'began'; //if interruption event has started or ended
```
michalsek:
```suggestion
  type: 'ended' | 'began'; // if interruption event has started or ended
```

packages/audiodocs/docs/system/recording-notification-manager.mdx
```diff
@@ -42,9 +42,14 @@ RecordingNotificationManager.hide();
 
 Shows the recording notification with the parameters.
 
+:::info
+Metadata is remembered between calls, so after initial passing the metadata to show function, you can only call it with elements that are supposed to change.
```
maciejmakowski2003:
```suggestion
Metadata is saved between calls, so after the initial pass to the show method, you need only call it with elements that are supposed to change.
```

maciejmakowski2003 (changes requested):

PR #891, opened 2026-01-16
Fix/return session activation errors

michalsek:
Closes #

## ⚠️ Breaking changes ⚠️

- `setAudioSessionActivity` now throws on failure attempts

## Introduced changes

- retruns (through promise reject) a error code and additional details about why the activity failed

## Testing

- Easiest way to repro the error path, is to run some voice/video call app (f.e. google meet) then try to start the playback or recording from within the fabric example. It should throw the insufficient priority error in this case

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/ios/audioapi/ios/AudioAPIModule.mm
+++ b/packages/react-native-audio-api/ios/audioapi/ios/AudioAPIModule.mm
@@ -129,16 +129,18 @@ RCT_EXPORT_METHOD(
 
     if (!success) {
       NSDictionary *meta = @{
-        @"nativeCode": @(error.code),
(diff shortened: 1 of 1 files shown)


packages/react-native-audio-api/ios/audioapi/ios/system/AudioEngine.mm
```diff
@@ -182,7 +182,9 @@ - (bool)startEngine
     return true;
   }
 
-  if (![self.sessionManager setActive:true]) {
+  if (![self.sessionManager setActive:true error:&error]) {
+    // TODO: return user facing error
```
michalsek, line 186: added to board

PR #898, opened 2026-01-19
Allow to override filename in recorder start method

michalsek:
Closes #

## Introduced changes

- allows to override autogenerated file name

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/inputs/AudioRecorder.h
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/inputs/AudioRecorder.h
@@ -25,7 +25,7 @@ class AudioRecorder {
       : audioEventHandlerRegistry_(audioEventHandlerRegistry) {}
   virtual ~AudioRecorder() = default;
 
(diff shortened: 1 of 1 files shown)


packages/react-native-audio-api/common/cpp/audioapi/core/inputs/AudioRecorder.h
```diff
@@ -25,7 +25,7 @@ class AudioRecorder {
       : audioEventHandlerRegistry_(audioEventHandlerRegistry) {}
   virtual ~AudioRecorder() = default;
 
-  virtual Result<std::string, std::string> start() = 0;
+  virtual Result<std::string, std::string> start(const std::string &fileNameOverride = "") = 0;
```
mdydek: imo pure virtual should not have defaults in its implementation
```cpp
class Base {
public:
    virtual void display(int value = 10) = 0; 
};

class Derived : public Base {
public:
    void display(int value = 20) override { 
        std::cout << value << std::endl; 
    }
};

Base* ptr = new Derived();
ptr->display(); // Outputs: 10
```

PR #905, opened 2026-01-21
Refactor/recorder event to separate thread

mdydek:
Closes RNAA-386

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- offloaded invocation of `onAudioReady` and resampling data to a different thread

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/utils/TaskOffloader.hpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/utils/TaskOffloader.hpp
@@ -10,7 +10,7 @@ using namespace audioapi::channels::spsc;
 namespace audioapi::task_offloader {
 
 /// @brief A utility class to offload task to a separate thread using a SPSC channel.
-/// @tparam T The type of data to be sent through the channel. Must be DefaultConstructible.
+/// @tparam T The type of data to be sent through the channel. Must be default_initializable.
 /// @tparam Strategy The overflow strategy for the SPSC channel.
 /// @tparam Wait The wait strategy for the SPSC channel.
 template <std::default_initializable T, OverflowStrategy Strategy, WaitStrategy Wait>


packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/utils/AndroidRecorderCallback.cpp
```diff
@@ -108,19 +118,27 @@ void AndroidRecorderCallback::cleanup() {
...
     processingBufferLength_ = 0;
   }
 
   for (size_t i = 0; i < circularBus_.size(); ++i) {
     circularBus_[i]->zero();
   }
+
+  stopCallbackThread_.store(true, std::memory_order_release);
+  // Send a dummy/empty message to unblock the receiver_.receive()
+  sender_.send({nullptr, 0});
```
poneciak57: very good 👍

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/utils/AndroidRecorderCallback.cpp
```diff
@@ -175,4 +170,36 @@ void AndroidRecorderCallback::deinterleaveAndPushAudioData(void *data, int numFr
...
+    ma_data_converter_process_pcm_frames(
+        converter_.get(), data, &inputFrameCount, processingBuffer_, &outputFrameCount);
+
+    deinterleaveAndPushAudioData(processingBuffer_, static_cast<int>(outputFrameCount));
+
+    if (circularBus_[0]->getNumberOfAvailableFrames() >= bufferLength_) {
+      emitAudioData();
+    }
+  }
+}
```
poneciak57: most handlers are really similar and there is some code duplication, also there are some possible subtle bugs as sending dummy data to unblock worker sleeping on spsc.

Maybe you can design some kind of an abstraction over the handler. HandlerManager would be responsible for stoping starting and lifetime and logic would be in some class implementing BaseHandler

packages/react-native-audio-api/common/cpp/audioapi/utils/TaskOffloader.hpp
```diff
@@ -0,0 +1,69 @@
+#pragma once
+
+#include <audioapi/utils/SpscChannel.hpp>
+#include <cassert>
+#include <concepts>
+#include <utility>
+
+template <typename T>
+concept DefaultConstructible = std::default_initializable<T>;
```
poneciak57: https://en.cppreference.com/w/cpp/named_req/DefaultConstructible.html

packages/react-native-audio-api/common/cpp/audioapi/utils/TaskOffloader.hpp
```diff
@@ -0,0 +1,69 @@
...
+    shouldRun_.store(true, std::memory_order_release);
+    workerThread_ = std::thread([this, task = std::forward<Func>(task)]() {
+      while (shouldRun_.load(std::memory_order_acquire)) {
+        auto data = receiver_.receive();
+        if (shouldRun_.load(std::memory_order_acquire)) {
+          task(std::move(data));
+        }
+      }
+    });
+  }
```
poneciak57, line 33: what if u write
```
to.offloadTask(task1);
to.offloadTask(task2);
```

it will spawn two threads waiting on same spsc that would result in undefined behaviour

packages/react-native-audio-api/common/cpp/audioapi/utils/TaskOffloader.hpp
poneciak57, line 1: that seems unsafe:

1. U are not sure that user will pass the proper sender to stop which can cause deadlock.
2. As mentioned in other comment user is able to call offloadTask multiple times resulting in ub

What i would recommend is:
- move worker thread construction to constructor so u make sure 1 thread per task
- keep sender in class as member to make sure u have right sender
- you might want to send std::optional as payload std::nullopt would mark the stop
- you should also stop in desctructor as throwing in destructor is not a good practice
- for sure not use default T as stop mark because u also require task to handle that stop mark gracefully which might be very unclear and cause some logical bugs
- also i would hardcode the template arguments for WaitStrategy and OverflowStrategy as they will be mostly the same

mdydek: How can I pass `std::optional<T>` in sender, when `send` function expects `T&&`, it will not compile. I do not agree with hardcoding template arguments, there are many problems that would f.e. require `WAIT_ON_FULL` overflow strategy (separate sender thread that can be blocked). And I think that passing default T is not wrong also, because 
```        
if (shouldRun_.load(std::memory_order_acquire)) {
   task(std::move(data));
}
```
simple flag check would not require calling task with some ambiguous value.

PR #939, opened 2026-02-05
feat: graph refactor chapter I

poneciak57:
Closes #

Base for many features that can not be implemented with current graph processing structure. This refactor does not change anything. It aims to implement new structure for graph processing which will allow:
- assynchronous fast updates
- multiple outputs from node
- cycles with delay node
- performance improvement
- it will allow to get rid of all allocations on audio thread
- it will allow to utilize fully asynchronous communication between JS and audio threads

This pull requests contains only not integrated implementations and a lot of unit tests

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Updated relevant documentation
- [ ] Added/Conducted relevant tests
- [ ] Performed self-review of the code
- [ ] Updated Web Audio API coverage
- [ ] Added support for web
- [ ] Updated old arch android spec file

PR #942, opened 2026-02-06
Refactor/communication between audio and js thread

maciejmakowski2003:
Closes RNAA-311
Closes RNAA-250
Closes RNAA-410
Part of RNAA-433

## ⚠️ Breaking changes ⚠️

- Replaced `initialize` method from `StreamerNode` with passing url via ctor or factory method.
- Drop support for hann window in `AnalyserNode`

## Introduced changes

- Enhanced communication between JS thread and Audio thread. Almost all of node's modifications from JS thread are scheduled on `CrossThreadEventScheduler` wrapper to be executed on Audio thread. `HostObject` plays a `ShadowAudioNode` role - it stores a copy of node's state. There are some exceptions like using atomics ot triple buffer pattern.

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web


ghdocs/audio-node.md
```diff
@@ -0,0 +1,280 @@
+# How to create new AudioNode
+
+In this docs we present recommended design pattern for creating new AudioNode.
```
mdydek:
```suggestion
In this docs we present recommended patterns for creating new AudioNodes.
```

ghdocs/audio-node.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Layers
+
+In common each AudioNode has three layers:
```
mdydek:
```suggestion
Usually, each AudioNode has three layers:
```

ghdocs/audio-node.md
```diff
@@ -0,0 +1,280 @@
...
+- Typescript (JS)
+
+  Elegant typescript wrapper around JS HO interface.
+
+```ts
+class GainNode extends AudioNode {
+  readonly gain: AudioParam;
+
+  constructor(context: BaseAudioContext, options?: TGainOptions) {
+    const gainNode: IGainNode = context.context.createGain(options || {});
```
mdydek: emphasize that `context.context` is underlying c++ object

ghdocs/audio-node.md
```diff
@@ -0,0 +1,280 @@
...
+  auto event = [analyserNode, minDecibels](BaseAudioContext&) {
+    analyserNode->setMinDecibels(minDecibels);
+  };
+    analyserNode->scheduleAudioEvent(std::move(event));
+    minDecibels_ = minDecibels;
+}
+```
+
+
+1. **function**
```
mdydek:
```suggestion
3. **function**
```

ghdocs/audio-node.md
```diff
@@ -0,0 +1,280 @@
...
+
+C++ counterpart is `JSI_HOST_FUNCTION`. It is a common function that can receive arguments and return some value.
+
+```ts
+const fftOutput = new Uint8Array(analyser.frequencyBinCount);
+analyserNode.getByteFrequencyData(fftOutput);
+```
+
+```cpp
+SI_HOST_FUNCTION_IMPL(AnalyserNodeHostObject, getByteFrequencyData) {
```
mdydek:
```suggestion
JSI_HOST_FUNCTION_IMPL(AnalyserNodeHostObject, getByteFrequencyData) {
```

ghdocs/audio-node.md
```diff
@@ -0,0 +1,280 @@
...
+    : /* ... */ {
+  addGetters(JSI_EXPORT_PROPERTY_GETTER(AnalyserNodeHostObject, fftSize));
+  addSetters(JSI_EXPORT_PROPERTY_SETTER(AnalyserNodeHostObject, fftSize));
+  addFunctions(JSI_EXPORT_FUNCTION(AnalyserNodeHostObject, getByteFrequencyData),);
+}
+```
+
+#### Shadow state (C++)
+
+Shadow state is mechanism introduced in order to make communication between JS and Audio thread lock-free. AudioNodeHostObject stores the set of properties modified only by JS thread (the same set C++ AudioNode has). Everytime we want to access some property from JS, we can just return property from shadow state, when we modify some property we have to update shadow state and schedule update event on Audio Event Loop (SPSC). By following that manner we can skip accessing AudioNode state, that is also accessed by Audio Thread - no need to lock or use atomic variables.
```
mdydek:
```suggestion
Shadow state is a mechanism introduced in order to make communication between JS and Audio thread lock-free. AudioNodeHostObject stores the set of properties, which are modified only by JS thread (the same set C++ AudioNode has). Everytime we want to access some property from JS, we can just return property from shadow state, when we modify some property we have to update shadow state and schedule update event on Audio Event Loop (SPSC). By following that manner we can skip accessing AudioNode state, that is also accessed by Audio Thread - no need to lock or use atomic variables.
```

ghdocs/audio-node.md
```diff
@@ -0,0 +1,280 @@
...
+}
+```
+
+There are a few rules that should be followed when implementing C++ AudioNode core.
+
+- **Thread safety**: Each AudioNode should be created in thread safe manner.
+
+- **Heap allocations**: Heap allocations are not allowed on Audio Thread, so all necessary data should be allocated in constructor, or pre-allocated on other thread and passed to AudioNode.
+
+- **Destructions** No destructions are allowed to happen on Audio Thread. In common AudioNode destruction are handled by already active AudioDestructor. If you need to perform some cleanup, you have to delegate it to AudioDestructor.
```
mdydek:
```suggestion
- **Destructions** No destructions are allowed to happen on Audio Thread. AudioNode destruction are handled by already active AudioDestructor. If you need to perform some cleanup, you have to delegate it to AudioDestructor.
```

ghdocs/audio-node.md
```diff
@@ -0,0 +1,280 @@
...
+
+  oscillatorNode->scheduleAudioEvent(std::move(event));
+}
+```
+
+#### Communication between JS Thread and Audio Thread
+
+**getters** and **setters**
+
+1. Property is primitive and is not modified by Audio Thread.
```
mdydek: `the` before each `Audio Thread`

ghdocs/audio-node.md
```diff
@@ -0,0 +1,280 @@
...
+
+   Shadow state design pattern should be followed.
+
+2. Property is not primitive and is not modified by Audio Thread.
+
+  In common it should be stored in TS layer and copied to AudioNode
```
mdydek:
```suggestion
  It should be stored in TS layer and copied to AudioNode
```

ghdocs/audio-node.md
```diff
@@ -0,0 +1,280 @@
...
+    auto event = [param = param_, value = static_cast<float>(value.getNumber())](BaseAudioContext &) {
+        param->setValue(value);
+    };
+
+    param_->scheduleAudioEvent(std::move(event));
+}
+```
+
+4. Property is not primitive and can be modified by Audio Thread.
+  In C++ core triple buffer pattern should be followed. It allows to have one copy of property for reader, one for writer and one for pending update. On each update we just swap pending update with writer, and on each read we just read from reader. In that manner we can skip locks and just operate on atomic indices.
```
mdydek, line 274: add link for further info about triple buffer pattern

packages/react-native-audio-api/common/cpp/audioapi/core/analysis/AnalyserNode.cpp
mdydek: we can use `std::clamp` here

packages/react-native-audio-api/common/cpp/audioapi/core/analysis/AnalyserNode.cpp
```diff
@@ -18,73 +16,36 @@ AnalyserNode::AnalyserNode(
...
 }
 
 void AnalyserNode::setMinDecibels(float minDecibels) {
-  minDecibels_ = minDecibels;
+  minDecibels_.store(minDecibels, std::memory_order_release);
 }
 
 void AnalyserNode::setMaxDecibels(float maxDecibels) {
-  maxDecibels_ = maxDecibels;
+  maxDecibels_.store(maxDecibels, std::memory_order_release);
```
mdydek: I don't see where `minDecibels_` and `maxDecibels_` can be accessed outside of js thread, thus there is no necessity for atomic variables

packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.h
```diff
@@ -39,10 +37,31 @@ class AudioNode : public std::enable_shared_from_this<AudioNode> {
       int framesToProcess,
       bool checkIsAlreadyProcessed);
 
+  float getContextSampleRate() const {
+    if (std::shared_ptr<BaseAudioContext> context = context_.lock()) {
+      return context->getSampleRate();
+    }
+
+    return DEFAULT_SAMPLE_RATE;
+  }
```
mdydek, line 45: I don't like this idea, it can be very misleading since system can output correct amount even though context is not valid

packages/react-native-audio-api/common/cpp/audioapi/core/effects/BiquadFilterNode.h
```diff
@@ -71,35 +70,34 @@ class BiquadFilterNode : public AudioNode {
...
+  static FilterCoefficients setBandpassCoefficients(float frequency, float Q);
+  static FilterCoefficients setLowshelfCoefficients(float frequency, float gain);
+  static FilterCoefficients setHighshelfCoefficients(float frequency, float gain);
+  static FilterCoefficients setPeakingCoefficients(float frequency, float Q, float gain);
+  static FilterCoefficients setNotchCoefficients(float frequency, float Q);
+  static FilterCoefficients setAllpassCoefficients(float frequency, float Q);
+  static FilterCoefficients
+  setNormalizedCoefficients(float b0, float b1, float b2, float a0, float a1, float a2);
+  FilterCoefficients
+  applyFilter(float frequency, float Q, float gain, float detune, BiquadFilterType type);
```
mdydek, line 103: do we need `type` as parameter? In all use cases it is property of a class, so we can retrieve its value inside a function

packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.h
```diff
@@ -39,10 +37,31 @@ class AudioNode : public std::enable_shared_from_this<AudioNode> {
...
+
+  /// @note JS Thread only
   bool isEnabled() const;
+  /// @note JS Thread only
   bool requiresTailProcessing() const;
-  void enable();
-  virtual void disable();
+
+  template <typename F>
+  bool inline scheduleAudioEvent(F &&event) noexcept {
```
mdydek, line 57: add more constraints what can and what cannot be passed as an event, right now we can pass this event
```cpp
auto event = []() {
   return;
 };
```
but it will fail, because context will call this event like that 
```cpp
event(*this);
```

ghdocs/audio-node.md
poneciak57, line 1: It would be nice to update this doc in future and include few repeating solutions for most common problems that are used in our codebase aka:
- prealocating
- SPSC with worker thread
- how to communicate Audio <-> JSI

Also it is worth noting that we do not block syscalls or allocations on audio thread it is just strongly unrecommended to do but in case if someone is crafting custom node processor for his custom needs it might work fine

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferQueueSourceNode.h
```diff
@@ -20,21 +20,26 @@ class AudioBufferQueueSourceNode : public AudioBufferBaseSourceNode {
...
+      size_t bufferId,
+      const std::shared_ptr<AudioBuffer> &tailBuffer);
   void dequeueBuffer(size_t bufferId);
   void clearBuffers();
   void disable() override;
 
   void setOnBufferEndedCallbackId(uint64_t callbackId);
 
+  /// @note Thread safe, because does not access state of the node.
+  void unregisterOnBufferEndedCallback(uint64_t callbackId);
```
poneciak57: it does not but might be inheritly unsafe through registry handler

packages/react-native-audio-api/ios/audioapi/ios/core/IOSAudioRecorder.mm
```diff
@@ -50,7 +50,7 @@
 
     if (isConnected()) {
       if (auto lock = Locker::tryLock(adapterNodeMutex_)) {
```
poneciak57, line 52: lock missed ?

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferQueueSourceNode.h
poneciak57: its thread safety depends on the thread safety of eventRegistryHandle. Besides these methods do not need to be thread safe as they are used on AudioThread through scheduled async events

packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.h
```diff
@@ -33,7 +33,7 @@ class AudioContext : public BaseAudioContext {
 #else
   std::shared_ptr<IOSAudioPlayer> audioPlayer_;
 #endif
-  bool isInitialized_;
+  std::atomic<bool> isInitialized_ = false;
```
poneciak57: it might be hard to swallow but i doubt it should be atomic as for node to be connected to graph thus processed it needs to be sent through spsc ... sooo its creation is done on the JS thread

packages/react-native-audio-api/src/interfaces.ts
```diff
@@ -101,7 +100,7 @@ export interface IBaseAudioContext {
   ) => IPeriodicWave;
   createAnalyser: (analyserOptions: AnalyserOptions) => IAnalyserNode;
   createConvolver: (convolverOptions?: IConvolverOptions) => IConvolverNode;
-  createStreamer: (streamerOptions?: StreamerOptions) => IStreamerNode | null; // null when FFmpeg is not enabled
+  createStreamer: (streamerOptions: StreamerOptions) => IStreamerNode | null; // null when FFmpeg is not enabled
```
mdydek, line 103: update docs accordingly

packages/react-native-audio-api/common/cpp/audioapi/core/BaseAudioContext.cpp
```diff
@@ -42,7 +42,10 @@ BaseAudioContext::BaseAudioContext(
       sampleRate_(sampleRate),
       graphManager_(std::make_shared<AudioGraphManager>()),
       audioEventHandlerRegistry_(audioEventHandlerRegistry),
-      runtimeRegistry_(runtimeRegistry) {}
+      runtimeRegistry_(runtimeRegistry),
+      audioEventScheduler_(
+          std::make_unique<CrossThreadEventScheduler<BaseAudioContext>>(AUDIO_SCHEDULER_CAPACITY)) {
+}
```
poneciak57: does it need to be unique_ptr?

packages/react-native-audio-api/common/cpp/audioapi/utils/TripleBuffer.hpp
poneciak57, line 1: looks really good but few neatpicks and overall good job but write tests for it

packages/react-native-audio-api/common/cpp/audioapi/utils/TripleBuffer.hpp
```diff
@@ -0,0 +1,49 @@
...
+ private:
+  struct State {
+    int index;
+    bool hasUpdate;
+  };
+
+  T buffers_[3];
+  int frontIndex_ = 0;
+  std::atomic<State> state_{{1, false}};
+  int backIndex_ = 2;
```
poneciak57: False sharing occurs here add alignas as in SPSC to prevent it. Also it might be better to use size_t instead of int but that is minor thing

packages/react-native-audio-api/common/cpp/audioapi/utils/TripleBuffer.hpp
```diff
@@ -0,0 +1,49 @@
...
+    return &buffers_[frontIndex_];
+  }
+
+ private:
+  struct State {
+    int index;
+    bool hasUpdate;
+  };
+
+  T buffers_[3];
```
poneciak57: buffers should also be alligned to cache line sizes, and u can maybe enchance it to use emplace new operator to allow T types which are not defaultly constructible

packages/react-native-audio-api/common/cpp/audioapi/utils/TripleBuffer.hpp
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    return &buffers_[frontIndex_];
+  }
+
+ private:
+  struct State {
+    int index;
+    bool hasUpdate;
+  };
```
poneciak57, line 72: use bitfields to fit this in 32 bits

packages/react-native-audio-api/common/cpp/audioapi/utils/TripleBuffer.hpp
```diff
@@ -0,0 +1,49 @@
...
+ public:
+  T *getForWriter() {
+    return &buffers_[backIndex_];
+  }
+
+  void publish() {
+    State newState{backIndex_, true};
+    auto prevState = state_.exchange(newState, std::memory_order_acq_rel);
+    backIndex_ = prevState.index;
+  }
```
poneciak57, line 51: it might have interface read(), write() instead of this it would be better for RAII i guess as u never forget to publish

packages/react-native-audio-api/common/cpp/audioapi/utils/TripleBuffer.hpp
poneciak57: and i am kind of against puting something like this in such a big PR, cuz whole PR LGTM besides this part

ghdocs/audio-node.md
poneciak57: 🤓

packages/react-native-audio-api/ios/audioapi/ios/core/IOSAudioRecorder.mm
maciejmakowski2003: I did not want to touch other files than nodes and core (AudioParam, contexts) in this PR. We can add separate issue for getting rid off other locks

ghdocs/audio-node.md
maciejmakowski2003: I am going to add it in skills. Actually whole internal docs can be moved to skills file in appropriate subdirectory -> separate issue

packages/react-native-audio-api/common/cpp/audioapi/utils/TripleBuffer.hpp
maciejmakowski2003: 🫡

packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.h
maciejmakowski2003: I am not sure if it will ever return default value. It can happen only in destruction phase in which context will not play, so it does not really matter what it returns.

packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.h
maciejmakowski2003: hmm for context it is not true. isInitialized can be accessed from both threads:
- `AudioScheduledSourceNode::start` - Audio Thread
- `AudioContext::resume` - JS Thread

packages/react-native-audio-api/common/cpp/audioapi/core/AudioNode.h
maciejmakowski2003: it is not true

packages/react-native-audio-api/common/cpp/audioapi/utils/TripleBuffer.hpp
maciejmakowski2003: do you have any specific design in mind?

packages/react-native-audio-api/common/cpp/audioapi/utils/TripleBuffer.hpp
maciejmakowski2003: added tests

packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.h
poneciak57: oh its on audio context not audio node
that is true my bad

packages/react-native-audio-api/common/cpp/audioapi/utils/TripleBuffer.hpp
poneciak57: void write(T&& val)
T& read()

something like that

poneciak57: this way u delegate construction of T out of this class and it allows more flexibility. JSI types are not defaultly constructible

maciejmakowski2003: So it will not be possible to do some custom operations in order to copy val to triple buffer? In addition we will need some preallocated object of type T to pass it, right?

packages/react-native-audio-api/common/cpp/audioapi/utils/TripleBuffer.hpp
```diff
@@ -0,0 +1,71 @@
...
+/// @tparam T The type of the buffer.
+template <typename T>
+class TripleBuffer {
+ public:
+  template <typename... Args>
+  explicit TripleBuffer(Args &&...args) {
+    new (&buffers_[0]) T(args...);
+    new (&buffers_[1]) T(args...);
+    new (&buffers_[2]) T(args...);
+  }
```
poneciak57, line 29: u don't need to initialize those in constructor thats the point. You want to keep the raw memory call constructor only when u need to

packages/react-native-audio-api/common/cpp/test/src/utils/TripleBufferTest.cpp
poneciak57, line 1: > [!NOTE]
> u can use later the guard that i introduced in #939 to check for disarm locks and context switches

packages/audiodocs/docs/sources/streamer-node.mdx
```diff
@@ -21,18 +21,18 @@ Similar to all of `AudioScheduledSourceNodes`, it can be started only once. If y
...
-constructor(context: BaseAudioContext, options?: StreamerOptions)
+constructor(context: BaseAudioContext, options: StreamerOptions)
 ```
 
 ### `StreamerOptions`
 
 | Parameter | Type | Default | |
 | :---: | :---: | :----: | :---- |
-| `streamPath` <Optional /> | `string` | - | Initial value for [`streamPath`](/docs/sources/streamer-node#properties) |
+| `streamPath` <Optional /> | `string` | - | Value for [`streamPath`](/docs/sources/streamer-node#properties) |
```
mdydek: streamPath is no longer optional

packages/react-native-audio-api/common/cpp/audioapi/utils/TripleBuffer.hpp
```diff
@@ -0,0 +1,74 @@
...
```
poneciak57, line 29: why still have args?

maciejmakowski2003: I have decided to leave it and added appropriate concept checking copyability of args and constructability of type

packages/react-native-audio-api/common/cpp/audioapi/core/analysis/AnalyserNode.cpp
```diff
@@ -2,9 +2,7 @@
 #include <audioapi/core/analysis/AnalyserNode.h>
 #include <audioapi/dsp/AudioUtils.hpp>
 #include <audioapi/dsp/VectorMath.h>
-#include <audioapi/dsp/Windows.hpp>
```
mdydek, line 5: we don't use Hann Window, so I think it can be deleted from this file completely

packages/react-native-audio-api/common/cpp/audioapi/core/analysis/AnalyserNode.cpp
```diff
@@ -18,73 +16,28 @@ AnalyserNode::AnalyserNode(
...
-}
-
-void AnalyserNode::setWindowType(AnalyserNode::WindowType type) {
-  setWindowData(type, fftSize_);
+  fft_ = std::make_unique<dsp::FFT>(fftSize);
+  complexData_ = std::vector<std::complex<float>>(fftSize);
+  magnitudeArray_ = std::make_unique<AudioArray>(fftSize / 2);
+  tempArray_ = std::make_unique<AudioArray>(fftSize);
+  windowData_ = std::make_unique<AudioArray>(fftSize);
+  dsp::Blackman().apply(windowData_->span());
```
mdydek: i think its a good place to refactor this, to me, class is overkill, we don't use any state inside of it, neither reuse it, we can implement it as simple function maybe with default values

packages/react-native-audio-api/common/cpp/audioapi/core/effects/BiquadFilterNode.h
```diff
@@ -83,23 +85,22 @@ class BiquadFilterNode : public AudioNode {
...
+  static FilterCoefficients setLowpassCoefficients(float frequency, float Q);
+  static FilterCoefficients setHighpassCoefficients(float frequency, float Q);
+  static FilterCoefficients setBandpassCoefficients(float frequency, float Q);
+  static FilterCoefficients setLowshelfCoefficients(float frequency, float gain);
+  static FilterCoefficients setHighshelfCoefficients(float frequency, float gain);
+  static FilterCoefficients setPeakingCoefficients(float frequency, float Q, float gain);
+  static FilterCoefficients setNotchCoefficients(float frequency, float Q);
+  static FilterCoefficients setAllpassCoefficients(float frequency, float Q);
+  static FilterCoefficients
+  setNormalizedCoefficients(float b0, float b1, float b2, float a0, float a1, float a2);
```
mdydek: names of these method are kind of misleading, because they do not set anything, they just return proper coefficients

packages/react-native-audio-api/common/cpp/audioapi/core/effects/IIRFilterNode.h
```diff
@@ -72,5 +74,18 @@ class IIRFilterNode : public AudioNode {
...
   }
+
+  static AudioArray createNormalizedArray(
+      const std::vector<float> &inputVector,
+      float scaleFactor) {
+    AudioArray result(inputVector.data(), inputVector.size());
+    if (scaleFactor != 1.0f && scaleFactor != 0.0f && result.getSize() > 0) {
+      for (float &val : result) {
+        val /= scaleFactor;
+      }
```
mdydek:
```suggestion
result.scale(1.0f / scaleFactor);
```

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferSourceNode.cpp
```diff
@@ -73,40 +45,43 @@ void AudioBufferSourceNode::setLoopEnd(double loopEnd) {
...
-  } else {
-    alignedBuffer_ = std::make_shared<AudioBuffer>(*buffer_);
+  if (buffer == nullptr) {
+    loopEnd_ = 0;
+    channelCount_ = 1;
+
+    buffer_ = nullptr;
+    playbackRateBuffer_ = nullptr;
+    audioBuffer_ =
+        std::make_shared<AudioBuffer>(RENDER_QUANTUM_SIZE, channelCount_, context->getSampleRate());
```
mdydek: why do we create a new one node buffer here?

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferQueueSourceNodeHostObject.cpp
```diff
@@ -25,40 +27,37 @@ AudioBufferQueueSourceNodeHostObject::AudioBufferQueueSourceNodeHostObject(
...
-
   auto audioBufferQueueSourceNode = std::static_pointer_cast<AudioBufferQueueSourceNode>(node_);
 
-  if (!args[1].isNumber()) {
-    audioBufferQueueSourceNode->start(when);
-  } else {
-    auto offset = args[1].getNumber();
+  auto event = [audioBufferQueueSourceNode,
+                when = args[0].getNumber(),
+                offset = args[1].isNumber() ? args[1].getNumber() : -1](BaseAudioContext &) {
```
mdydek: there is a difference in check convention, in the first argument we assume that it will be a number, in the second one we explicitly check, so are we sure and we can stick with first approach or not and stick to a 2nd one

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferSourceNode.cpp
maciejmakowski2003: because, `audioBuffer` is AudioNode property used for mixing across the graph. If user call `setBuffer(null)` after previously setting buffer, `AudioBufferSourceNode` should have `audioBuffer` with one channel.

WebAudioAPI reference -> `The number of channels of the output equals the number of channels of the AudioBuffer assigned to the [buffer](https://webaudio.github.io/web-audio-api/#dom-audiobuffersourcenode-buffer) attribute, or is one channel of silence if [buffer](https://webaudio.github.io/web-audio-api/#dom-audiobuffersourcenode-buffer) is null.`

maciejmakowski2003: btw it was allocation on Audio Thread that I missed to remove. good catch!!!

PR #954, opened 2026-02-19
feat: android notification improvement

zocario:
## ⚠️ Breaking changes ⚠️

- Remove `pauseIconResourceName` and `resumeIconResourceName` attributes from recording notification

## Introduced changes

- Remove custom layout of notifications on Android, as it wasn't fitting with other default notifications of the system.
- Display the notification silently but sets high priority so it is visible in the notification center
- Use standard action button instead of custom button to match the system layout of notifications.

### Before

Screenshot_20260212-160758 (2)

### After

Screenshot_20260219-110926
Screenshot_20260219-110930

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web


mdydek (changes requested):
could you make custom layout and icons optional instead of completely disabling this option?

PR #991, opened 2026-03-19
feat: Graph refactor chapter II

poneciak57:
Closes RNAA-433

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
poneciak57, line 1: just checking in
is it possible to call start and stop concurently (eg at the same time)?

packages/react-native-audio-api/common/cpp/audioapi/core/effects/delay/DelayWriter.cpp
```diff
@@ -0,0 +1,47 @@
...
+  auto delayBuffer = delayLine_->getBuffer();
+  auto delayTime = delayLine_->getDelayTimeParam()->processKRateParam(
+      framesToProcess, context->getCurrentTime());
+  const size_t readForWrite = delayLine_->readSnapshotForWrite();
+  size_t writeIndex =
+      static_cast<size_t>(static_cast<float>(readForWrite) + delayTime * context->getSampleRate()) %
+      delayBuffer->getSize();
+
+  delay_ring::bufferOperation(
+      delayBuffer, audioBuffer_, framesToProcess, writeIndex, delay_ring::BufferAction::WRITE);
```
poneciak57, line 36: why u have Action type instead of having two methods write/read?

packages/react-native-audio-api/common/cpp/audioapi/core/effects/ConvolverNode.cpp
```diff
@@ -26,7 +24,7 @@ ConvolverNode::ConvolverNode(
 void ConvolverNode::setBuffer(
     const std::shared_ptr<AudioBuffer> &buffer,
     std::vector<std::unique_ptr<Convolver>> convolvers,
-    const std::shared_ptr<ThreadPool> &threadPool,
+    const std::shared_ptr<ThreadPool<32>> &threadPool,
```
poneciak57: 32 is deffinitly too much
threads are used here for compute heavy operations so should be determined by hardware_concurrency. Otherwise they will hog cpu and steal power from each other.
Usually it does not matter that much but here they are not sleeping theyre job is to compute

packages/react-native-audio-api/common/cpp/audioapi/core/effects/ConvolverNode.cpp
```diff
@@ -91,14 +96,21 @@ float ConvolverNode::calculateNormalizationScale(const std::shared_ptr<AudioBuff
...
-      return;
-    }
+  if (buffer_ == nullptr) {
+    return;
+  }
+
+  if (framesToProcess != RENDER_QUANTUM_SIZE) {
+    printf(
+        "[AUDIOAPI WARN] convolver requires 128 buffer size for each render quantum, otherwise quality of convolution is very poor\n");
+  }
```
poneciak57, line 117: printf does not show in android logcat use if macro based on system

packages/react-native-audio-api/common/cpp/audioapi/core/effects/IIRFilterNode.cpp
poneciak57, line 1: make sure this file is compiled with -O3 -ffast-math

packages/react-native-audio-api/common/cpp/audioapi/core/effects/IIRFilterNode.h
```diff
@@ -56,16 +56,42 @@ class IIRFilterNode : public AudioNode {
...
 
+  /// Envelope threshold at which the impulse response is considered inaudible (-80 dB).
+  static constexpr double kTailEpsilon = 1e-4;
+
+  /// Hard upper bound on the computed tail length (seconds).
+  static constexpr float kMaxTailSeconds = 30.0f;
+
+  /// Keeps `log(r)` bounded away from zero when the pole sits on or outside
+  /// the unit circle.
+  static constexpr double kPoleRadiusEpsilon = 1e-5;
```
poneciak57, line 78: constexpr does not need to be marked static

packages/react-native-audio-api/common/cpp/audioapi/jsi/JsiPromise.h
```diff
@@ -108,7 +108,7 @@ class PromiseVendor {
  private:
   jsi::Runtime *runtime_;
   std::shared_ptr<react::CallInvoker> callInvoker_;
-  std::shared_ptr<ThreadPool> threadPool_;
+  std::shared_ptr<ThreadPool<96>> threadPool_;
```
poneciak57: 96?

packages/react-native-audio-api/common/cpp/audioapi/core/utils/graph/Graph.cpp
```diff
@@ -0,0 +1,174 @@
...
+    if (eventReceiver_.rcvCursor() < front->barrier) {
+      drainA();
+      if (eventReceiver_.rcvCursor() < front->barrier) {
+        // Producer is mid-send on Channel A; the event will arrive
+        // imminently. Defer this orphan to the next processEvents()
+        // tick rather than burning audio-thread cycles spinning.
+        break;
+      }
+    }
+    OrphanEnvelope consumed;
```
poneciak57, line 65: this "vector-clock-like" aproach is great, but the logic seems to be split between few places. It would be nice to maybe make some wrapper on top of the SPSC to handle that scenario

poneciak57 (commented):
Overally it looks really great. One global issue is lack of high level documentation. It would be great to write/generate some high level design doc on top of that or keep existing ones in sync with the code

But as i said overally good job 👍

packages/react-native-audio-api/common/cpp/audioapi/core/effects/IIRFilterNode.h
mdydek: but it is inside of class, 
"Non-static data member cannot be constexpr; did you intend to make it static? (fix available)clang(invalid_constexpr_member)"

packages/react-native-audio-api/common/cpp/audioapi/jsi/JsiPromise.h
mdydek: I rewrote threadpool to fatfunction instead of moveonly function and task size needs to be passed in threadpool with new changes

packages/react-native-audio-api/common/cpp/audioapi/core/effects/ConvolverNode.cpp
mdydek: i get that this number can be confusing, but its not amount of threads, but task size so threadpool can be safely used on audio thread

packages/react-native-audio-api/common/cpp/audioapi/core/effects/delay/DelayWriter.cpp
mdydek: because it shares a lot of code under the hood, but it would be hard to split it into two separate functions and put common code elsewhere, this way we can easily control what is shared and what is not

packages/react-native-audio-api/common/cpp/audioapi/jsi/JsiPromise.h
mdydek: now it should be slightly more informative

packages/react-native-audio-api/common/cpp/audioapi/core/effects/delay/host_nodes/DelayReaderHostNode.cpp
maciejmakowski2003, line 1: lets merge it with .h

packages/react-native-audio-api/common/cpp/audioapi/core/effects/delay/host_nodes/DelayWriterHostNode.cpp
maciejmakowski2003, line 1: lets merge it with .h

packages/react-native-audio-api/common/cpp/audioapi/core/effects/delay/DelayLine.h
```diff
@@ -0,0 +1,66 @@
...
+
+  /// Call at the start of DelayReader/DelayWriter `processNode` (same `currentSampleFrame` for
+  /// the whole graph pass). First touch each quantum refreshes `readSnapshotForWrite_` from
+  /// `readIndex_`.
+  void syncReadSnapshotForQuantum(size_t currentSampleFrame) {
+    if (currentSampleFrame != quantumSampleFrame_) {
+      writeIndex_ = readIndex_;
+      quantumSampleFrame_ = currentSampleFrame;
+    }
+  }
```
maciejmakowski2003, line 38: why?

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferSourceNodeHostObject.cpp
```diff
@@ -173,7 +173,11 @@ void AudioBufferSourceNodeHostObject::setOnLoopEndedCallbackId(uint64_t callback
   };
 
   audioBufferSourceNode->unregisterOnLoopEndedCallback(onLoopEndedCallbackId_);
-  audioBufferSourceNode->scheduleAudioEvent(std::move(event));
+  if (callbackId == 0) {
+    audioBufferSourceNode->scheduleGCEvent(std::move(event));
+  } else {
+    audioBufferSourceNode->scheduleAudioEvent(std::move(event));
+  }
```
maciejmakowski2003: 0 could be sent from JS as well

packages/react-native-audio-api/common/cpp/audioapi/utils/SpscChannel.hpp
```diff
@@ -299,6 +352,10 @@ class InnerChannel {
     buffer_[rcvCursor].~T(); // Call destructor
 
     rcvCursor_.store(next_index(rcvCursor), std::memory_order_release);
+    // Monotonic sequence: paired with sendSeq_ so external code can compare
+    // "events received" against a snapshot of "events sent" without having
+    // to reason about wrap-around in the masked cursors above.
+    rcvSeq_.store(rcvSeq_.load(std::memory_order_relaxed) + 1, std::memory_order_release);
```
maciejmakowski2003: why dont you use `fetch_add`?

packages/react-native-audio-api/common/cpp/audioapi/utils/SpscChannel.hpp
```diff
@@ -332,6 +403,7 @@ class InnerChannel {
     new (&buffer_[sendCursor]) T(std::forward<U>(value));
 
     sendCursor_.store(next_sendCursor, std::memory_order_release);
+    sendSeq_.store(sendSeq_.load(std::memory_order_relaxed) + 1, std::memory_order_release);
```
maciejmakowski2003: `fetch add`???

maciejmakowski2003 (changes requested):
a few more remarks

packages/react-native-audio-api/common/cpp/audioapi/core/effects/delay/DelayWriter.cpp
poneciak57: @copilot address this architecture choice discussion

packages/react-native-audio-api/common/cpp/audioapi/core/effects/ConvolverNode.cpp
poneciak57: but thread pool spawns this many threads, no?

packages/react-native-audio-api/common/cpp/audioapi/core/effects/IIRFilterNode.h
poneciak57: oh alright

packages/react-native-audio-api/common/cpp/audioapi/core/effects/delay/DelayWriter.cpp
poneciak57: hope he will not empty out my wallet for this question ....

poneciak57: eh... 
I would challenge the assumption that minimizing code duplication should be the primary goal.

Read and write are distinct operations with different semantics. The fact that they share some implementation details doesn't necessarily mean they should be exposed through a single function with an Action enum.

I'd generally prefer separate read() and write() APIs and keep any shared logic internal. A slightly larger implementation is often a reasonable trade-off for a cleaner and more explicit interface.

poneciak57:
🎉

PR #993, opened 2026-03-20
Feat/long file playback

### LINKED ISSUE #735: Memory-savvy playback of longer files
### Feature Request

Currently, the go-to way of playing an audio file seems to be loading the file into memory, creating `AudioBufferSourceNode` out of it and connecting it to the graph.

When, however, the file gets bigger (say, an hour), this approach falls short as both time to load and memory requirements become unreasonable, especially on mobile devices. 

It seems to be possible to work around this limitation by using `AudioBufferQueueSourceNode` and loading chunks of audio and keeping the queue full. Extra care, however, needs to be taken to support rewinding, delays in data reading etc.

I was curious if anything like that is planned to be implemented? Web Audio API provides [MediaElementAudioSourceNode](https://developer.mozilla.org/en-US/docs/Web/API/MediaElementAudioSourceNode) class for that purpose, and, while not exactly portable, might be a good inspiration for something like `AudioURLSourceNode`?

mdydek:
Closes #735, RNAA-285, RNAA-13

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioFileSourceNode.h
```diff
@@ -0,0 +1,66 @@
...
+
+ protected:
+  std::shared_ptr<DSPAudioBuffer> processNode(
+      const std::shared_ptr<DSPAudioBuffer> &processingBuffer,
+      int framesToProcess) override;
+
+ private:
+  std::shared_ptr<AudioFileDecoderState> decoderState_;
+  std::atomic<float> volume_;
+  bool FFmpegNeeded_;
```
michalsek: nit
```suggestion
  bool requiresFFmpeg_;
```

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -1,9 +1,45 @@
...
+export const AudioComponentContext = createContext<AudioComponentContextType>({
+  play: () => {},
+  pause: () => {},
+  volume: 1,
+  setVolume: () => {},
+  muted: false,
+  setMuted: () => {},
+  isReady: false,
+  playbackState: 'idle',
+});
```
michalsek: would be great to extract it to separate file ie. `audioTagContex.ts`

michalsek: and on second though we could go with 

```ts
createContext<AudioComponentContextType | null>(null);
```

plus a helper hook

```ts
function useAudioTagContext() {
  const context = useContext(AudioComponentContext);

  if (!context) {
    throw new Error("useAudioTagContext used within a provider tag!");
  }

  return context; // always returns AudioComponentContextType
}

File: packages/react-native-audio-api/src/development/react/Audio/Audio.tsx [hunk 3]
```diff
@@ -15,10 +51,165 @@ const Audio: React.FC<AudioProps> = (inProps) => {
...
+
+  const nodeRef = useRef<IAudioFileSourceNode | null>(null);
+  const [volumeState, setVolumeState] = useState(volume);
+  const [mutedState, setMutedState] = useState(muted);
+  const [isReady, setIsReady] = useState(false);
+  const [playbackState, setPlaybackState] =
+    useState<AudioTagPlaybackState>('idle');
+
+  const volumeStateRef = useRef(volumeState);
+  volumeStateRef.current = volumeState;
```
michalsek: should be an useEffect

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -15,10 +51,165 @@ const Audio: React.FC<AudioProps> = (inProps) => {
...
+
+  const mutedStateRef = useRef(mutedState);
+  mutedStateRef.current = mutedState;
```
michalsek: should be an useEffect

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -15,10 +51,165 @@ const Audio: React.FC<AudioProps> = (inProps) => {
...
+
+  const contextRef = useRef(context);
+  contextRef.current = context;
```
michalsek: same

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -15,10 +51,165 @@ const Audio: React.FC<AudioProps> = (inProps) => {
...
+      const prev = nodeRef.current;
+      if (prev) {
+        prev.onEnded = '0';
+      }
+      prev?.disconnect(undefined);
+      nodeRef.current = null;
+      setIsReady(false);
+      setPlaybackState('idle');
+    };
+  }, [path, context, attachNode]);
```
michalsek: this means that each time we toggle play state whe fetch and load the file again

michalsek: You can try using the vercels react best practices to cleanup a bit the component code :)

packages/react-native-audio-api/src/development/react/Audio/types.ts
```diff
@@ -32,6 +42,8 @@ interface AudioControlProps {
   playbackRate: number;
   preservesPitch: boolean;
   volume: number;
+  children?: ReactNode;
+  context: BaseAudioContext | null; // null on web, since web do not use AudioContext for audio tag
```
michalsek:
```suggestion
  context?: BaseAudioContext; // optional on web, since web do not use AudioContext for audio tag
```

michalsek (changes requested):

packages/react-native-audio-api/src/development/react/Audio/utils.ts
```diff
@@ -13,10 +24,22 @@ export function withPropsDefaults(props: AudioProps): AudioPropsBase {
...
+  const defaultContextRef = useRef<BaseAudioContext | null>(null);
+  const resolvedContext: BaseAudioContext | undefined =
+    Platform.OS === 'web'
+      ? undefined
+      : (() => {
+          if (defaultContextRef.current === null) {
+            defaultContextRef.current = new AudioContext();
+          }
+          return props.context ?? defaultContextRef.current ?? undefined;
+        })();
```
michalsek:
```suggestion
  const context = useMemo(() => {
    return props.context ?? new AudioContext();
  }, [props.context]);
```

packages/react-native-audio-api/src/development/react/Audio/AudioTagContext.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const AudioComponentContext = createContext<
+  AudioComponentContextType | undefined
+>(undefined);
+
+export function useAudioTagContext(): AudioComponentContextType {
+  const context = useContext(AudioComponentContext);
+
+  if (context === undefined) {
+    throw new Error('useAudioTagContext must be used within an <Audio> tag.');
```
michalsek:
```suggestion
    throw new Error('useAudioTagContext must be used within an <Audio> component.');
```

packages/react-native-audio-api/src/development/react/Audio/AudioFileSourceNode.ts
```diff
@@ -0,0 +1,125 @@
...
+    if (!this.node) {
+      return;
+    }
+    if (!this.didConnectToDestination) {
+      // @ts-expect-error destination.node is the underlying graph node
+      this.node.connect(baseContext.destination.node);
+      this.didConnectToDestination = true;
+    }
+    this.node.start();
+  }
```
michalsek: why not pass the context through constructor?

packages/react-native-audio-api/src/development/react/Audio/AudioFileSourceNode.ts
```diff
@@ -0,0 +1,125 @@
...
+import type { EventEmptyType } from '../../../events/types';
+import type BaseAudioContext from '../../../core/BaseAudioContext';
+import type { IAudioFileSourceNode } from '../../../interfaces';
+
+type AttachFileSourceOptions = {
+  loop: boolean;
+  onEnded: () => void;
+};
+
+export class AudioFileSourceNode {
```
michalsek: Can't this class extend from AudioNode and/or follow its structure (i.e. constructor)

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
```diff
@@ -0,0 +1,346 @@
...
+  };
+
+  const collapse = () => {
+    height.value = withTiming(TRACK_BAR_HEIGHT, {
+      duration: TRACK_BAR_ANIM_MS,
+    });
+  };
+
+  return { animatedStyle, expand, collapse };
+}
```
michalsek: would be nice to move that to separate files: `utils.ts` `constants.ts` etc.

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
```diff
@@ -0,0 +1,346 @@
...
+    });
+  };
+
+  const collapse = () => {
+    height.value = withTiming(TRACK_BAR_HEIGHT, {
+      duration: TRACK_BAR_ANIM_MS,
+    });
+  };
+
+  return { animatedStyle, expand, collapse };
```
michalsek: useMemo 🤔

packages/react-native-audio-api/src/development/react/Audio/controls/AudioControls.tsx
```diff
@@ -0,0 +1,346 @@
...
+        scheduleOnRN(onUpdate, e.x);
+      })
+      .onEnd((e) => {
+        scheduleOnRN(onEnd, e.x);
+      })
+      .onFinalize((_e, success) => {
+        if (!success) {
+          scheduleOnRN(onCancel);
+        }
+      });
```
michalsek, line 127:
```suggestion
    const panGesture = Gesture.Pan()
      .runOnJS(true)
      .minDistance(SCRUB_PAN_MIN_DISTANCE)
      .onStart((e) => {
        onStart(e.x);
      })
      .onUpdate((e) => {
        onUpdate(e.x);
      })
      .onEnd((e) => {
        onEnd(e.x);
      })
      .onFinalize((_e, success) => {
        if (!success) {
          onCancel();
        }
      });
```

packages/react-native-audio-api/src/development/react/Audio/controls/AudioControls.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+    const tapGesture = Gesture.Tap()
+      .maxDistance(14)
+      .onEnd((e, success) => {
+        if (success) {
+          scheduleOnRN(onTapSeek, e.x);
+        }
+      });
```
michalsek, line 136:
```suggestion
    const tapGesture = Gesture.Tap()
      .runOnJS(true)
      .maxDistance(14)
      .onEnd((e, success) => {
        if (success) {
          onTapSeek(e.x);
        }
      });
```

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+  const onProgressTrackLayout = (e: LayoutChangeEvent) => {
+    const w = e.nativeEvent.layout.width;
+    setProgressWidth(w);
+    progressWidthRef.current = w;
+    progressTrackRef.current?.measureInWindow((left, _y, width, _h) => {
+      progressMetricsRef.current.left = left;
+      progressMetricsRef.current.width = width;
+    });
+  };
```
michalsek: useCallback

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioFileSourceNodeHostObject.cpp
```diff
@@ -0,0 +1,132 @@
...
+  auto sourceNode = std::static_pointer_cast<AudioFileSourceNode>(node_);
+
+  auto event = [sourceNode, callbackId](BaseAudioContext &) {
+    sourceNode->setOnPositionChangedCallbackId(callbackId);
+  };
+
+  sourceNode->unregisterOnPositionChangedCallback(onPositionChangedCallbackId_);
+  sourceNode->scheduleAudioEvent(std::move(event));
+  onPositionChangedCallbackId_ = callbackId;
+}
```
poneciak57, line 115:
```suggestion
void AudioFileSourceNodeHostObject::setOnPositionChangedCallbackId(uint64_t callbackId) {
  auto sourceNode = std::static_pointer_cast<AudioFileSourceNode>(node_);

  auto event = [sourceNode, callbackId, onPositionChangedCallbackId_ = this.onPositionChangedCallbackId_](BaseAudioContext &) {
    sourceNode->unregisterOnPositionChangedCallback(onPositionChangedCallbackId_);
    sourceNode->setOnPositionChangedCallbackId(callbackId);
  };

  sourceNode->scheduleAudioEvent(std::move(event));
  onPositionChangedCallbackId_ = callbackId;
}
```

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
```diff
@@ -0,0 +1,346 @@
...
+    return Gesture.Race(panGesture, tapGesture);
+  }, [onStart, onUpdate, onEnd, onCancel, onTapSeek]);
+
+  const onPlayPausePress = () => {
+    if (playbackState === 'playing') {
+      pause();
+    } else {
+      play();
+    }
+  };
```
michalsek: useCallback for consistency

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioFileSourceNodeHostObject.cpp
```diff
@@ -0,0 +1,132 @@
...
+  auto sourceNode = std::static_pointer_cast<AudioFileSourceNode>(node_);
+
+  auto event = [sourceNode, callbackId](BaseAudioContext &) {
+    sourceNode->setOnEndedCallbackId(callbackId);
+  };
+
+  sourceNode->unregisterOnEndedCallback(onEndedCallbackId_);
+  sourceNode->scheduleAudioEvent(std::move(event));
+  onEndedCallbackId_ = callbackId;
+}
```
poneciak57, line 127:
```suggestion
void AudioFileSourceNodeHostObject::setOnEndedCallbackId(uint64_t callbackId) {
  auto sourceNode = std::static_pointer_cast<AudioFileSourceNode>(node_);

  auto event = [sourceNode, callbackId, onEndedCallbackId_ = this.onEndedCallbackId_](BaseAudioContext &) {
    sourceNode->unregisterOnEndedCallback(onEndedCallbackId_);
    sourceNode->setOnEndedCallbackId(callbackId);
  };

  sourceNode->scheduleAudioEvent(std::move(event));
  onEndedCallbackId_ = callbackId;
}
```

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
```diff
@@ -0,0 +1,346 @@
...
+
+  const onProgressTrackLayout = (e: LayoutChangeEvent) => {
+    const w = e.nativeEvent.layout.width;
+    setProgressWidth(w);
+    progressWidthRef.current = w;
+    progressTrackRef.current?.measureInWindow((left, _y, width, _h) => {
+      progressMetricsRef.current.left = left;
+      progressMetricsRef.current.width = width;
+    });
```
michalsek: instead of a tracking width directly, we could convert that to percentage and base all of the calculations above on that, this will make us free on recalculating everything on device rotation change or window size change (in case of mac catalyst f.e.)

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
```diff
@@ -0,0 +1,346 @@
...
+      } else {
+        wasPlayingBeforeScrubRef.current = false;
+      }
+      setTimeout(() => {
+        seekTo(x);
+        setTimeout(() => {
+          onEnd(x, false);
+        }, NATIVE_DELAY_MS);
+      }, NATIVE_DELAY_MS);
+    },
```
michalsek: can/can't the seekToTime (and JSI call underneath) be synchronous or a Promise?

michalsek: With dangling timeouts like this, we can hit the state that if component unmounts during the timeout, it will try to update state that no longer exists

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/BaseAudioContextHostObject.cpp
```diff
@@ -250,6 +253,39 @@ JSI_HOST_FUNCTION_IMPL(BaseAudioContextHostObject, createBufferSource) {
...
+        auto arrayBuffer = obj.getArrayBuffer(runtime);
+        auto *data = arrayBuffer.data(runtime);
+        auto size = arrayBuffer.size(runtime);
+        auto format = AudioDecoder::detectAudioFormat(data, size);
+#if RN_AUDIO_API_FFMPEG_DISABLED
+        if (format == AudioFormat::MP4 || format == AudioFormat::M4A ||
+            format == AudioFormat::AAC) {
+          return jsi::Value::undefined();
+        }
+#endif // RN_AUDIO_API_FFMPEG_DISABLED
```
poneciak57: it seems like a duplicate of this, why are we checking this twice? https://github.com/software-mansion/react-native-audio-api/blob/2440c8f72492e4493314bb9fb1cd1aa76c15e7bf/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioFileSourceNode.cpp#L34-L54

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
```diff
@@ -0,0 +1,346 @@
...
+      const d = durationRef.current;
+      const w = progressMetricsRef.current.width || progressWidthRef.current;
+      const t = timeFromLocationX(x, w, d);
+      seekToTime(t);
+    },
+    [seekToTime]
+  );
+
+  const onEnd = useCallback(
+    (x: number, flag: boolean = true) => {
```
michalsek: flag?

michalsek: could use a bit more self-explanatory name

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
```diff
@@ -0,0 +1,346 @@
...
+  const [progressWidth, setProgressWidth] = useState(0);
+  const [scrubTime, setScrubTime] = useState<number | null>(null);
+
+  const progressTrackAnim = useExpandableTrackHeight();
+
+  const progressTrackRef = useRef<View>(null);
+  const progressMetricsRef = useRef({ left: 0, width: 0 });
+  const progressWidthRef = useRef(0);
+  const durationRef = useRef(duration);
+  const wasPlayingBeforeScrubRef = useRef(false);
```
michalsek: since we're depending on reanimated+GH, these could safely be a shared values

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
```diff
@@ -0,0 +1,346 @@
...
+  locationX: number,
+  trackWidth: number,
+  durationSeconds: number
+): number {
+  if (trackWidth <= 0 || durationSeconds <= 0) {
+    return 0;
+  }
+  const pct = Math.max(0, Math.min(1, locationX / trackWidth));
+  return pct * durationSeconds;
+}
```
michalsek: utils.ts :)

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
```diff
@@ -0,0 +1,346 @@
...
+import { scheduleOnRN } from 'react-native-worklets';
+import Animated, {
+  useAnimatedStyle,
+  useSharedValue,
+  withTiming,
+} from 'react-native-reanimated';
+import { useAudioTagContext } from './AudioTagContext';
+
+function formatTime(seconds: number): string {
+  if (!Number.isFinite(seconds) || seconds < 0) return '0:00';
```
michalsek:
```suggestion
  if (!Number.isFinite(seconds) || seconds < 0) {
    return '0:00';
  }
```

poneciak57 (changes requested):

packages/react-native-audio-api/src/development/react/Audio/Audio.web.tsx
```diff
@@ -1,9 +1,175 @@
...
+    const el = audioRef.current;
+    if (!el) return;
+    const onLoadedMetadata = () => {
+      setDuration(el.duration);
+      setCurrentTime(el.currentTime);
+    };
+    const onTimeUpdate = () => setCurrentTime(el.currentTime);
+    el.addEventListener('loadedmetadata', onLoadedMetadata);
+    el.addEventListener('timeupdate', onTimeUpdate);
+    if (!isNaN(el.duration)) setDuration(el.duration);
```
michalsek:
```suggestion
    if (!isNaN(el.duration)) {
      setDuration(el.duration);
    }
```

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioFileSourceNodeHostObject.cpp
maciejmakowski2003: why? `unregisterOnPositionChangedCallback(onPositionChangedCallbackId_)`  calls `audioEventHandlerRegistry_->unregisterHandler(AudioEvent::ENDED, callbackId)` that calls invokeAsync on JS thread

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioFileSourceNodeHostObject.cpp
maciejmakowski2003: same as above

packages/react-native-audio-api/common/cpp/audioapi/libs/ffmpeg/FFmpegDecoding.cpp
```diff
@@ -8,325 +8,511 @@
...
-  return decodeAudioFrames(fmt_ctx.get(), codec_ctx.get(), audio_stream_index, sample_rate);
+  size_t frames = interleaved.size() / static_cast<size_t>(channels);
+  auto buf = std::make_shared<AudioBuffer>(frames, channels, static_cast<float>(sample_rate));
+  for (int c = 0; c < channels; ++c) {
+    auto span = buf->getChannel(c)->span();
+    for (size_t i = 0; i < frames; ++i) {
+      span[i] = interleaved[i * static_cast<size_t>(channels) + static_cast<size_t>(c)];
+    }
+  }
+  return buf;
```
maciejmakowski2003, line 437: you can use AudioBuffer method `deinterleaveFrom`

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioFileSourceNodeHostObject.h
```diff
@@ -0,0 +1,43 @@
...
+  ~AudioFileSourceNodeHostObject() override;
+
+  JSI_PROPERTY_GETTER_DECL(volume);
+  JSI_PROPERTY_SETTER_DECL(volume);
+  JSI_PROPERTY_GETTER_DECL(loop);
+  JSI_PROPERTY_SETTER_DECL(loop);
+  JSI_PROPERTY_GETTER_DECL(currentTime);
+  JSI_PROPERTY_GETTER_DECL(duration);
+  JSI_PROPERTY_SETTER_DECL(onPositionChanged);
+  JSI_PROPERTY_SETTER_DECL(onEnded);
```
maciejmakowski2003:
```suggestion
  JSI_PROPERTY_GETTER_DECL(volume);
  JSI_PROPERTY_GETTER_DECL(loop);
  JSI_PROPERTY_GETTER_DECL(currentTime);
  JSI_PROPERTY_GETTER_DECL(duration);
  
  JSI_PROPERTY_SETTER_DECL(volume);
  JSI_PROPERTY_SETTER_DECL(loop);
  JSI_PROPERTY_SETTER_DECL(onPositionChanged);
  JSI_PROPERTY_SETTER_DECL(onEnded);
```

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioFileSourceNodeHostObject.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+JSI_PROPERTY_GETTER_IMPL(AudioFileSourceNodeHostObject, currentTime) {
+  auto node = std::static_pointer_cast<AudioFileSourceNode>(node_);
+  return {node->getCurrentTime()};
+}
+
+JSI_PROPERTY_GETTER_IMPL(AudioFileSourceNodeHostObject, duration) {
+  auto node = std::static_pointer_cast<AudioFileSourceNode>(node_);
+  return {node->getDuration()};
```
maciejmakowski2003: you can store duration in shadow state, as well as volume and loop.

packages/react-native-audio-api/common/cpp/audioapi/core/BaseAudioContext.cpp
```diff
@@ -177,6 +178,15 @@ std::shared_ptr<AudioBufferSourceNode> BaseAudioContext::createBufferSource(
   return bufferSource;
 }
 
+#if !RN_AUDIO_API_TEST
+std::shared_ptr<AudioFileSourceNode> BaseAudioContext::createFileSource(
+    const AudioFileSourceOptions &options) {
+  auto fileSource = std::make_shared<AudioFileSourceNode>(shared_from_this(), options);
+  //  graphManager_->addProcessingNode(fileSource);
```
maciejmakowski2003: ??

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioFileSourceNode.cpp
```diff
@@ -0,0 +1,349 @@
...
+      state->sampleRate = static_cast<float>(maDecoder_->outputSampleRate);
+      ma_uint64 length = 0;
+      if (ma_decoder_get_length_in_pcm_frames(maDecoder_.get(), &length) == MA_SUCCESS) {
+        duration_ = static_cast<double>(length) / state->sampleRate;
+      }
+    } else {
+      ma_decoder_uninit(maDecoder_.get());
+      maDecoder_.reset();
+    }
+  }
```
maciejmakowski2003: is there any way to hide ma decoder initialization impl details inside MiniAudio Decoder class?

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioFileSourceNode.cpp
```diff
@@ -0,0 +1,349 @@
...
+    return ffmpegDecoder_.readPcmFrames(buf, frameCount);
+#else
+    return 0;
+#endif
+  }
+  if (maDecoder_ == nullptr) {
+    return 0;
+  }
+  ma_uint64 framesRead = 0;
+  ma_decoder_read_pcm_frames(maDecoder_.get(), buf, frameCount, &framesRead);
```
maciejmakowski2003: lets hide mini audio impl details

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioFileSourceNode.cpp
```diff
@@ -0,0 +1,349 @@
...
+    return;
+  }
+  auto numOutputChannels = static_cast<int>(processingBuffer->getNumberOfChannels());
+  for (size_t i = 0; i < frameCount; i++) {
+    for (int ch = 0; ch < numOutputChannels; ch++) {
+      int srcCh = ch < state.channels ? ch : state.channels - 1;
+      processingBuffer->getChannel(ch)->span()[destSampleOffset + i] =
+          vol * state.interleavedBuffer[i * state.channels + srcCh];
+    }
+  }
```
maciejmakowski2003: use deinterleaveFrom AudioBuffer method

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioFileSourceNode.cpp
```diff
@@ -0,0 +1,349 @@
...
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+namespace audioapi {
+
+AudioFileSourceNode::AudioFileSourceNode(
+    const std::shared_ptr<BaseAudioContext> &context,
+    const AudioFileSourceOptions &options)
+    : AudioNode(context, options),
```
maciejmakowski2003: I should inherit from AudioScheduledSourceNode

maciejmakowski2003: it is highly misleading that AudioNode that is not a child of ASSN is a source

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioFileSourceNode.h
```diff
@@ -0,0 +1,147 @@
...
+
+  std::shared_ptr<AudioFileDecoderState> decoderState_;
+  std::unique_ptr<ma_decoder> maDecoder_;
+  std::atomic<float> volume_;
+  bool requiresFFmpeg_;
+#if !RN_AUDIO_API_FFMPEG_DISABLED
+  ffmpegdecoder::FFmpegDecoder ffmpegDecoder_;
+  ffmpegdecoder::FFmpegDecoderConfig cfg;
+#endif // RN_AUDIO_API_FFMPEG_DISABLED
+  std::atomic<bool> filePaused_{false};
```
maciejmakowski2003: maybe it will be better to store state of node like in ASSN. why pause is not send via events queue? I do not think that atomic usage here is necessary. is there any reason behind it?

maciejmakowski2003 (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioFileSourceNodeHostObject.cpp
poneciak57: hmmmm

poneciak57: but what if we call a callback in between unregistration and setOnPositionChangeCallback
in this case we try to invoke non existent callback
are we sure this will work fine?

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioFileSourceNodeHostObject.cpp
poneciak57: same as above

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioFileSourceNodeHostObject.cpp
maciejmakowski2003: map with callback is accessed only inside invokeAsync block, so indeed it is thread safe

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioFileSourceNode.cpp
mdydek: I based my implementation to mimic in a way https://webaudio.github.io/web-audio-api/#MediaElementAudioSourceNode, this node is not an `AudioScheduledSourceNode`. Audio tag does not expose any methods to "schedule" playback, it would not benefit from inheritance nor it should be it

mdydek: in a future `MediaElementAudioSourceNode` would be wrapper on this node preferably

maciejmakowski2003: so how would you like to treat AudioFileSourceNode inside a graph?

mdydek: deconstruction-wise?

maciejmakowski2003: exactly

apps/fabric-example/metro.config.js
```diff
@@ -13,6 +13,27 @@ const appsRoot = path.resolve(monorepoRoot, 'apps');
 const config = {
   projectRoot: __dirname,
   watchFolders: [monorepoRoot, appsRoot],
+  server: {
+    rewriteRequestUrl: (url) => {
```
michalsek, line 19: lets add some comment on why this is here

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -15,10 +33,236 @@ const Audio: React.FC<AudioProps> = (inProps) => {
...
+      loop: loopRef.current,
+      volume: effectiveVolumeRef.current,
+    });
+    if (!node) {
+      throw new NotSupportedError('This file format requires FFmpeg build');
+    }
+    const { duration: d } = handle.attach(node, {
+      loop: loopRef.current,
+      onEnded: () => handlePlaybackEnded(),
+    });
```
michalsek: can I have some of them? 
Empty lines?

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -15,10 +33,236 @@ const Audio: React.FC<AudioProps> = (inProps) => {
...
+      isReady,
+      playbackState,
+      currentTime,
+      duration,
+    ]
+  );
+
+  if (context == null) {
+    return null;
+  }
```
michalsek: warn or throw? 🤔

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -15,10 +33,236 @@ const Audio: React.FC<AudioProps> = (inProps) => {
...
+      if (path.startsWith('http')) {
+        const response = await fetch(path, {
+          headers: (source as AudioURISource).headers,
+        });
+        const arrayBuffer = await response.arrayBuffer();
+        attachNode(arrayBuffer);
+      } else {
+        attachNode(path);
+      }
+    };
```
michalsek, line 214: can't we handle that through the context decode method? Handles headers etc. already

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -15,10 +33,236 @@ const Audio: React.FC<AudioProps> = (inProps) => {
...
+  }
+
+  return (
+    <AudioComponentContext.Provider value={ctxValue}>
+      <View style={{ alignSelf: 'stretch', width: '100%' }}>
+        {controls && <AudioControls />}
+        {children}
+      </View>
+    </AudioComponentContext.Provider>
+  );
```
michalsek, line 317: kind of nit, kind of something for later: this component could be split into multiple single-purpose hooks to heavily improve code readability and clarity

packages/react-native-audio-api/src/development/react/Audio/Audio.web.tsx
```diff
@@ -1,9 +1,187 @@
...
+
+  useEffect(() => {
+    const el = audioRef.current;
+    if (el) {
+      el.muted = mutedState;
+    }
+  }, [mutedState]);
+
+  const play = useCallback(() => {
+    audioRef.current?.play()?.catch(() => {});
```
michalsek: empty catch? 🤔 
i think it would be better to allow it throwing in case of issues

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
```diff
@@ -0,0 +1,355 @@
...
+      progressMetricsRef.current.width = width;
+    });
+  };
+
+  if (!isReady) {
+    return (
+      <View style={styles.container}>
+        <View style={styles.loadingRow}>
+          <ActivityIndicator color="#333" size="small" />
+          <Text style={[styles.loadingText, { marginLeft: 8 }]}>Loading…</Text>
```
michalsek: without text or customizable :P

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -15,10 +33,236 @@ const Audio: React.FC<AudioProps> = (inProps) => {
...
+    }
+    // number
+    if (typeof source === 'number') {
+      return Image.resolveAssetSource(source).uri;
+    }
+    // AudioURISource
+    return source.uri ?? '';
+  }, [source]);
+
+  const fileSourceHandleRef = useRef(new AudioFileSourceNode());
```
michalsek:
```suggestion
  const fileSourceHandle = useMemo(() => new AudioFileSourceNode(), []);
```

stable handle between re-renders

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -15,10 +33,236 @@ const Audio: React.FC<AudioProps> = (inProps) => {
...
+  const loadedSourceRef = useRef<ArrayBuffer | string | null>(null);
```
michalsek: could be a state

michalsek: which will be less painfull during development, with ref, each code save means you look already loaded buffer

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -15,10 +33,236 @@ const Audio: React.FC<AudioProps> = (inProps) => {
...
+  const [mutedState, setMutedState] = useState(muted);
+  const [isReady, setIsReady] = useState(false);
+  const [playbackState, setPlaybackState] =
+    useState<AudioTagPlaybackState>('idle');
+  const [currentTime, setCurrentTime] = useState(0);
+  const [duration, setDuration] = useState(0);
+
+  useEffect(() => {
+    setVolumeState(volume);
+  }, [volume]);
```
michalsek: it would be good to decide on staying with just one api approach:
1. control props `<Audio volume={5} />`
2. context control with internal state

Currently it is in weird middle state: 
you can pass `volume` but there is no `onVolumeChange` callback to use your own state from parent component, underneath there is context and useAudio that allows this.

At the current state, it will probably be better to go with just the context and removing the props i've added to the boilerplate

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
```diff
@@ -0,0 +1,355 @@
...
+      </View>
+    );
+  }
+
+  const displayTime = scrubTime ?? currentTime;
+  const progress = duration > 0 ? displayTime / duration : 0;
+
+  return (
+    <View style={styles.container}>
+      <GestureHandlerRootView style={styles.topRow}>
```
michalsek: we shouldn't add root view inside the component

michalsek (changes requested):

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
mdydek: right now it works correctly since seeking is offloaded to different thread behind the hood and it does not stall the audio thread and we can get rid of timeouts

packages/react-native-audio-api/src/development/react/Audio/AudioControls.tsx
mdydek: old version 😅 , now it works without this flag

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
mdydek: but those methods do not expose `ArrayBuffer` which is the intermediate step in decoding, I do not need decoded `AudioBuffer` here

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
mdydek: I don't think it was supposed to exist here

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
mdydek: here you go
<br/>
<br/>
<br/>
  ^
   |
there

michalsek (approved):

maciejmakowski2003 (approved):

PR #995, opened 2026-03-23
refactor: audio param automation events

closetcaiman:
Closes RNAA-440

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- refactored `AudioParamEventQueue` to support between-element insertion (to comply with AP spec)
  - added a base class (`AutomationQueueBase`) with `BoundedPriorityQueue` (`std::multiset` based) and `AutomationRenderQueue` and `AutomationControlQueue` subclasses both `AutomationTime` sorted
  - added `resolveEventValues` to connect event `start/end` `time/value` between elements
- added a `AutomationControlQueue` to JSI for handling `curveExclusion` error
  - added purging on push to the queue (on each successful `AutomationMethod`)
  - added JS-side error throwing (+ new types)
- refactored event creation
  - split `AutomationEvent` into base (`AutomationEvent`) class and `RenderAutomationEvent` class
    - added a helper constuctor in base class to help with metadata-only event processing (in `AutomationControlQueue`)
  - added a peer factory class `RenderAutomationEventFactory` 
- addded new unit tests to check new spec compliance
- fixed interface type on `IBiquadFilterNode` to use RNAA types

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/core/types/AutomationEventType.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <string_view>
+namespace audioapi {
+
+enum class AutomationEventType {
```
mdydek:
```suggestion
enum class AutomationEventType : uint8_t {
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamEvent.h
mdydek, line 1: if the file is not using templates stick to classic .h and .cpp

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationControlQueue.cpp
```diff
@@ -0,0 +1,80 @@
+#include "audioapi/core/utils/automation/AutomationControlQueue.h"
+#include <cstddef>
+#include <format>
+#include <string>
+#include "audioapi/core/types/AutomationEventType.h"
+#include "audioapi/core/utils/automation/AutomationEvent.hpp"
+#include "audioapi/utils/Result.hpp"
```
mdydek:
```suggestion
#include <audioapi/core/utils/automation/AutomationControlQueue.h>
#include <cstddef>
#include <format>
#include <string>
#include <audioapi/core/types/AutomationEventType.h>
#include <audioapi/core/utils/automation/AutomationEvent.hpp>
#include <audioapi/utils/Result.hpp>
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationControlQueue.cpp
```diff
@@ -0,0 +1,80 @@
...
+  // Check for an event with an exact automationTime match
+  auto lo = eventQueue_.lowerBound(automationTime);
+  if (lo != it) {
+    return &(*lo);
+  }
+
+  return nullptr;
+}
+
+const AutomationEvent *AutomationControlQueue::findEventInInterval(
```
mdydek: do you see any use case for this function, apart for finding conflicts? if not I think we can safely return bool value from it and if logic will be easier with bool as returning value

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationControlQueue.cpp
```diff
@@ -0,0 +1,80 @@
...
+    double startTime,
+    double endTime) {
+  // Non-ramp events have automationTime == startTime, so lowerBound/lowerBound brackets them.
+  // Ramp events have startTime == 0 (unresolved on the control thread), so getStartTime() >= startTime
+  // filters them out.
+  for (auto it = eventQueue_.lowerBound(startTime), hi = eventQueue_.lowerBound(endTime); it != hi;
+       ++it) {
+    if (it->getStartTime() >= startTime) {
+      return &(*it);
```
mdydek: I don't like this structure `&(*it)`, and in all of those function there are many awkwards casts `&` and `*` just to return ptr to it or work around deleted copy operators in AutomationEvent, this class only has 3 fields and if you use floats instead of doubles it will be less expensive to copy.

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationEvent.hpp
```diff
@@ -0,0 +1,82 @@
...
+  /// @brief Construct from a single automationTime value, setting startTime or endTime based on type.
+  /// Ramp events (LINEAR_RAMP, EXPONENTIAL_RAMP) store automationTime as endTime.
+  /// All other types store it as startTime.
+  explicit AutomationEvent(AutomationEventType type, double automationTime)
+      : type_(type),
+        startTime_(isRamp(type) ? 0.0 : automationTime),
+        endTime_(isRamp(type) ? automationTime : 0.0) {}
+
+  AutomationEvent(const AutomationEvent &) = delete;
+  AutomationEvent &operator=(const AutomationEvent &) = delete;
```
mdydek: why we cant copy this class, it does not store any ownership or something like that and code that uses this class will be much easier

mdydek: btw all fields are trivially copyable, I think all 5 functions generated by the compilator will be correct

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamEvent.h
mdydek: or if there are no other functions than those already present I think only .h will be ok

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationQueueBase.hpp
```diff
@@ -0,0 +1,50 @@
+#pragma once
+
+#include <utility>
+#include "audioapi/core/utils/Constants.h"
+#include "audioapi/utils/BoundedPriorityQueue.hpp"
```
mdydek:
```suggestion
#include <audioapi/core/utils/Constants.h>
#include <audioapi/utils/BoundedPriorityQueue.hpp>
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationQueueBase.hpp
```diff
@@ -0,0 +1,50 @@
...
+  { event.getAutomationTime() } -> std::convertible_to<double>;
+};
+template <AutomationEventConcept TEvent>
+class AutomationQueueBase {
+ public:
+  AutomationQueueBase() = default;
+  AutomationQueueBase(const AutomationQueueBase &) = delete;
+  AutomationQueueBase &operator=(const AutomationQueueBase &) = delete;
+  AutomationQueueBase(AutomationQueueBase &&) noexcept = delete;
+  AutomationQueueBase &operator=(AutomationQueueBase &&) noexcept = delete;
```
mdydek:
```suggestion
#include <audioapi/utils/Macros.h>
...
DELETE_COPY_AND_MOVE(AutomationQueueBase);
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationQueueBase.hpp
```diff
@@ -0,0 +1,50 @@
...
+#include "audioapi/core/utils/Constants.h"
+#include "audioapi/utils/BoundedPriorityQueue.hpp"
+
+namespace audioapi {
+
+template <typename TEvent>
+concept AutomationEventConcept = requires(TEvent event) {
+  { event.getAutomationTime() } -> std::convertible_to<double>;
+};
+template <AutomationEventConcept TEvent>
```
mdydek:
```suggestion
concept AutomationEvent = requires(TEvent event) {
  { event.getAutomationTime() } -> std::convertible_to<double>;
};
template <AutomationEvent TEvent>
```
I like to hide usage of concepts even more 👍

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationRenderQueue.cpp
```diff
@@ -0,0 +1,175 @@
...
+      currentEvent_->getStartTime(),
+      currentEvent_->getEndTime(),
+      currentEvent_->getStartValue(),
+      currentEvent_->getEndValue(),
+      time);
+}
+
+bool AutomationRenderQueue::push(RenderAutomationEvent &&event) {
+  resolveEventValues(event);
+  return eventQueue_.push(std::move(event));
```
mdydek:
```suggestion
  return AutomationQueueBase::push(std::move(event));
```

mdydek: this way we rely only on one place where push is implemented

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioParamHostObject.cpp
```diff
@@ -1,10 +1,13 @@
 #include <audioapi/HostObjects/AudioParamHostObject.h>
 
 #include <audioapi/core/AudioParam.h>
+#include <audioapi/core/utils/automation/AutomationEvent.hpp>
 #include <audioapi/utils/AudioArray.hpp>
 
 #include <memory>
+#include <string>
 #include <utility>
+#include "audioapi/jsi/JsiHostObject.h"
```
mdydek:
```suggestion
#include <audioapi/jsi/JsiHostObject.h>
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationControlQueue.h
```diff
@@ -0,0 +1,40 @@
...
+#include <audioapi/utils/Result.hpp>
+#include <string>
+
+namespace audioapi {
+
+/// @brief A queue for managing audio parameter change events on the JS/control thread.
+/// @note The invariant of the queue is that its internal buffer always contains non-overlapping events.
+class AutomationControlQueue : public AutomationQueueBase<AutomationEvent> {
+ public:
+  explicit AutomationControlQueue() = default;
```
mdydek: do we need this constructor?

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamQueueBase.hpp
```diff
@@ -0,0 +1,50 @@
...
+    eventQueue_.erase(eventQueue_.lowerBound(cancelTime), eventQueue_.end());
+  }
+
+  virtual bool push(TEvent &&event) {
+    return eventQueue_.push(std::move(event));
+  }
+
+ protected:
+  struct EventComparator {
+    using is_transparent = void;
```
mdydek, line 33: ?

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamQueueBase.hpp
```diff
@@ -0,0 +1,50 @@
...
+    bool operator()(const TEvent &a, const TEvent &b) const {
+      return a.getAutomationTime() < b.getAutomationTime();
+    }
+    bool operator()(const TEvent &a, double time) const {
+      return a.getAutomationTime() < time;
+    }
+    bool operator()(double time, const TEvent &b) const {
+      return time < b.getAutomationTime();
+    }
```
mdydek, line 42:
```suggestion
    bool operator()(const TEvent &a, double time) const {
      return a.getAutomationTime() < time;
    }
    bool operator()(const TEvent &a, const TEvent &b) const {
      return operator()(a, b.getAutomationTime());
    }
    bool operator()(double time, const TEvent &b) const {
      return operator()(b, time);
    }
```
its a good pattern to reuse implementation

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationRenderEventFactory.hpp
```diff
@@ -0,0 +1,146 @@
+#pragma once
+
+#include <memory>
+#include <utility>
+#include "audioapi/core/utils/automation/RenderAutomationEvent.hpp"
+#include "audioapi/dsp/AudioUtils.hpp"
+#include "audioapi/utils/AudioArray.hpp"
```
mdydek:
```suggestion
#include <audioapi/core/utils/automation/RenderAutomationEvent.hpp>
#include <audioapi/dsp/AudioUtils.hpp>
#include <audioapi/utils/AudioArray.hpp>
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationRenderQueue.cpp
```diff
@@ -0,0 +1,175 @@
+#include <audioapi/core/utils/automation/AutomationRenderEventFactory.hpp>
+#include <audioapi/core/utils/automation/AutomationRenderQueue.h>
+#include <cstddef>
+#include <optional>
+#include <utility>
+#include "audioapi/core/types/AutomationEventType.h"
```
mdydek:
```suggestion
#include <audioapi/core/types/AutomationEventType.h>
```

packages/react-native-audio-api/src/core/AudioParam.ts
mdydek, line 1: I think that this is also a good place to refactor this file, because there is a lot of code repetition existing here

packages/react-native-audio-api/common/cpp/test/src/core/AudioParamTest.cpp
mdydek, line 1: 🐐

packages/react-native-audio-api/common/cpp/audioapi/utils/BoundedPriorityQueue.hpp
```diff
@@ -0,0 +1,155 @@
...
+  SetType set_{Compare{}, &pool_};
+
+ public:
+  explicit BoundedPriorityQueue() = default;
+  ~BoundedPriorityQueue() = default;
+
+  BoundedPriorityQueue(const BoundedPriorityQueue &) = delete;
+  BoundedPriorityQueue &operator=(const BoundedPriorityQueue &) = delete;
+  BoundedPriorityQueue(BoundedPriorityQueue &&) noexcept = delete;
+  BoundedPriorityQueue &operator=(BoundedPriorityQueue &&) noexcept = delete;
```
mdydek: similar to prev comment use `DELETE_COPY_AND_MOVE` macro

packages/react-native-audio-api/common/cpp/audioapi/utils/BoundedPriorityQueue.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+  /// @brief Insert a value in sorted order. Amortized O(1) when inserting the largest element
+  /// (common case: events scheduled in chronological order), O(log n) otherwise.
+  /// @return True if inserted, false if full.
+  template <typename U>
+  bool push(U &&value) {
+    if (isFull()) {
+      [[unlikely]] return false;
+    }
```
mdydek:
```suggestion
    if (isFull()) [[unlikely]] {
       return false;
    }
```

packages/react-native-audio-api/common/cpp/audioapi/utils/BoundedPriorityQueue.hpp
```diff
@@ -0,0 +1,155 @@
...
+    set_.insert(set_.end(), std::forward<U>(value));
+    return true;
+  }
+
+  /// @brief Remove and return the smallest element (front).
+  /// @return True if successful, false if empty.
+  bool pop(T &out) {
+    if (isEmpty()) {
+      [[unlikely]] return false;
+    }
```
mdydek:
```suggestion
    if (isEmpty()) [[unlikely]] {
       return false;
    }
```

packages/react-native-audio-api/common/cpp/audioapi/utils/BoundedPriorityQueue.hpp
```diff
@@ -0,0 +1,155 @@
...
+    out = std::move(node.value());
+    return true;
+  }
+
+  /// @brief Remove the smallest element (front) without retrieving it.
+  /// @return True if successful, false if empty.
+  bool pop() {
+    if (isEmpty()) {
+      [[unlikely]] return false;
+    }
```
mdydek:
```suggestion
    if (isEmpty()) [[unlikely]] {
       return false;
    }
```

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioParamHostObject.h
```diff
@@ -5,6 +5,9 @@
 #include <jsi/jsi.h>
 #include <cstddef>
 #include <memory>
+#include <string>
+#include "audioapi/core/utils/automation/AutomationControlQueue.h"
+#include "audioapi/utils/Result.hpp"
```
mdydek:
```suggestion
#include <audioapi/core/utils/automation/AutomationControlQueue.h>
#include <audioapi/utils/Result.hpp>
```

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioParamHostObject.h
```diff
@@ -30,12 +33,19 @@ class AudioParamHostObject : public JsiHostObject {
   JSI_HOST_FUNCTION_DECL(cancelScheduledValues);
   JSI_HOST_FUNCTION_DECL(cancelAndHoldAtTime);
 
+  JSI_HOST_FUNCTION_DECL(checkCurveExclusion);
```
mdydek, line 34:
```suggestion
JSI_HOST_FUNCTION_DECL(cancelAndHoldAtTime);
JSI_HOST_FUNCTION_DECL(checkCurveExclusion);
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/RenderAutomationEvent.hpp
```diff
@@ -1,89 +1,75 @@
 #pragma once
 
-#include <audioapi/core/types/ParamChangeEventType.h>
+#include <audioapi/core/utils/automation/AutomationEvent.hpp>
 
 #include <functional>
 #include <utility>
+#include "audioapi/core/types/AutomationEventType.h"
```
mdydek:
```suggestion
#include <audioapi/core/types/AutomationEventType.h>
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/Constants.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cmath>
+#include <cstddef>
```
mdydek: i think not needed

mdydek (commented):
I would also rename some of the files and classes, maybe `ParamControlQueue`, `ParamAudioQueue` and so on

mdydek (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationControlQueue.cpp
closetcaiman: If we change it to return `bool` we won't have information about the conflicting event, i.e. `type`, `automationTime`, thus the error message won't be as informative as it is now.

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationQueueBase.hpp
closetcaiman: Sure, didn't see this one, thanks!

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationQueueBase.hpp
closetcaiman: Doesn't it introduce name collision?

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationRenderQueue.cpp
closetcaiman: Nice catch!

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamQueueBase.hpp
closetcaiman: Transparent comparator is needed to perform comparison between different types - like searching for an event at given `automationTime`. Otherwise an event would need to be passed to make comparisons.

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamQueueBase.hpp
closetcaiman:
```cpp
bool operator()(double time, const TEvent &b) const {
    return operator()(b, time);
}
```
is not equal to

```cpp
bool operator()(double time, const TEvent &b) const {
    return time < b.getAutomationTime();
}
```

While I agree that we could reuse this, it won't suit this case. I would prefer to stick to this as it is decently verbose.

packages/react-native-audio-api/common/cpp/audioapi/core/utils/automation/AutomationControlQueue.cpp
closetcaiman: Is it really? While I agree with copying, the downside of using a copy is no easy `nullptr` - we probably would need to use `std::optional`.

mdydek: hmm maybe `Result.hpp` file can be useful

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/RenderParamEvent.h
mdydek, line 1: Can be safely .h file and all implementation can be here since it is very basic class

mdydek (commented):
Files are still not renamed `ParamControlQueue` ...

mdydek (commented):
can you also update docs section about new errors?

packages/react-native-audio-api/common/cpp/audioapi/utils/BoundedPriorityQueue.hpp
```diff
@@ -0,0 +1,152 @@
+#pragma once
+
+#include <array>
+#include <cstddef>
+#include <functional>
+#include <memory_resource>
+#include <set>
+#include <utility>
+#include "audioapi/utils/Macros.h"
```
mdydek: 👀

packages/react-native-audio-api/common/cpp/test/src/core/AudioParamTest.cpp
```diff
@@ -1,9 +1,9 @@
...
-#include <audioapi/core/OfflineAudioContext.h>
-#include <audioapi/core/utils/worklets/SafeIncludes.h>
-#include <gtest/gtest.h>
-#include <test/src/MockAudioEventHandlerRegistry.h>
+#include "audioapi/core/AudioParam.h"
 #include <memory>
+#include "audioapi/core/OfflineAudioContext.h"
+#include "audioapi/core/utils/worklets/SafeIncludes.h"
+#include "gtest/gtest.h"
+#include "test/src/MockAudioEventHandlerRegistry.h"
```
mdydek: 👀

mdydek (changes requested):

mdydek (approved):

PR #1009, opened 2026-04-07
fix: CI integration tests

closetcaiman:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- added `js` integration tests to CI `Tests` workflow
- fixed tests for `StreamerNode`
- fixed `eslint` config errors
  - added config for tests

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,16 +5,27 @@ on:
   workflow_dispatch:
 
 jobs:
-  tests:
+  cpp-tests:
     runs-on: ubuntu-latest
     steps:
-
       - name: Checkout
         uses: actions/checkout@v4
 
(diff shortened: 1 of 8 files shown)


.github/workflows/tests.yml
mdydek, line 1: you are using the same steps: checkout and setup for each job, so it can be easily wrapped to run in one container instead of two

mdydek (changes requested):
I think also that this pr is a good place to slightly refactor where invoking tests is taking place, we should not run them from monorepo root, but from `packages/react-native-audio-api` path

mdydek (approved):

PR #1005, opened 2026-03-31
Invalid engine states on recording or playback start

michalsek:
Closes #1004 
Closes RNAA-407
Closes RNAA-382

## ⚠️ Breaking changes ⚠️

- None.

## Introduced changes

- Fixes issue when audio session was deactivated without suspending the context leading to broken state.
- Both iOS player and recorder will try to activate or re-activate the session when starting.
- Each time AVAudioEngine enters "iddle" state it will be teared down, to reconstruct later
- Fixes missing `isPaused` in AudioRecorder JSI interface.
- Adds sloppy (code quality) playback->recording->playback automated test screen

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added/Conducted relevant tests
- [x] Performed self-review of the code
- [ ] Updated Web Audio API coverage
- [ ] Added support for web
- [ ] Updated old arch android spec file

PR #1014, opened 2026-04-10
fix(ios): reuse AVAudioEngine instance in rebuildAudioEngine to prevent owningEngine assertion crash

### LINKED ISSUE #1013: iOS: interruption recovery can crash with AVAudioEngine owningEngine assertion when rebuildAudioEngine recreates the engine
### Description

We hit a production iOS crash in `react-native-audio-api` during interruption recovery while recording.

Crash signature:

```
SIGABRT — com.apple.coreaudio.avfaudio
required condition is false: nil == owningEngine || GetEngine() == owningEngine
```

The crash happens in the `AudioEngine.mm` interruption-recovery path, specifically when `rebuildAudioEngine` detaches nodes from the current `AVAudioEngine`, creates a new `AVAudioEngine` instance, and then re-attaches the same `AVAudioNode` objects to that new engine.

Current upstream `main` still does this in:

- `packages/react-native-audio-api/ios/audioapi/ios/system/AudioEngine.mm`
- `onInterruptionEnd`
- `startEngine`
- `rebuildAudioEngine`

Relevant flow on `main` as of 2026-04-10:

1. `onInterruptionEnd` calls `stopIfNecessary`, `reset`, then `rebuildAudioEngine`
2. `rebuildAudioEngine` detaches existing source/input nodes
3. `rebuildAudioEngine` creates a brand new `AVAudioEngine`
4. the old node objects are attached to the new engine instance

That appears to violate AVFAudio's engine ownership assumptions and can trip:

```
required condition is false: nil == owningEngine || GetEngine() == owningEngine
```

In our app, replacing the rebuild strategy with "reuse the same engine instance, disconnect/reconnect the graph, and do not move existing nodes across engine instances" stopped this specific crash.

Important caveat: I do not yet have a small public deterministic repro. The root cause is inferred from the production stack trace, the upstream code path, and the fact that the patch below removed the crash in production.

Observed production details:

- `react-native-audio-api`: `0.11.6` / `0.11.7` package line
- React Native: `0.83.4`
- Platform: iOS
- Device: iPhone 15 Pro Max
- OS: iOS 26.3.1
- App state: recording flow, around 5 minutes after app start
- Trigger pattern: interruption recovery / recorder lifecycle around phone calls, Siri, or similar audio-session interruptions

Representative stack excerpt:

```
0  CoreFoundation     (exception handling)
1  libobjc.A.dylib    objc_exception_throw
2  CoreFoundation     (assertion failure)
3  AVFAudio           (owningEngine assertion check)
4  AVFAudio           (internal node operation)
5  AVFAudio           (internal node operation)
6  app code           rebuildAudioEngine path
7  libdispatch.dylib  dispatch queue execution
```

Why I think the current code is unsafe:

- `AVAudioNode` instances appear to retain engine ownership internally beyond what this rebuild flow assumes.
- The current implementation detaches nodes from one engine and reuses those same node instances on a newly created engine.
- The interruption path also has extra `reset` calls before rebuild, which may increase the chance of stale internal state.
- There may also be a race component, but even without proving that, the cross-engine node reuse looks sufficient to explain the assertion.

Proposed fix: Reuse the same `AVAudioEngine` instance during rebuild instead of creating a new one.

The patch that fixed this for us:

- stop the engine if needed
- reset the existing engine
- disconnect node inputs/outputs
- reconnect the existing graph on the same engine instance
- remove the extra `reset` call sites before rebuild

### Steps to reproduce

I do not currently have a deterministic minimal repro.

What we know so far:

1. Start recording on iOS with `AudioRecorder`
2. Keep the recording/session alive for a while
3. Trigger an audio-session interruption or interruption-like recovery path
4. Let the library rebuild the engine
5. In rare cases, the app aborts inside AVFAudio with the `owningEngine` assertion above

I realize this is not ideal for maintainers. I'm filing because:

- the crash signature is specific
- the problematic code path is narrow
- `main` still contains the same rebuild logic
- the downstream patch is small and has already mitigated the crash for us

No public minimal repro yet.

0.11.7

### React Native version

0.83.4

### Platforms

iOS

### JavaScript runtime

Hermes

### Workflow

Expo Dev Client

### Architecture

Fabric (New Architecture)

### Build type

Release app & production bundle

### Device

Real device

### Device model

iPhone 15 Pro Max (iOS 26.3.1)

### Acknowledgements

Yes

radko93:
Closes #1013

## ⚠️ Breaking changes ⚠️

- None.

## Introduced changes

- reworks iOS interruption recovery in `AudioEngine.mm` to rebuild graph connections on the current `AVAudioEngine` instance instead of creating a new engine and re-attaching the same node objects
- removes redundant `reset` calls before `rebuildAudioEngine`, since rebuild now resets the current engine internally
- avoids reusing `AVAudioSourceNode` / `AVAudioSinkNode` instances across different `AVAudioEngine` instances, which can trigger AVFAudio's `owningEngine` assertion

## Notes for reviewers

- This change is intentionally narrow and only touches the iOS rebuild path in `packages/react-native-audio-api/ios/audioapi/ios/system/AudioEngine.mm`.
- I do not have a deterministic public repro yet. The fix is based on a production crash signature plus a downstream patch that mitigated the crash in production.
- Upstream `main` still had the detach-create-reattach flow when this PR was prepared on 2026-04-10.
- Open PR #1005 also touches `AudioEngine.mm`, but it still recreates the engine in `rebuildAudioEngine`, so it does not address this specific `owningEngine` assertion. If that PR lands first, this patch will probably need a small rebase.

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/ios/audioapi/ios/system/AudioEngine.mm
+++ b/packages/react-native-audio-api/ios/audioapi/ios/system/AudioEngine.mm
@@ -108,9 +108,8 @@ static AudioEngine *_sharedInstance = nil;
     return;
   }
 
-  // Stop just in case, reset the engine and build it from scratch
+  // Stop just in case and rebuild graph connections.
   [self stopIfNecessary];
-  [self.audioEngine reset];
   [self rebuildAudioEngine];
 
   // If shouldResume is false, mark the engine as paused and wait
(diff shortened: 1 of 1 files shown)


michalsek:
@radko93 Hey, thanks for the PR, this is a huge help in hardening the engine! 🙌 

Although we won't be able to use it as it is - reusing the engine, will cause crash/issues for other audio paths (i.e. media services were reset system notification to name one). 

But i've included fixes for your issue in my pull request :)

It needs some cleanup in the test code, but should be merged within a day or two, and available in nightly version shortly after. Although I can't promise when it will be released as an stable version or patch

radko93:
@michalsek no worries, thanks for the update feel free to close this one and update the linked issues when it's fixed.

mdydek:
fixes to the ios audio engine released in 0.12

PR #1019, opened 2026-04-14
fix: removed convolver trimming and unnecesary rerender

mdydek:
Convolver is not lagging with 15s impulse responses, thus I think trimming is no longer necessary and it can introduce some unwanted behavior

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- removed tail trimming in convolver
- fixed unnecesary rerenders in audio tag
- updated metro with additional audio file formats supported by us
- added ConvolverNode test screen

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


closetcaiman (approved):
lgtm!

PR #1023, opened 2026-04-15
feat: add param tests to test app

closetcaiman:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- added param tests to test app
- fixed broken tests for:
  - streamer
  - recorder

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/test-app-screen/src/AudioParamTest.ts
mdydek, line 1: make this test more verbose, add maybe artificial pauses, these tests should also provide visual explanation what is currently going on, but this implementation displays something for a fraction of a second and at the end it displays "scheduling exponential..."

mdydek (changes requested):
.

mdydek (approved):

PR #1045, opened 2026-05-02
feat: add api to concatenate audio files

Traviskn:
Refs #1043

  ## ⚠️ Breaking changes ⚠️

  

  - None

  ## Introduced changes

  

  - Add `concatAudioFiles(inputPaths, outputPath): Promise<string>` exported from the package root, joining a list of compatible local audio files into a single output file. Pairs naturally with the recorder's `rotateIntervalBytes` option to stitch rotated segments back into one file after recording.
  - Implement the underlying engine in C++ (`core/utils/AudioFileConcatenator`) using the bundled FFmpeg, exposed to JS through a new `AudioFileUtilsHostObject` and wired up in `AudioAPIModuleInstaller` as the `createAudioFileUtils` JSI global.
  - Add a web stub that rejects with a clear "not supported on web" error, plus mock + Jest coverage in `tests/mock.test.ts`, and a C++ unit test (`common/cpp/test/src/utils/AudioFileConcatenatorTest.cpp`).
  - Add docs page `packages/audiodocs/docs/utils/file-concatenation.mdx` and cross-link it from the `rotateIntervalBytes` section of the audio recorder docs.
  - Minor: exclude `common/cpp/test/build` from `format:*:common` and `cpplint.sh` so local C++ test builds don't trip linting/formatting.

## Checklist

  - [x] Linked relevant issue (discussion #1043)
  - [x] Added support for web (stub that rejects with a clear message)


packages/audiodocs/docs/utils/file-concatenation.mdx
```diff
@@ -0,0 +1,95 @@
+---
+sidebar_position: 3
+---
```
mdydek, line 4:
```suggestion
---
sidebar_position: 3
sidebar_label: File concatenation
---
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.h
mdydek, line 1: I think there is no reason for `AudioFileConcatenator` to be class, because it does not use any internal state and all of the functions are static, so they can be normal functions

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
mdydek, line 1: can you move declaration of classes to the .h file? I get it that it can be kind of inconvenient having to guard everything with ffmpeg macro, but right now this file is very big and hard to read. Also add brief comments what those methods do, it would help greatly in review process

mdydek (changes requested):
some changes needed before proper review, but at first glance it looks solid

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.h
Traviskn: Good call, I'll update this to just be functions

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
Traviskn: Yes I will do that!

Traviskn:
Updated this PR with the recommended changes!

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
```diff
@@ -0,0 +1,571 @@
...
+  auto *inputStream = firstInput.audioStream();
+  result = avcodec_parameters_copy((*outputStream)->codecpar, inputStream->codecpar);
+  if (result < 0) {
+    return Err("Failed to copy audio stream parameters: " + parseFFmpegError(result));
+  }
+
+  (*outputStream)->codecpar->codec_tag = 0;
+  (*outputStream)->time_base = inputStream->time_base;
+  if ((*outputStream)->time_base.num <= 0 || (*outputStream)->time_base.den <= 0) {
+    (*outputStream)->time_base = AVRational{1, inputStream->codecpar->sample_rate};
```
mdydek:
```suggestion
    (*outputStream)->time_base = AVRational{.num=1, .den=inputStream->codecpar->sample_rate};
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
```diff
@@ -0,0 +1,571 @@
...
+      return Err("Input file '" + filePath + "' does not contain an audio stream.");
+    }
+  }
+
+  if (!hasUsableAudioParameters(rawContext->streams[audioStreamIndex])) {
+    return Err("Input file '" + filePath + "' is missing required audio parameters.");
+  }
+
+  guard.release();
+  return Ok(InputFormatContext(filePath, rawContext, audioStreamIndex));
```
mdydek:
```suggestion
  return Ok(InputFormatContext(filePath, guard.release(), audioStreamIndex));
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
```diff
@@ -0,0 +1,571 @@
...
+}
+
+std::string parseFFmpegError(int errorCode) {
+  char errorBuffer[AV_ERROR_MAX_STRING_SIZE];
+
+  if (av_strerror(errorCode, errorBuffer, sizeof(errorBuffer)) < 0) {
+    return "Unknown FFmpeg error: " + std::to_string(errorCode);
+  }
+
+  return std::string(errorBuffer);
```
mdydek:
```suggestion
  std::array<char, AV_ERROR_MAX_STRING_SIZE> errorBuffer{};

  if (av_strerror(errorCode, errorBuffer.data(), sizeof(errorBuffer)) < 0) {
    return "Unknown FFmpeg error: " + std::to_string(errorCode);
  }

  return {errorBuffer.data()};
```

mdydek (approved):
impressive work, wonderful contribution! Only few nitpicks from the linter

mdydek (changes requested):
unfortunately, have to revert approve, code looks good, but it concats only m4a files (ffmpeg is not setup for wav and flac formats, miniaudio handles them and it fails because it meets unknown data), can you add also concatenating those formats, miniaudio operates on higher level than ffmpeg so it should be fairly simple. Also it would make us less dependent on ffmpeg, which is always better

Traviskn:
Unfortunately miniaudio only supports WAV encoding (mentioned in the features list on their README and verified in code https://github.com/mackron/miniaudio), so while I was able to use it for WAV files I still needed to use ffmpeg for concatenating other file formats. I tested WAV and FLAC and both were working. I also ran into a consistent cocoa pods error when installing into a local test app and included a fix in the download-prebuilt-binaries.sh script

packages/audiodocs/docs/utils/file-concatenation.mdx
```diff
@@ -0,0 +1,99 @@
...
+recording is finished.
+
+:::warning
+`concatAudioFiles` is not supported on web.
+:::
+
+:::caution
+FFmpeg-backed formats require FFmpeg to be available in the native build. WAV
+concatenation uses miniaudio and does not require FFmpeg. FLAC concatenation
+uses miniaudio for decoding and FFmpeg for encoding the output file.
```
mdydek:
```suggestion
concatenation uses miniaudio and does not require FFmpeg.
```
I guess for the end user it does not matter whether encoding or decoding requires ffmpeg, since they would not concatenate this format without ffmpeg

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
```diff
@@ -0,0 +1,1344 @@
...
+}
+
+std::string lowercaseExtension(const std::string &path) {
+  const auto dotIndex = path.find_last_of('.');
+  if (dotIndex == std::string::npos) {
+    return "";
+  }
+
+  std::string extension = path.substr(dotIndex + 1);
+  std::transform(extension.begin(), extension.end(), extension.begin(), [](unsigned char c) {
```
mdydek:
```suggestion
  std::ranges::transform(extension, extension.begin(), [](unsigned char c) {
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
```diff
@@ -0,0 +1,1344 @@
...
+    return static_cast<char>(std::tolower(c));
+  });
+
+  return extension;
+}
+
+bool hasExtension(const std::string &path, const std::vector<std::string> &extensions) {
+  const std::string extension = lowercaseExtension(path);
+  return std::find(extensions.begin(), extensions.end(), extension) != extensions.end();
```
mdydek:
```suggestion
  return std::ranges::find(extensions, extension) != extensions.end();
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
```diff
@@ -0,0 +1,1344 @@
...
+bool isFFmpegEncodedOutputPath(const std::string &path) {
+  return hasExtension(path, {"flac"});
+}
+
+bool isFFmpegOutputPath(const std::string &path) {
+  return isFFmpegRemuxOutputPath(path) || isFFmpegEncodedOutputPath(path);
+}
+
+std::string parseMiniAudioError(ma_result errorCode) {
+  return std::string(ma_result_description(errorCode));
```
mdydek:
```suggestion
  return {ma_result_description(errorCode)};
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
mdydek, line 1: same case with moving classes declaration to .h MiniAudioEncoderGuard, FFmpegEncodedOutputContext, MiniAudioDecoderGuard

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
```diff
@@ -0,0 +1,1344 @@
...
+    const int availableSpace = av_audio_fifo_space(audioFifo_);
+    if (availableSpace < frameCount) {
+      const int result = av_audio_fifo_realloc(
+          audioFifo_, av_audio_fifo_size(audioFifo_) + frameCount + frameSize_);
+      if (result < 0) {
+        return Err("Failed to grow FLAC encoding FIFO: " + parseFFmpegError(result));
+      }
+    }
+
+    void *inputData[1] = {const_cast<void *>(decodedFrames)};
```
mdydek: do not use const cast in this case: https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/pro-type-const-cast.html

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
```diff
@@ -0,0 +1,1344 @@
...
+
+  int result = av_write_trailer(output->get());
+  if (result < 0) {
+    return Err("Failed to write output trailer: " + parseFFmpegError(result));
+  }
+
+  return Ok(outputPath);
+}
+
+AudioFileConcatResult concatFLACFilesWithMiniAudioAndFFmpeg(
```
mdydek: I don't like the idea of mixing those two libraries together, concatenating flac still requires ffmpeg (I do not remember whether we have flac format in config, if not I think we can skip this format totally for now, and after ffmpeg update it should work immediately). But maybe it was some special reason you did that

mdydek (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
Traviskn: Yeah the current ffmpeg is not configured with the flac demuxer, so I had to workaround by decoding with miniaudio and encoding with ffmpeg. I can remove this code and flac file concatenation support for now and wait on an ffmpeg update instead 👍

Traviskn: The current configured miniaudio has a similar but opposite issue, it can only decode flac but not encode it

mdydek: yeah, that can be confusing sorry for that, but ffmpeg config as you can see is a bit tricky what it understands and what not

Traviskn: no worries, it is tricky!

Traviskn: removed this in the latest push

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
Traviskn: fixed in the latest push

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
Traviskn: fixed in the latest push

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
Traviskn: fixed in the latest push

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
Traviskn: fixed in the latest push

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
Traviskn: fixed in the latest push

packages/audiodocs/docs/utils/file-concatenation.mdx
Traviskn: good point, cleaned this up in the latest push

Traviskn:
The latest push should address all the comments/requested changes. Let me know if any more changes are needed, happy to make any adjustments!

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
```diff
@@ -0,0 +1,866 @@
...
+  }
+
+  if (isMiniaudioOutputPath(normalizedOutputPath)) {
+    return concatAudioFilesWithMiniAudio(normalizedInputPaths, normalizedOutputPath)
+        .map([&outputPath](const std::string &) { return outputPath; });
+  }
+
+  if (!isFFmpegOutputPath(normalizedOutputPath)) {
+    return Err(
+        "concatAudioFiles supports WAV output with miniaudio and M4A/MP4/CAF output with FFmpeg.");
```
mdydek: have you tested CAF concatenation? because in my testing it is not working (I think it is not supposed to in the first place), i can manage to get it working only with wav and m4a, if that is also yours case, just update mentions to exclude caf

mdydek: i managed to get it working with broader ffmpeg build, so this the last thing to correct

mdydek (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioFileConcatenator.cpp
Traviskn: Ah sorry I must have missed that, I'll update this to not mention CAF. Thanks for testing and catching that!

Traviskn: I have removed mentions of CAF files and removed an explicit check for CAF file extensions in the latest push

mdydek (approved):

PR #1050, opened 2026-05-06
refactor: decoder utilizing ffmpeg and ma decoders

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- decoder now utilizes already present ffmpeg and miniaudio decoding code

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioDecoder.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioDecoder.cpp
@@ -1,16 +1,15 @@
 #include <audioapi/core/utils/AudioDecoder.h>
 #include <audioapi/dsp/VectorMath.h>
 #include <audioapi/libs/base64/base64.h>
+#include <audioapi/libs/decoding/IncrementalAudioDecoder.h>
+#include <audioapi/libs/miniaudio/MiniAudioDecoding.h>
 #include <audioapi/utils/AudioArray.hpp>
-
-#include <audioapi/libs/miniaudio/decoders/libopus/miniaudio_libopus.h>
-#include <audioapi/libs/miniaudio/decoders/libvorbis/miniaudio_libvorbis.h>
-#include <audioapi/libs/miniaudio/miniaudio.h>
 
 #if !RN_AUDIO_API_FFMPEG_DISABLED
 #include <audioapi/libs/ffmpeg/FFmpegDecoding.h>
 #endif // RN_AUDIO_API_FFMPEG_DISABLED
 
+#include <algorithm>
 #include <memory>
 #include <string>
 #include <utility>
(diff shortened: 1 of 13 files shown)


packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioDecoder.cpp
```diff
@@ -148,7 +124,7 @@ AudioBufferResult AudioDecoder::decodeWithPCMInBase64(
   auto audioBuffer =
       std::make_shared<AudioBuffer>(numFramesDecoded, inputChannelCount, inputSampleRate);
 
-  for (size_t ch = 0; ch < inputChannelCount; ++ch) {
+  for (int ch = 0; ch < inputChannelCount; ++ch) {
```
closetcaiman, line 180: Why do we need to change `size_t` to `int`?

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioDecoder.cpp
```diff
@@ -1,139 +1,115 @@
...
-
-  ma_decoder_uninit(&decoder);
+bool needsFFmpeg(AudioFormat format) {
+  return format == AudioFormat::MP4 || format == AudioFormat::M4A || format == AudioFormat::AAC;
+}
 
-  return result;
+bool needsFFmpegByPath(const std::string &path) {
+  return AudioDecoder::pathHasExtension(path, {".mp4", ".m4a", ".aac"});
 }
```
closetcaiman: Shouldn't those be included in the header?

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioDecoder.cpp
```diff
@@ -1,139 +1,115 @@
...
-          return MA_INVALID_ARGS;
-        }
-      },
-      source);
-
-  if (initResult != MA_SUCCESS) {
-    return Err(
-        "Failed to initialize miniaudio decoder: " +
-        std::string(ma_result_description(initResult)));
-  }
```
closetcaiman, line 91: It would be cool to preserve decoder API errors (from ffmpeg/ma) provided they are informative. Not a must have as it would require to refactor `decoding::IncrementalDecoder` implementations.

closetcaiman (approved):
lgtm,

nice job deduplicating this!

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioDecoder.cpp
mdydek: because comparing unsigned to signed is a slightly bad thing (inputChannelCount is int in this case)

packages/react-native-audio-api/common/cpp/audioapi/libs/miniaudio/MiniAudioDecoding.cpp
closetcaiman, line 30: from clang-tidy: `Do not declare C-style arrays, use 'std::array' instead`

either resolve or ignore if it's `ma`-specific, but add a comment.

closetcaiman, line 35: from clang-tidy: `Do not implicitly decay an array into a pointer; consider using gsl::array_view or an explicit cast instead`

either resolve or ignore if it's ma-specific, but add a comment.

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioDecoder.h
```diff
@@ -1,119 +1,40 @@
...
-      const std::string &path,
-      float sampleRate);
-  [[nodiscard]] static AudioBufferResult
-  decodeWithMemoryBlock(const void *data, size_t size, float sampleRate);
-  [[nodiscard]] static AudioBufferResult decodeWithPCMInBase64(
-      const std::string &data,
-      float inputSampleRate,
-      int inputChannelCount,
-      bool interleaved);
+inline bool needsFFmpegByPath(const std::string &path) {
```
closetcaiman: Aren't those `[[nodiscard]]` too?

closetcaiman (changes requested):
Nice job including the decoder-specific errors. I also like the namespace usage instead of a class with static methods, it feels a lot more cpp-like and modern. I left a few nitpicks.

packages/react-native-audio-api/common/cpp/audioapi/libs/miniaudio/MiniAudioDecoding.cpp
mdydek: miniaudio is basically c library, so I think in this case we would like to leave it as it is

mdydek: miniaudio is basically c library, so I think in this case we would like to leave it as it is

closetcaiman: sure, It seemed like something specific to `miniaudio`, we can add explicit ignore-lint here then

closetcaiman (approved):

PR #1054, opened 2026-05-07
fix(recorder): copy audio buffer before async handoff to fix use-after-free

### LINKED ISSUE #1053: Use-after-free in iOS/Android recorder: buffer pointer dereferenced after callback returns
### Description

`IOSRecorderCallback::receiveAudioData` and `AndroidRecorderCallback::receiveAudioData` push the raw audio buffer pointer they receive from the platform onto an SPSC queue and dereference it asynchronously from the offloader worker thread. The buffer is only valid for the duration of the synchronous platform callback (CoreAudio render callback on iOS, Oboe data callback on Android), so by the time the worker thread reads from it the memory may have been reused or freed. This is a classic use-after-free.

The hot loop that crashes (iOS):

```cpp
// IOSRecorderCallback.mm — receiveAudioData
offloader_->getSender()->send({inputBuffer, numFrames});  // pointer goes onto a cross-thread queue
```

```cpp
// IOSRecorderCallback.mm — taskOffloaderFunction (runs on a different thread)
for (size_t i = 0; i < bufferFormat_.channelCount; ++i) {
  memcpy(
      converterInputBuffer_.mutableAudioBufferList->mBuffers[i].mData,
      inputBuffer->mBuffers[i].mData,    // <-- inputBuffer may be freed here
      inputBuffer->mBuffers[i].mDataByteSize);
}
```

The Android path has the same shape: `void *data` from Oboe is pushed onto the queue and read later from the worker.

### Production stack trace (iOS)

App version: react-native-audio-api 0.12.1, iOS, hit on a real device:

```
Crashed: Thread
0  libsystem_platform.dylib       0x9c0  _platform_memmove + 144
1  ApolloScooters                 0x1eecb84  audioapi::IOSRecorderCallback::taskOffloaderFunction(CallbackData) + 164 (IOSRecorderCallback.mm:164)
2  ApolloScooters                 0x1eed310  std::__thread_proxy<...IOSRecorderCallback::prepare(AVAudioFormat*, unsigned long)::$_0&...>(void*) + 81
3  libsystem_pthread.dylib        0x4438  _pthread_start + 136
4  libsystem_pthread.dylib        0x8cc  thread_start + 8
```

`+ 164` lands inside the `memcpy` loop on line 164–169 of `IOSRecorderCallback.mm`.

### Why it's hard to MRE

The bug is a memory race: the offloader worker has to be preempted long enough that CoreAudio reuses the buffer slot before the memcpy reads it, then it has to read into unmapped or zero-mapped pages to actually crash (otherwise it just silently consumes garbage audio). Reproducing it deterministically would require slowing the worker thread or dirtying buffers under the audio system. The bug is, however, clearly visible from inspection, and it shows up in production crash reports under sustained mic load (continuous wake-word / VAD streaming).

### Steps to reproduce in production

1. Open a long-running recording session via `AudioRecorder.onAudioReady(...)` and `recorder.start()` — anything that keeps the CoreAudio input render callback firing for minutes.
2. Cause heavy CPU contention on the device (background work, low-power mode, etc.).
3. Eventually the SPSC worker is delayed long enough that the source buffer slot is recycled before its memcpy runs → segfault inside `_platform_memmove`.

### Fix

PR: #1054 — copies the buffer into an owned allocation inside `receiveAudioData` and frees it after `taskOffloaderFunction` is done (along both fast and resampling paths). Same approach for both iOS (`AudioBufferList` + per-channel `mData`) and Android (`ma_malloc`/`ma_free`).

### Environment

- Library: react-native-audio-api 0.12.1 (also reproduced upstream `main` by inspection)
- Platforms: iOS (confirmed crash in the wild); Android has the same shape and is fixed in the same PR

christian-apollo:
Closes #1053

## Introduced changes

`IOSRecorderCallback::receiveAudioData` and `AndroidRecorderCallback::receiveAudioData` were pushing the raw audio buffer pointer received from the platform (CoreAudio render callback on iOS, Oboe data callback on Android) onto an SPSC queue and then dereferencing it asynchronously from the offloader worker thread. Both the iOS `AudioBufferList *` and the Android `void *` are owned by the audio system only for the duration of the synchronous callback — by the time the worker thread runs `memcpy` (iOS) or `ma_data_converter_process_pcm_frames` (Android), the source buffer can already be reused or freed. That's a use-after-free; in production it surfaces as a `_platform_memmove` segfault inside `IOSRecorderCallback::taskOffloaderFunction`.

This PR copies the buffer into an owned allocation in `receiveAudioData` and frees it after the worker is done with it on every code path:

- **iOS** — allocate a new `AudioBufferList` header plus per-channel `mData`, deep-copy the contents, send the owned pointer through the queue, and free it via a small `freeOwnedAudioBufferList` helper after both the fast (matching format) and the converter (resampling) paths have read from it.
- **Android** — `ma_malloc` an owned buffer, `std::memcpy` the frames, and `ma_free` after both the fast and resampling paths in `taskOffloaderFunction`. Also added a guard that ignores the `data == nullptr` sentinel that the `TaskOffloader` destructor sends to unblock the worker, since that sentinel carries no allocation to free.

The fix is symmetric on both platforms; the underlying mistake (capturing a borrowed buffer and reading it across threads) was the same on both sides.

## Reproduction

The race is hard to reproduce deterministically — it depends on the SPSC worker being preempted long enough for the audio system to reuse the source slot before the worker reads it. Inspection alone shows the bug, and the production crash trace in #1053 lands exactly inside the iOS `memcpy` loop at `IOSRecorderCallback.mm:164`.

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation — n/a, internal correctness fix
- [ ] Added/Conducted relevant tests — the bug is a memory race in native audio thread paths; I'm not aware of an existing harness for these; happy to add one if you can point me at the right place
- [ ] Updated Web Audio API coverage — n/a
- [ ] Added support for web — n/a
- [ ] Updated old arch android spec file — n/a

### DIFF
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/utils/AndroidRecorderCallback.cpp
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/utils/AndroidRecorderCallback.cpp
@@ -7,6 +7,7 @@
 #include <audioapi/utils/CircularArray.hpp>
 
 #include <algorithm>
+#include <cstring>
 #include <memory>
 #include <string>
 #include <unordered_map>
@@ -141,7 +142,18 @@ void AndroidRecorderCallback::receiveAudioData(void *data, int numFrames) {
     return;
   }
 
-  offloader_->getSender()->send({data, numFrames});
+  // Oboe owns `data` only for the duration of this synchronous callback.
+  // Copy into an owned buffer before handing off to the worker thread; the
+  // consumer in taskOffloaderFunction frees it.
+  size_t bytes =
+      static_cast<size_t>(numFrames) * streamChannelCount_ * ma_get_bytes_per_sample(ma_format_f32);
+  void *owned = ma_malloc(bytes, nullptr);
+  if (owned == nullptr) {
+    return;
+  }
+  std::memcpy(owned, data, bytes);
+
+  offloader_->getSender()->send({owned, numFrames});
 }
 
 /// @brief Deinterleaves the audio data and pushes it into the circular buffer.
@@ -160,6 +172,13 @@ void AndroidRecorderCallback::deinterleaveAndPushAudioData(void *data, int numFr
 /// processes it (resampling and deinterleaving if necessary), and pushes it into the circular buffer.
 void AndroidRecorderCallback::taskOffloaderFunction(CallbackData callbackData) {
   auto [data, numFrames] = callbackData;
+
+  // The TaskOffloader destructor sends a default-constructed CallbackData
+  // (data == nullptr) to unblock the receiver; ignore it here.
+  if (data == nullptr) {
+    return;
+  }
+
   ma_uint64 inputFrameCount = numFrames;
   ma_uint64 outputFrameCount = 0;
 
@@ -170,6 +189,7 @@ void AndroidRecorderCallback::taskOffloaderFunction(CallbackData callbackData) {
     if (circularBuffer_[0]->getNumberOfAvailableFrames() >= bufferLength_) {
       emitAudioData();
     }
+    ma_free(data, nullptr);
     return;
   }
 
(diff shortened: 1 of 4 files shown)


mdydek (approved):
nice one, while reviewing found another fatal race condition (destructor of the callback object could be fired in the same time when receiver function, thus making f.e. converterOutputBuffer_ invalid), I fixed it in this pr

closetcaiman (approved):

PR #1073, opened 2026-05-18
fix: cleaning up nodes

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- previous implementation always skipped first element of the vector, due to bug (analyze case where vector had 1 element and `AudioGraphManager::canBeDestructed(vec[0])` would return true

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


closetcaiman (approved):
lgtm

PR #1075, opened 2026-05-19
feat: preloading in audio tag

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioFileUtilsHostObject.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioFileUtilsHostObject.cpp
@@ -3,6 +3,7 @@
 #if !RN_AUDIO_API_FFMPEG_DISABLED
 #include <audioapi/libs/ffmpeg/FFmpegDecoding.h>
 #endif // RN_AUDIO_API_FFMPEG_DISABLED
+#include <audioapi/core/utils/AudioDecoding.h>
 #include <audioapi/jsi/JsiPromise.h>
 #include <audioapi/libs/miniaudio/MiniAudioDecoding.h>
 
@@ -19,14 +20,15 @@ namespace audioapi {
 namespace {
 
 std::optional<double> probeDurationWithDecoder(const uint8_t *data, size_t size, int sampleRate) {
-  if (auto miniaudioDuration =
-          miniaudio_decoder::MiniAudioDecoder::probeDuration(data, size, sampleRate);
-      miniaudioDuration.has_value()) {
-    return miniaudioDuration;
+  auto duration = std::optional<double>();
+  duration =
+      audiodecoding::probeDuration<miniaudio_decoder::MiniAudioDecoder>(data, size, sampleRate);
+  if (duration.has_value()) {
+    return duration;
   }
-
 #if !RN_AUDIO_API_FFMPEG_DISABLED
-  return ffmpegdecoder::FFmpegDecoder::probeDuration(data, size, sampleRate);
+  duration = audiodecoding::probeDuration<ffmpeg_decoder::FFmpegDecoder>(data, size, sampleRate);
+  return duration;
 #else
   return std::nullopt;
 #endif // RN_AUDIO_API_FFMPEG_DISABLED
--- a/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioFileUtilsHostObject.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioFileUtilsHostObject.cpp
@@ -20,8 +20,7 @@ namespace audioapi {
 namespace {
 
 std::optional<double> probeDurationWithDecoder(const uint8_t *data, size_t size, int sampleRate) {
-  auto duration = std::optional<double>();
-  duration =
+  auto duration =
       audiodecoding::probeDuration<miniaudio_decoder::MiniAudioDecoder>(data, size, sampleRate);
   if (duration.has_value()) {
     return duration;
--- a/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioFileUtilsHostObject.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioFileUtilsHostObject.cpp
@@ -1,20 +1,46 @@
 #include <audioapi/HostObjects/utils/AudioFileUtilsHostObject.h>
 #include <audioapi/core/utils/AudioFileConcatenator.h>
+#if !RN_AUDIO_API_FFMPEG_DISABLED
+#include <audioapi/libs/ffmpeg/FFmpegDecoding.h>
+#endif // RN_AUDIO_API_FFMPEG_DISABLED
 #include <audioapi/jsi/JsiPromise.h>
+#include <audioapi/libs/miniaudio/MiniAudioDecoding.h>
 
 #include <jsi/jsi.h>
+#include <cstdint>
 #include <memory>
+#include <optional>
 #include <string>
 #include <utility>
 #include <vector>
 
 namespace audioapi {
 
+namespace {
+
+std::optional<double> probeDurationWithDecoder(const uint8_t *data, size_t size, int sampleRate) {
+  if (auto miniaudioDuration =
+          miniaudio_decoder::MiniAudioDecoder::probeDuration(data, size, sampleRate);
+      miniaudioDuration.has_value()) {
+    return miniaudioDuration;
+  }
+
+#if !RN_AUDIO_API_FFMPEG_DISABLED
+  return ffmpegdecoder::FFmpegDecoder::probeDuration(data, size, sampleRate);
+#else
+  return std::nullopt;
+#endif // RN_AUDIO_API_FFMPEG_DISABLED
+}
+
+} // namespace
+
 AudioFileUtilsHostObject::AudioFileUtilsHostObject(
     jsi::Runtime *runtime,
     const std::shared_ptr<react::CallInvoker> &callInvoker) {
   promiseVendor_ = std::make_shared<PromiseVendor>(runtime, callInvoker);
-  addFunctions(JSI_EXPORT_FUNCTION(AudioFileUtilsHostObject, concatAudioFiles));
+  addFunctions(
+      JSI_EXPORT_FUNCTION(AudioFileUtilsHostObject, concatAudioFiles),
+      JSI_EXPORT_FUNCTION(AudioFileUtilsHostObject, probeDuration));
 }
 
 JSI_HOST_FUNCTION_IMPL(AudioFileUtilsHostObject, concatAudioFiles) {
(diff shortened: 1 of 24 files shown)


packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioFileUtilsHostObject.cpp
```diff
@@ -1,20 +1,46 @@
...
+std::optional<double> probeDurationWithDecoder(const uint8_t *data, size_t size, int sampleRate) {
+  if (auto miniaudioDuration =
+          miniaudio_decoder::MiniAudioDecoder::probeDuration(data, size, sampleRate);
+      miniaudioDuration.has_value()) {
+    return miniaudioDuration;
+  }
+
+#if !RN_AUDIO_API_FFMPEG_DISABLED
+  return ffmpegdecoder::FFmpegDecoder::probeDuration(data, size, sampleRate);
+#else
```
closetcaiman: Do these need to be methods of `IncrementalDecoder`? Can't they be just of the `decoding` namespace? Both initialize the respective decoders and operate on them in the scope of the function, just like `readWithMemoryBlock`.

closetcaiman, line 31: Namespace case seems inconsistent: `miniaudio_decoder` vs `ffmpegdecoder`. Maybe just all in lowercase?

packages/react-native-audio-api/common/cpp/audioapi/libs/ffmpeg/FFmpegDecoding.cpp
```diff
@@ -462,6 +462,24 @@ decoding::DecoderResult FFmpegDecoder::seekToTime(double seconds) {
...
+  }
+
+  const auto duration = static_cast<double>(decoder.getDurationInSeconds());
+  if (duration <= 0) {
+    return std::nullopt;
+  }
+
+  return duration;
+}
+
```
closetcaiman: There is no difference in the implementation between `miniaudiodecoder` and `ffmpegdecoder`. Maybe we can actually abstract that into a `template` method:

```cpp
template <typename DecoderType>
std::optional<double> probeDuration();
```

this can live in `IncrementalDecoder` probably.

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -154,29 +169,46 @@ const Audio = React.forwardRef<AudioTagHandle, AudioProps>((props, ref) => {
...
+              url: path,
+              headers: requestHeaders,
+              startBytes: SEGMENT_SIZE,
+              endBytes: SEGMENT_SIZE,
+            });
+            const probedDuration = await probeDuration(
+              prefetchedData,
+              context?.sampleRate
+            );
+            if (probedDuration != null && probedDuration > 0) {
```
closetcaiman: If `probedDuration` is not null it is for sure `> 0`. `null` is falsy in js, so probably this check can just be `if(probledDuration)`.

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioFileUtilsHostObject.cpp
```diff
@@ -1,20 +1,46 @@
...
 namespace audioapi {
 
+namespace {
+
+std::optional<double> probeDurationWithDecoder(const uint8_t *data, size_t size, int sampleRate) {
+  if (auto miniaudioDuration =
+          miniaudio_decoder::MiniAudioDecoder::probeDuration(data, size, sampleRate);
+      miniaudioDuration.has_value()) {
+    return miniaudioDuration;
+  }
```
closetcaiman: This is not pleasant to read. Can we move the initialisation out of the if statement?

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
closetcaiman: Can this logic be segmented? The data fetching logic is already complicated and this only makes it harder to read. We can also add some comments (like why `SEGMENT_SIZE` is set to 1024*16). Some of the `bool` checks can also be extracted out of ifs (like the coded check can be assigned to a variable so that its more readable).

packages/react-native-audio-api/src/utils/metadataPrefetching.ts
```diff
@@ -0,0 +1,75 @@
...
+  buffer: ArrayBuffer;
+  status: number;
+};
+
+export async function prefetchFileSegments({
+  url,
+  headers,
+  startBytes,
+  endBytes,
+}: PrefetchConfig): Promise<ArrayBuffer> {
```
closetcaiman, line 39: What is the purpose of this helper? Why does it try to prefetch both startBytes and endBytes and outputs either the combination or the single segments? Can we add a docstring here please?

packages/react-native-audio-api/src/development/react/Audio/utils.ts
```diff
@@ -17,13 +17,18 @@ export function withPropsDefaults(
   props: AudioProps,
   resolvedContext: BaseAudioContext | undefined
 ): AudioPropsBase {
+  const normalizedPreload =
+    (props.preload as string | undefined) === ''
+      ? 'auto'
+      : (props.preload ?? 'auto');
+
```
closetcaiman: This can be just:

```js
props.preload || 'auto'
```

as '' is falsy. If it's not obvious at the first glance you can add a comment above it, but I'm not a fan of this frankenstein that is now.

packages/react-native-audio-api/common/cpp/audioapi/libs/ffmpeg/FFmpegDecoding.h
```diff
@@ -73,6 +74,12 @@ class FFmpegDecoder : public decoding::IncrementalAudioDecoder {
 
   [[nodiscard]] decoding::DecoderResult seekToTime(double seconds) override;
 
+  /// Opens only enough decoder state to probe media duration from memory and then closes.
+  [[nodiscard]] static std::optional<double> probeDuration(
+      const void *data,
+      size_t size,
+      int outputSampleRate = 0);
+
```
closetcaiman: This method is not defined in `IncrementalDecoder` header yet both classes implement it as if it was a part of the interface.

closetcaiman (changes requested):
nice!

I'm not a fan of the current state of `Audio.tsx`. Maybe we can consider refactoring it for readability in the future?

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioDecoding.cpp
```diff
@@ -197,4 +197,17 @@ AudioBufferResult decodeWithPCMInBase64(
   return Ok(std::move(audioBuffer));
 }
 
-} // namespace audioapi::audiodecoder
+template <typename T>
+concept Decoder = std::is_base_of_v<decoding::IncrementalAudioDecoder, T>;
+
+template <Decoder D>
```
closetcaiman:
```suggestion
template <typename D>
  requires std::is_base_of_v<decoding::IncrementalAudioDecoder, D>
```

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/utils/AudioFileUtilsHostObject.cpp
```diff
@@ -1,20 +1,48 @@
...
 #include <vector>
 
 namespace audioapi {
 
+namespace {
+
+std::optional<double> probeDurationWithDecoder(const uint8_t *data, size_t size, int sampleRate) {
+  auto duration = std::optional<double>();
+  duration =
+      audiodecoding::probeDuration<miniaudio_decoder::MiniAudioDecoder>(data, size, sampleRate);
```
closetcaiman:
```suggestion
  auto duration =
      audiodecoding::probeDuration<miniaudio_decoder::MiniAudioDecoder>(data, size, sampleRate);
```

packages/react-native-audio-api/src/core/AudioFileUtils.ts
```diff
@@ -36,6 +37,13 @@ class AudioFileUtils {
...
     return this.fileUtils.concatAudioFiles(inputPaths, outputPath);
   }
+
+  public async probeDurationInstance(
+    data: ArrayBuffer,
+    sampleRate?: number
+  ): Promise<number | null> {
+    return this.fileUtils.probeDuration(data, sampleRate);
+  }
 }
```
closetcaiman, line 47: Is this used anywhere else? I think it can be safely removed and the fetching line can be moved to `probeDuration`. Also, remember to use `await` here.

packages/react-native-audio-api/src/development/react/Audio/utils.ts
```diff
@@ -17,13 +17,15 @@ export function withPropsDefaults(
...
   return {
     ...props,
     autoPlay: props.autoPlay ?? false,
     controls: props.controls ?? false,
     loop: props.loop ?? false,
     muted: props.muted ?? false,
-    preload: props.preload ?? 'auto',
+    preload: normalizedPreload,
     source: props.source ?? [],
```
closetcaiman:
```suggestion
return {
...props,
autoPlay: props.autoPlay ?? false,
controls: props.controls ?? false,
loop: props.loop ?? false,
muted: props.muted ?? false,
preload: props.preload || 'auto',
source: props.source ?? [],
```

packages/react-native-audio-api/src/development/react/Audio/Audio.tsx
```diff
@@ -44,35 +42,18 @@ const Audio = React.forwardRef<AudioTagHandle, AudioProps>((props, ref) => {
...
-
   const [playbackState, setPlaybackState] =
     useState<AudioTagPlaybackState>('idle');
   const [currentTime, setCurrentTime] = useState(0);
-  const [duration, setDuration] = useState(0);
+
+  const path = useMemo(() => resolveSourcePath(source), [source]);
+
+  const preloadMode: PreloadType =
+    preload === 'none' || preload === 'metadata' ? preload : 'auto';
```
closetcaiman: Is this even necessary if we use `useStableAudioProps`?

packages/react-native-audio-api/common/cpp/audioapi/libs/ffmpeg/FFmpegDecoding.cpp
```diff
@@ -580,13 +580,16 @@ decoding::DecoderResult FFmpegDecoder::seekToTime(double) {
 size_t FFmpegDecoder::readPcmFrames(float *, size_t) {
   return 0;
 }
+std::optional<double> FFmpegDecoder::probeDuration(const void *, size_t, int) {
+  return std::nullopt;
+}
```
closetcaiman: This is not defined in the header. What is this for?

closetcaiman (changes requested):
Small nitpicks. `JD` overall!

packages/react-native-audio-api/src/core/AudioFileUtils.ts
closetcaiman: Ughh, the line selection was cut. The full function is:

```js
public async probeDurationInstance(
    data: ArrayBuffer,
    sampleRate?: number
  ): Promise<number | null> {
    return await this.fileUtils.probeDuration(data, sampleRate);
  }
```

mdydek: i think its ok, because otherwise we will have to call heavy `global.createAudioFileUtils()`, but in this implementation it is called once and remembered in singleton

closetcaiman (approved):

PR #1079, opened 2026-05-28
feat: audio tag hls support

closetcaiman:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- added `HLS` support in `<Audio>`
  -  added `SeekDecoderDaemon` class which hold `SPSC` s for communication between JS thread (seek commands) and Audio thread (decoded frames). The class is used as a daemon thread (via `operator()()`) that is spawned on `play` and joined on `disable`. Uses `decoding::IncrementalAudioDecoder` depending on the file format.
  - now `AudioFileSourceNode` handles 3 threads:
    - JS -> initializes the node and requests seek commands
    - SeekDecoderDaemon -> listens for seek commands, performs `decoder->seekToTime`, drains frame `SPSC` of stale frames in case of seek, pushes the decoded frames into the frame `SPSC`
    - Audio -> processes the node, reads from frame `SPSC`, in case of seek request outputs silence
  - updated `NodeOptionsParser` to handle `HLS` format and throw in case of ffmpeg-less build when processing ffmpeg-required formats
- deprecated `StreamerNode`
  - the implementation is still in the codebase
  - added `@deprecated` annot on the JS side
  - updated the docs

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/audiodocs/docs/sources/streamer-node.mdx
```diff
@@ -7,6 +7,10 @@ import { Optional, ReadOnly } from '@site/src/components/Badges';
 
 # StreamerNode
 
+:::danger
+Deprecated. Use the [`<Audio>`](/docs/experimental/audio-tag) element for streaming audio instead. This node will be removed in a future major release.
```
mdydek: to achieve same behviour audio element should be used with `MediaElementAudioSourceNode` mention it here also

mdydek: and maybe also for now do not mention about when this node will be removed

packages/react-native-audio-api/common/cpp/audioapi/core/utils/decoding/SeekDecoderDaemon.cpp
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    if (!hasPendingChunk) {
+      continue;
+    }
+
+    if (frameSender_.try_send(localData) == ResponseStatus::SUCCESS) {
+      hasPendingChunk = false;
+    } else {
+      std::this_thread::yield();
```
mdydek: as we discussed, sleeping thread here cuts down processor usage by 20x

packages/react-native-audio-api/common/cpp/audioapi/core/utils/decoding/SeekDecoderDaemon.cpp
```diff
@@ -0,0 +1,120 @@
...
+      sharedState_->onPositionChangedFlush.store(true, std::memory_order_release);
+    }
+    seekHappened = true;
+  }
+  if (seekHappened) {
+    // Drain stale pre-seek frames before releasing the seek gate on the audio thread
+    DecoderData drop;
+    while (frameReceiverForDrain_->try_receive(drop) == ResponseStatus::SUCCESS) {}
+    sharedState_->pendingOffloadedSeeks.fetch_sub(1, std::memory_order_release);
+  }
```
mdydek, line 79:
```suggestion
  while (commandReceiver_.try_receive(seekReq) == ResponseStatus::SUCCESS) {
    if (decoder_ && decoder_->isOpen() && decoder_->seekToTime(seekReq.seconds).is_ok()) {
      sharedState_->currentTime.store(seekReq.seconds, std::memory_order_release);
      sharedState_->onPositionChangedFlush.store(true, std::memory_order_release);
          sharedState_->pendingOffloadedSeeks.fetch_sub(1, std::memory_order_release);
    }
    seekHappened = true;
  }
  if (seekHappened) {
    // Drain stale pre-seek frames before releasing the seek gate on the audio thread
    DecoderData drop;
    while (frameReceiverForDrain_->try_receive(drop) == ResponseStatus::SUCCESS) {}
  }
```

has to be there, because when more than one command about seeking is in spsc, it would increment atomic variable f.e. to 2, but it would decrease it only one freezing the playback infinitely

packages/react-native-audio-api/common/cpp/audioapi/core/utils/decoding/SeekDecoderDaemon.cpp
mdydek, line 1: also seeking behavior (disabling playback because of pending seeks) maybe could be orchestrated differently, right now adding few of them makes the pause a little bit long to me, bad ux

packages/react-native-audio-api/common/cpp/audioapi/core/utils/decoding/SeekDecoderDaemon.h
```diff
@@ -0,0 +1,115 @@
...
+};
+
+struct DecoderData {
+  std::array<
+      float,
+      static_cast<size_t>(audioapi::RENDER_QUANTUM_SIZE) *
+          static_cast<size_t>(audioapi::MAX_CHANNEL_COUNT)>
+      interleavedBuffer{};
+  size_t size{};
+};
```
mdydek, line 64: kind of big array, can we maybe optimize it in some way? lazy loading and creating it after we know channel count from the decoder, just suggestions

mdydek (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/utils/decoding/SeekDecoderDaemon.h
closetcaiman: it's not that straightforward for the `std::array`. it is in fact the consequence of the architectural decision to have a daemon thread handle `hls` for audio which is correct. we can try to investigate possible improvements later

packages/react-native-audio-api/common/cpp/audioapi/core/utils/decoding/SeekDecoderDaemon.cpp
closetcaiman: added seek debouncing, to drain command SPSC of all requests and then exectute only the last one, after that `fetch_sub` `request_count`

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioFileSourceNode.cpp
```diff
@@ -23,39 +28,84 @@ AudioFileSourceNode::AudioFileSourceNode(
...
-      spsc::WaitStrategy::ATOMIC_WAIT>>(
-      SEEK_OFFLOADER_WORKER_COUNT, [this](OffloadedSeekRequest req) { runOffloadedSeekTask(req); });
-
   isInitialized_.store(true, std::memory_order_release);
 }
 
+AudioFileSourceNode::~AudioFileSourceNode() {
+  // AudioNode::~AudioNode() -> cleanup() does not call the virtual disable(),
+  // so the daemon thread is not guaranteed to have been joined yet. A still
+  // joinable std::thread would call std::terminate() on destruction.
```
mdydek: i don't think this comment gives some important insight, tbh it can be removed

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.h
```diff
@@ -71,6 +71,8 @@ namespace audioapi {
 
 struct StreamerOptions;
 
+[[deprecated(
+    "StreamerNode is legacy and will be removed. Use AudioFileSourceNode via the <Audio> component instead.")]];
```
mdydek: we do not want to expose that probably, I mean our end users won't see it anyway because they touch only js side not underlying cpp

packages/react-native-audio-api/common/cpp/audioapi/core/utils/decoding/SeekDecoderDaemon.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+  decoding::DecoderResult openResult = Ok(None);
+
+  int contextSampleRate = static_cast<int>(options.contextSampleRate);
+  if (!options.filePath.empty()) {
+    openResult = decoder_->openFile(contextSampleRate, options.filePath);
+  } else if (!options.memoryData.empty()) {
+    openResult = decoder_->openMemory(
+        contextSampleRate, options.memoryData.data(), options.memoryData.size());
```
mdydek:
```suggestion
  } else {
    openResult = decoder_->openMemory(
        contextSampleRate, options.memoryData.data(), options.memoryData.size());
```
existence of one of this option is checked in `AudioFileSourceNode`, we can probably simplified this branch and maybe add comment about it

packages/react-native-audio-api/common/cpp/audioapi/core/utils/decoding/SeekDecoderDaemon.cpp
```diff
@@ -0,0 +1,135 @@
...
+      frameReceiverForDrain_(std::move(frameReceiver)) {
+  if (options.requiresFFmpeg) {
+#if !RN_AUDIO_API_FFMPEG_DISABLED
+    decoder_ = std::make_unique<ffmpeg_decoder::FFmpegDecoder>();
+#endif
+  } else {
+    decoder_ = std::make_unique<miniaudio_decoder::MiniAudioDecoder>();
+  }
+
+  decoding::DecoderResult openResult = Ok(None);
```
mdydek:
```suggestion
  decoding::DecoderResult openResult = Err("Some init value");
```
I think initializing it with err value would be semantically more correct

mdydek: it should be replaced of course further in the code

packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioFileSourceNode.h
```diff
@@ -175,16 +152,15 @@ class AudioFileSourceNode : public AudioScheduledSourceNode {
   uint64_t onPositionChangedCallbackId_ = 0;
   int onPositionChangedInterval_;
   int onPositionChangedTime_ = 0;
-  std::atomic<bool> onPositionChangedFlush_{true};
+  // True when the audio thread itself wrote onPositionChangedFlush (e.g. EOF),
+  // so sendOnPositionChangedEvent knows not to sync currentTime_ from the daemon.
+  bool audioThreadSetFlush_{false};
```
mdydek: I'm wondering if that is needed, can't we think of a way to simplify this logic and then sync times between threads in simpler manner?

packages/react-native-audio-api/common/cpp/audioapi/core/sources/StreamerNode.h
closetcaiman: Okay, it was more of a developer notice anyway, removed.

mdydek (approved):

PR #1086, opened 2026-06-04
fix: fall back to non-pmr BoundedPriorityQueue on runtimes without std::pmr

### LINKED ISSUE #1071: iOS 16 simulator/runtime crash caused by std::pmr usage in BoundedPriorityQueue.hpp
# iOS 16 simulator/runtime crash caused by std::pmr usage in BoundedPriorityQueue.hpp

## Environment

- react-native-audio-api: 0.12.2 (also reproduced on nightly)
- Expo SDK: 54
- React Native: 0.81.x
- Xcode: 26.3
- iOS simulator runtime: iOS 16.4
- Architecture: arm64

## Crash

App crashes immediately at launch on iOS 16 simulator with:
```
txt dyld: Symbol not found: __ZNSt3__13pmr15memory_resourceD2Ev  Referenced from: MyAppName.debug.dylib  Expected in: .../iOS 16.4.simruntime/.../usr/lib/libc++.1.dylib 
```
Demangled:
```
txt std::__1::pmr::memory_resource::~memory_resource() 
```
This is a dyld launch crash before JS executes.

## Investigation

Initially I suspected FFmpeg / bundled static libs (related to #987), but even after using:

```js
 disableFFmpeg: true, 
 disableStaticExternalLibs: true, 
```

the crash still occurred.

The actual source appears to be:

```cpp 
common/cpp/audioapi/utils/BoundedPriorityQueue.hpp 
```

which currently uses:

```cpp
#include <memory_resource>  std::pmr::multiset std::pmr::monotonic_buffer_resource std::pmr::unsynchronized_pool_resource 
```

## Workaround / Patch

Replacing the PMR containers/resources with standard STL containers fixes the issue completely on iOS 16 simulator.

[react-native-audio-api+0.12.2.patch](https://github.com/user-attachments/files/27763353/react-native-audio-api%2B0.12.2.patch)

After this patch:

- app launches correctly on iOS 16 simulator
- no dyld crash
- no need to disable FFmpeg/static external libs

## Question

Is `std::pmr` intended to be supported on iOS 15/16 runtimes?

It looks like current builds require newer libc++ runtime support than what is available in iOS 16 runtimes.

Would you consider:
- removing PMR usage,
- gating it behind newer deployment targets,
- or providing a fallback STL implementation for older iOS runtimes?

Leppard:
## Summary

`BoundedPriorityQueue` is backed by `std::pmr` (`multiset` + `monotonic_buffer_resource` + `unsynchronized_pool_resource`). On Apple platforms those runtime symbols only exist in the system `libc++` from **iOS 17.0 / macOS 14.0 / tvOS 17.0 / watchOS 10.0** (`_LIBCPP_AVAILABILITY_HAS_PMR` → `_LIBCPP_INTRODUCED_IN_LLVM_16`).

libc++ currently leaves the pmr availability **markup empty** (see the `// TODO: Enable std::pmr markup once llvm/llvm-project#40340 ...` note in `<__configuration/availability.h>`), so using `std::pmr` below those versions **compiles without any diagnostic** but aborts at **dyld load time**:

```
Symbol not found: std::__1::pmr::memory_resource::~memory_resource()
```

That crashes every app at launch on iOS < 17 — see #1071 (also confirmed there on production App Store builds and real devices).

## Approach

Following @mdydek's preference in #1071 (keep `std::pmr`, add an STL fallback for older runtimes):

- Gate the `std::pmr` implementation behind a new `AUDIOAPI_HAS_PMR` macro, **auto-detected from libc++'s own `_LIBCPP_AVAILABILITY_HAS_PMR`** so it tracks each platform's threshold automatically (Android keeps pmr). It can be overridden by defining `AUDIOAPI_HAS_PMR` before including the header.
- When pmr is unavailable, back the **same `std::multiset`** with an in-object, non-pmr pool allocator (`FixedBlockPool` + `PoolAllocator`) that preserves the original design intent: a fixed buffer, node recycling via an intrusive free list, and **no heap allocation** — so the audio-thread / real-time behavior is kept on the fallback too, not regressed to a heap-backed `std::multiset`.

The container type and the entire public API are unchanged, so **iOS 17+ / Android keep the existing `std::pmr` fast path untouched** — only sub‑iOS‑17 (and equivalent) builds compile the fallback.

## Tests

Adds `common/cpp/test/src/utils/BoundedPriorityQueueTest.cpp`. It `#define`s `AUDIOAPI_HAS_PMR 0` so the **fallback path is exercised even on CI hosts whose libc++ has pmr** (the default path is already compiled into the rest of the suite). Covers:

- ascending order + capacity bounds (full → rejected),
- `pop` returns the smallest element first,
- **node recycling under 100k insert/erase cycles** (proves the free list reclaims; no pool exhaustion),
- heterogeneous `lowerBound` / `upperBound` (transparent comparator),
- `extract` → mutate key → reinsert with hint, and range `erase`.

Verified locally: builds clean for `arm64-apple-ios15.1` with **zero `pmr` symbols** in the resulting object (`nm`), the default (pmr) branch still compiles on a pmr-capable host, and all 4 gtest cases pass. (ASan/UBSan/TSan variants left to CI — the local macOS 26 ASan runtime hangs at startup even for a trivial program.)

## Notes

- No podspec change needed — the `ios_min_version = '14.0'` claim becomes correct again with the fallback in place.
- Happy to adjust the macro name/placement, or split the fallback allocator into its own header, if you'd prefer.

Fixes #1071

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/utils/BoundedPriorityQueue.hpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/utils/BoundedPriorityQueue.hpp
@@ -4,11 +4,132 @@
 #include <array>
 #include <cstddef>
 #include <functional>
-#include <memory_resource>
 #include <set>
 #include <utility>
 
+// --- std::pmr availability gate ---------------------------------------------
+// std::pmr's runtime symbols ship in Apple's system libc++ only from
+// iOS 17.0 / macOS 14.0 / tvOS 17.0 / watchOS 10.0 onward
+// (_LIBCPP_AVAILABILITY_HAS_PMR maps to _LIBCPP_INTRODUCED_IN_LLVM_16).
+// libc++ currently leaves the pmr availability *markup* empty
+// (see llvm/llvm-project#40340), so using std::pmr below those versions
+// compiles without any diagnostic but fails at dyld load time:
+//   Symbol not found: std::__1::pmr::memory_resource::~memory_resource()
+// On those targets we fall back to a std::multiset backed by an in-object,
+// non-pmr pool allocator that preserves the original design intent: a fixed
+// buffer, node recycling, and no heap allocation (audio-thread / real-time
+// safe). Define AUDIOAPI_HAS_PMR before including this header to override.
+#if !defined(AUDIOAPI_HAS_PMR)
+#  if defined(_LIBCPP_VERSION) && defined(_LIBCPP_AVAILABILITY_HAS_PMR) && \
+      !_LIBCPP_AVAILABILITY_HAS_PMR
+#    define AUDIOAPI_HAS_PMR 0
+#  else
+#    define AUDIOAPI_HAS_PMR 1
+#  endif
+#endif
+
+#if AUDIOAPI_HAS_PMR
+#  include <memory_resource>
+#else
+#  include <new>
+#endif
+
 namespace audioapi {
+
+#if !AUDIOAPI_HAS_PMR
+namespace bpq_detail {
+
+/// @brief Fixed-capacity, in-object block pool with an intrusive free list.
+/// Hands out fixed-size slots and recycles freed ones; never touches the heap.
+/// Throws std::bad_alloc on exhaustion, matching the std::pmr path's
+/// null_memory_resource upstream behaviour.
+template <std::size_t Capacity, std::size_t BlockSize, std::size_t Align>
+class FixedBlockPool {
+ public:
+  FixedBlockPool() = default;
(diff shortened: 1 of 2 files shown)


mdydek (approved):

sleekLancelot:
Added this as a patch and it fixed the crash on iOS 16 devices, thanks!

planfit-lucas:
Added this as a patch and it fixed the crash on iOS 16 devices, thanks!

If possible, we would appreciate it if you could release a version including this patch within a few weeks.

PR #1089, opened 2026-06-05
fix(android): stop resurrecting stopped recorders on stream disconnect

### LINKED ISSUE #1088: Android: stopped AudioRecorder is resurrected by onErrorAfterClose on audio route change (mic stays hot; SIGSEGV in oboe fireDataCallback on 0.12.x)
## Summary

`AndroidAudioRecorder` has a set of related lifecycle bugs around oboe's `ErrorDisconnected` handling that (a) turn the microphone back on after `stop()` whenever the audio route changes, and (b) on 0.12.x crash the app with a native SIGSEGV in `oboe::AudioStream::fireDataCallback`.

## Environment

- react-native-audio-api **0.12.2** (crash reproduced in production); the same logic bugs are still present on current `main`
- Android, arm64-v8a

## Production crash (Crashlytics tombstone, 0.12.2)

```
#00 pc 0x2e9d0  oboe::AudioStream::fireDataCallback(void*, int)  [liboboe.so]
#01 pc 0x2c6d8  oboe::AudioStreamAAudio::callOnAudioReady(AAudioStreamStruct*, void*, int)  [liboboe.so]
#02 pc 0x48b90  aaudio::AudioStream::maybeCallDataCallback(void*, int)  [libaaudio_internal.so]
#03 pc 0x55090  aaudio::AudioStreamLegacy::callDataCallbackFrames(unsigned char*, int)
#04 pc 0x545b4  aaudio::AudioStreamLegacy::onMoreData(android::AudioRecord::Buffer const&)
#05 pc 0x6d9a4  android::AudioRecord::processAudioBuffer()  [libaudioclient.so]
#06 pc 0x6d0bc  android::AudioRecord::AudioRecordThread::threadLoop()
```

## Root causes (all in `AndroidAudioRecorder.cpp`)

1. **`stop()` never closes the stream** — it only calls `mStream_->requestStop()`. The stream stays open with its data/error callbacks registered indefinitely.

2. **`onErrorAfterClose()` unconditionally restarts the stream** on `ErrorDisconnected`, with no lock and no state check. Any audio route change (Bluetooth headset connecting/disconnecting, speaker switch) on a *stopped* recorder reopens the stream and flips `state_` back to `Recording` — the mic silently turns back on after `stop()`. On `main`, where the stream holds a `shared_ptr` to the recorder (`shared_from_this()`), the resurrected stream also keeps the recorder alive forever after JS drops it (reference cycle: recorder → `mStream_` → recorder), so this is a privacy bug *and* a leak.

3. **Unsynchronized `mStream_` access across threads.** oboe's error callback runs on a *detached* thread (`AudioStreamAAudio::internalErrorCallback` → `std::thread(...).detach()`), and `onErrorAfterClose` → `cleanup()` resets/reopens `mStream_` there while `start()`/`stop()`/the destructor mutate it on the JS thread with no common lock. On 0.12.2, where the stream holds a *raw* pointer to the recorder (`setDataCallback(this)`), the interleaving

   - JS GC destroys the recorder host object → destructor reads `mStream_` just as the error thread's `cleanup()` reset it → skips the close → object freed
   - error thread then reopens a stream whose data callback points at the freed recorder

   produces exactly the `fireDataCallback` SIGSEGV above on the next mic buffer.

4. Because the error thread is detached it can also arrive **late**: if `stop()`/`start()` already replaced the stream, the handler closes the *healthy new* stream and reopens it without re-running `prepare()`.

5. `isRecording()` dereferences `mStream_` without a null check; the error thread can reset it concurrently (and `pause()` calls `mStream_->pause(0)` based on that check).

## Reproduction sketch

1. `recorder.onAudioReady(...)`; `recorder.start()`; `recorder.stop()`; drop the JS reference.
2. Connect or disconnect a Bluetooth audio device → mic recording restarts (verifiable via the mic-in-use indicator / `dumpsys audio`).
3. On 0.12.x, keep toggling the audio route while the GC collects the old recorder → eventual SIGSEGV in `oboe::AudioStream::fireDataCallback`.

## Suggested fix

- `stop()` fully closes and resets the stream (also breaks the `shared_from_this` reference cycle); `start()` already reopens on demand.
- `onErrorAfterClose()` takes the same locks as `start()`/`stop()`, bails when the recorder is idle, and bails when the erroring stream is no longer `mStream_`.
- Null-check `mStream_` in `isRecording()`; take the recorder locks in `pause()`/`resume()`.

PR with this fix incoming. We've been running it in production as a patch-package patch on 0.12.2 (compiled for arm64-v8a; lock ordering verified against oboe 1.9.3 — the data callback only ever `try_lock`s the recorder mutexes and error threads are detached, so closing under the locks cannot deadlock).

christian-apollo:
Fixes #1088

## Problem

On Android, a stopped `AudioRecorder` keeps its oboe input stream open (`stop()` only calls `requestStop()`), and `onErrorAfterClose()` unconditionally reopens and restarts the stream on `ErrorDisconnected`. Any audio route change — e.g. a Bluetooth headset connecting or disconnecting — therefore turns the microphone back on after `stop()`. Since the stream holds a `shared_ptr` to the recorder (`shared_from_this()`), the resurrected stream also keeps the recorder alive forever once JS drops it (reference cycle).

`onErrorAfterClose` additionally runs on oboe's *detached* error thread and mutates `mStream_` (via `cleanup()`/`openAudioStream()`) with no synchronization against `start()`/`stop()` on the JS thread. On 0.12.x, where the callbacks were raw pointers, this race produced a production SIGSEGV in `oboe::AudioStream::fireDataCallback` (full tombstone in #1088): the destructor lost the race, the error thread reopened a stream pointing at the freed recorder, and the next mic buffer crashed.

## Changes

- **`stop()`** fully closes and resets the stream, so a stopped recorder has no open stream — no callbacks can fire, no error-restart can resurrect it, and the recorder→stream→recorder reference cycle is broken. `start()` already reopens the stream on demand when `mStream_` is null.
- **`onErrorAfterClose()`**:
  - takes the same `callbackMutex_`/`fileWriterMutex_`/`adapterNodeMutex_` locks as `start()`/`stop()`, serializing all `mStream_` mutation;
  - bails out when the recorder is idle, so stopped recorders stay stopped;
  - bails out when the erroring stream is no longer `mStream_` — the detached error thread can arrive late, after `stop()`/`start()` already replaced the stream, and previously would tear down the healthy new stream (without re-running `prepare()`).
- **`isRecording()`** null-checks `mStream_`, which the error thread can reset concurrently; **`pause()`/`resume()`** take the recorder locks for the same reason (`resume()` also null-checks before `mStream_->start(0)`).

## Deadlock safety

Closing the stream while holding the recorder mutexes is safe:
- `onAudioReady` only ever uses `Locker::tryLock` on these mutexes, so the data callback never blocks on them and `close()`'s wait for in-flight callbacks is bounded.
- oboe error-callback threads are detached (`AudioStreamAAudio::internalErrorCallback`, oboe 1.9.3), so `close()` never joins a thread that might be blocked on these locks; a blocked error thread simply proceeds afterwards and bails on the idle/stale-stream checks.

## Testing

- Compiled for arm64-v8a and running in production as a patch-package patch on 0.12.2 (the same fix plus destructor locking, which `main` no longer needs thanks to `shared_from_this()`).
- Stopped-recorder + Bluetooth route-change no longer restarts the mic; repeated route changes during active recording recover as before via the (now guarded) reopen path.

### DIFF
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
@@ -182,13 +182,19 @@ AndroidAudioRecorder::stop() {
           "Recorder is not in recording state.");
     }
 
-    if (mStream_ == nullptr) {
-      return Result<std::tuple<std::vector<std::string>, double, double>, std::string>::Err(
-          "Audio stream is not initialized.");
-    }
-
     state_.store(RecorderState::Idle, std::memory_order_release);
-    mStream_->requestStop();
+
+    // Fully close the stream rather than just stopping it. A stopped-but-open
+    // stream keeps the data/error callbacks registered, so a later device
+    // disconnect fires onErrorAfterClose and resurrects the stream (the mic
+    // stays hot after stop()), and the stream's shared_ptr to this recorder
+    // keeps the recorder alive forever once JS drops it (reference cycle).
+    // start() reopens the stream when mStream_ is null.
+    if (mStream_ != nullptr) {
+      mStream_->requestStop();
+      mStream_->close();
+      mStream_.reset();
+    }
 
     hadFileOutput = usesFileOutput();
 
@@ -358,6 +364,8 @@ void AndroidAudioRecorder::disableFileOutput() {
 /// For session without active file output, this method acts same as stop().
 /// This method should be called from the JS thread only.
 void AndroidAudioRecorder::pause() {
+  std::scoped_lock pauseLock(callbackMutex_, fileWriterMutex_, adapterNodeMutex_);
+
   if (!isRecording()) {
     return;
   }
@@ -369,7 +377,9 @@ void AndroidAudioRecorder::pause() {
 /// @brief Resumes the audio recording stream if it was previously paused.
 /// This method should be called from the JS thread only.
 void AndroidAudioRecorder::resume() {
-  if (!isPaused()) {
+  std::scoped_lock resumeLock(callbackMutex_, fileWriterMutex_, adapterNodeMutex_);
+
+  if (!isPaused() || mStream_ == nullptr) {
     return;
   }
 
@@ -511,7 +521,7 @@ oboe::DataCallbackResult AndroidAudioRecorder::onAudioReady(
 
 bool AndroidAudioRecorder::isRecording() const {
   return state_.load(std::memory_order_acquire) == RecorderState::Recording &&
-      mStream_->getState() == oboe::StreamState::Started;
+      mStream_ != nullptr && mStream_->getState() == oboe::StreamState::Started;
 }
 
 bool AndroidAudioRecorder::isPaused() const {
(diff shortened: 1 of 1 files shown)


packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
```diff
@@ -537,29 +547,50 @@ void AndroidAudioRecorder::cleanup() {
...
 
-    if (!streamResult.is_ok()) {
-      uint64_t callbackId = errorCallbackId_.load(std::memory_order_acquire);
+  // Only restart if the recorder is still supposed to be running. Restarting
+  // unconditionally resurrected stopped recorders, so any audio route change
+  // (e.g. a Bluetooth headset connecting or disconnecting) turned the mic
+  // back on after stop().
+  if (isIdle()) {
+    return;
+  }
```
mdydek:
```suggestion
  if (isIdle() || isPaused()) {
    return;
  }
```

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
```diff
@@ -537,29 +547,50 @@ void AndroidAudioRecorder::cleanup() {
 /// @param oboeStream Pointer to the Oboe audio stream.
 /// @param error The oboe::Result error code.
 void AndroidAudioRecorder::onErrorAfterClose(oboe::AudioStream *stream, oboe::Result error) {
-  if (error == oboe::Result::ErrorDisconnected) {
-    cleanup();
+  if (error != oboe::Result::ErrorDisconnected) {
+    return;
+  }
```
mdydek, line 558: it would be good idea to also dispatch other `RECORDER_ERROR` events here

mdydek (changes requested):

mdydek (approved):

PR #1092, opened 2026-06-09
fix: better callbacks handling

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- improved callback registering and unregistering between js and audio thread
- setting callbacks objects are now done via atomic variables, not audio events which can be problematic when context is suspended

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioBufferSourceNode.cpp
```diff
@@ -155,15 +149,17 @@ void AudioBufferSourceNode::runBufferProcessor(
...
-      playbackState_ = PlaybackState::STOP_SCHEDULED;
-    }
+  if (processor_->didCrossLoopBoundary()) {
     sendOnLoopEndedEvent();
   }
 
+  if (processor_->atBoundary() && processor_->shouldStop()) {
+    playbackState_ = PlaybackState::STOP_SCHEDULED;
+  }
+
```
closetcaiman, line 161: This handling is inconsistent with `AudioBufferQueueSourceNode`.

Here the event dispatch is called outside of the processor while in `AudioBufferQueueSourceNode` a lambda is passed to be invoked.

It would be best to unify this.

packages/react-native-audio-api/common/cpp/audioapi/utils/events/PositionChangedDispatcher.h
```diff
@@ -0,0 +1,39 @@
...
+
+#include <atomic>
+#include <cstdint>
+#include <memory>
+
+namespace audioapi {
+
+/// @brief Wrapper around EventCaller that dispatches position changed events
+/// with interval throttling and optional flush-on-seek behavior.
+class PositionChangedDispatcher {
```
closetcaiman, line 13: If we are planning to add more dispatchers it would be good to add a base class for them. Just a thought.

packages/react-native-audio-api/common/cpp/audioapi/utils/events/PositionChangedDispatcher.cpp
```diff
@@ -0,0 +1,57 @@
...
+}
+
+void PositionChangedDispatcher::setIntervalInFrames(int intervalInFrames) {
+  intervalInFrames_ = std::max(0, intervalInFrames);
+}
+
+void PositionChangedDispatcher::setIntervalMs(int intervalInMs, float sampleRate) {
+  //NOLINTNEXTLINE(cppcoreguidelines-avoid-magic-numbers, readability-magic-numbers)
+  setIntervalInFrames(static_cast<int>(sampleRate * static_cast<float>(intervalInMs) / 1000.0f));
+}
```
closetcaiman, line 35: I'm not a fan of ignoring the linter here. Don't we have any methods that handle calculation between frames and {m}s? I believe we have a unit conversion `namespace` somewhere.

closetcaiman (changes requested):
Nice job reducing the scatter here!

packages/react-native-audio-api/common/cpp/audioapi/utils/events/PositionChangedDispatcher.h
mdydek: I think right now there is no plan to add more dispatchers, it was introduced only because of the presence of `positionChanged` event in few nodes unrelated to their base class. Also I don't know right now what would be the common interface for them, so I think it's better to leave it as it is and maybe do refactor when there will be need for another one

.claude/skills/thread-safety-itc/SKILL.md
```diff
@@ -65,37 +65,43 @@ oscillatorNode->scheduleAudioEvent(std::move(event));
...
-// Audio thread: fire event with payload
-audioEventHandlerRegistry_->invokeHandlerWithEventBody(
-    AudioEvent::POSITION_CHANGED, {{"position", currentPosition}});
+// With typed payload — compile error if payload does not match the event
+errorEvent_.dispatch(StringPayload{.name = "message", .reason = message});
+positionChangedEvent_.dispatch(DoubleValuePayload{.value = position});
 ```
 
-Internally calls `callInvoker_->invokeAsync()` — safe to call from the audio thread.
+Throttled events (e.g. `positionChanged`) use `PositionChangedDispatcher`, which **contains** an `EventCaller` plus interval/flush logic — still composition, not node inheritance from `EventCaller`.
```
maciejmakowski2003, line 89: I would create sth like ThrottledEventCaller for that purpose, that inherits from EventCaller

packages/react-native-audio-api/common/cpp/audioapi/events/AudioEventHandlerRegistry.cpp
maciejmakowski2003, line 1: add docstring with info about that these methods are js thread only

maciejmakowski2003 (changes requested):

closetcaiman (approved):

maciejmakowski2003 (approved):

PR #1097, opened 2026-06-11
fix(ios): push the frames AVAudioConverter actually produced in the recorder callback

### LINKED ISSUE #1096: iOS: recorder resampling path pushes estimated frame counts, injecting a stale sample (audible click) at render-block boundaries
### Description

When `AudioRecorder`'s `onAudioReady` callback is configured with a sample rate that differs from the hardware input rate (the common case — e.g. requesting 16 kHz for STT while the iOS input node runs at 48 kHz), the resampling path in `IOSRecorderCallback` pushes a **computed estimate** of the converter's output frame count into the circular buffer instead of the count AVAudioConverter actually produced:

https://github.com/software-mansion/react-native-audio-api/blob/936533dc/packages/react-native-audio-api/ios/audioapi/ios/core/utils/IOSRecorderCallback.mm#L205

```objc
size_t outputFrameCount = ceil(numFrames * (sampleRate_ / bufferFormat_.sampleRate));
...
[converter_ convertToBuffer:converterOutputBuffer_ error:&error withInputFromBlock:inputBlock];
converterOutputBuffer_.frameLength = sampleRate_ / bufferFormat_.sampleRate * numFrames;  // overwrites the converter's actual output count
...
circularBuffer_[i]->push_back(data, outputFrameCount);  // pushes the ceil() estimate
```

`convertToBuffer:error:withInputFromBlock:` sets `frameLength` to the number of frames it actually wrote, but that value is immediately overwritten with the estimate. The SRC's filter phase makes the real per-call output fluctuate around `numFrames * ratio` (e.g. a 512-frame block at 48 kHz → 16 kHz yields 170 *or* 171 frames depending on accumulated phase). Whenever the converter produces fewer frames than `ceil(...)`, the extra pushed samples are stale data from the end of `converterOutputBuffer_` — i.e. garbage from a previous render block.

### Result

A single-sample impulse (audible click) at the boundary of nearly every render block, in every recording where the requested callback rate differs from the hardware rate. Recordings sound "choppy"/crackly.

We hit this in production (voice-note recording at 16 kHz on devices with 48 kHz input). PCM inspection of the recorded output shows isolated single-sample spikes, e.g.:

```
..., -291, -251, 3689, -332, -538, ...     (≈ -30 dBFS signal, +3689 impulse)
..., -989, -935, 3657, -1014, -1285, ...
```

The spike offsets are exact multiples of `512 / 3 = 170.67` samples — one render block at the 3:1 conversion ratio — which pins the insertion point to the converter block boundary.

### Reproduction

1. `recorder.onAudioReady({ sampleRate: 16000, bufferLength: 1600, channelCount: 1 }, cb)` on an iOS device whose input runs at 48 kHz (any modern iPhone, built-in mic).
2. `recorder.start()`, capture a few seconds of speech, concatenate the chunks, write a WAV at 16 kHz.
3. Play it back / plot it: periodic single-sample impulses at block boundaries.

(Repros identically on the 44.1 kHz simulator, where the non-integer ratio makes the estimate wrong even more often.)

### Fix

Push `converterOutputBuffer_.frameLength` (the converter's actual output count) instead of the `ceil()` estimate, and don't overwrite `frameLength` after conversion. PR incoming.

### Environment

- react-native-audio-api: bug confirmed in production on 0.11.6; same code present on current `main` (936533dc)
- Platform: iOS (Android's `AndroidRecorderCallback` is unaffected — `ma_data_converter_process_pcm_frames` reports the produced frame count and that count is used)

yocontra:
Closes #1096

## ⚠️ Breaking changes ⚠️

None.

## Introduced changes

- `IOSRecorderCallback::taskOffloaderFunction` now pushes `converterOutputBuffer_.frameLength` — the frame count `AVAudioConverter` actually produced — into the circular buffer, instead of a `ceil(numFrames * ratio)` estimate.
- Removed the post-conversion `converterOutputBuffer_.frameLength = ...` assignment that was overwriting the converter's real output count with that estimate.
- Early-return when the converter produced zero frames.

### Why

The SRC's filter phase makes per-call output fluctuate around `numFrames * ratio` (a 512-frame block at 48 kHz → 16 kHz yields 170 or 171 frames depending on accumulated phase). Whenever the converter produced fewer frames than the `ceil()` estimate, the extra pushed samples were stale data from a previous render block — a single-sample impulse (audible click) at nearly every block boundary, in every recording where the requested `onAudioReady` sample rate differs from the hardware input rate. Full analysis with PCM evidence in #1096.

Android is unaffected: `AndroidRecorderCallback` already uses the produced count reported by `ma_data_converter_process_pcm_frames`.

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation — n/a, no API change
- [x] Added/Conducted relevant tests — verified in a production app: this fix (applied as a local patch) eliminates the periodic impulses from recorded PCM; before/after waveform analysis in #1096
- [ ] Updated Web Audio API coverage — n/a
- [ ] Added support for web — n/a, iOS-only bug fix
- [ ] Updated old arch android spec file — n/a

### DIFF
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/utils/AndroidRotatingFileWriter.cpp
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/utils/AndroidRotatingFileWriter.cpp
@@ -55,12 +55,14 @@ void AndroidRotatingFileWriter::writeAudioData(AudioDataType data, int numFrames
       rotateFiles();
     }
   }
-  framesWritten_.fetch_add(numFrames, std::memory_order_relaxed);
 }
 
 double AndroidRotatingFileWriter::getCurrentDuration() const {
-  return static_cast<double>(framesWritten_.load(std::memory_order_relaxed)) /
-      fileProperties_->sampleRate;
+  double currentSegmentDuration = 0.0;
+  if (currentWriter_ != nullptr) {
+    currentSegmentDuration = currentWriter_->getCurrentDuration();
+  }
+  return cumulativeDurationSec_ + currentSegmentDuration;
 }
 
 void AndroidRotatingFileWriter::rotateFiles() {
--- a/packages/react-native-audio-api/ios/audioapi/ios/core/utils/IOSFileWriter.h
+++ b/packages/react-native-audio-api/ios/audioapi/ios/core/utils/IOSFileWriter.h
@@ -18,6 +18,7 @@ typedef struct objc_object AVAudioConverter;
 #endif // __OBJC__
 
 struct WriterData {
+  // Owned by the worker thread; freed in IOSFileWriter::taskOffloaderFunction.
   const AudioBufferList *audioBufferList;
   int numFrames;
 };
--- a/packages/react-native-audio-api/ios/audioapi/ios/core/utils/IOSFileWriter.mm
+++ b/packages/react-native-audio-api/ios/audioapi/ios/core/utils/IOSFileWriter.mm
@@ -4,11 +4,15 @@
 #include <audioapi/events/AudioEventHandlerRegistry.h>
 #include <audioapi/ios/core/utils/FileOptions.h>
 #include <audioapi/ios/core/utils/IOSFileWriter.h>
+#include <audioapi/ios/core/utils/OwnedAudioBufferList.h>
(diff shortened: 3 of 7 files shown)


mdydek:
thank you for pointing out the issue

mdydek (approved):

PR #1107, opened 2026-06-12
feat: enable low latency audio thread on android

### LINKED ISSUE #1105: Android: Oboe streams open with PerformanceMode::None — consider LowLatency (or making it configurable)
## Description

Both the output and input Oboe streams are opened with `PerformanceMode::None`:

```cpp
// android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp (0.12.2)
builder.setSharingMode(SharingMode::Exclusive)
    ->setFormat(AudioFormat::Float)
    ->setFormatConversionAllowed(true)
    ->setPerformanceMode(PerformanceMode::None)
```

```cpp
// android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp (0.12.2)
builder.setSharingMode(oboe::SharingMode::Exclusive)
    ...
    ->setPerformanceMode(oboe::PerformanceMode::None)
```

[Oboe's own guidance](https://github.com/google/oboe/blob/main/docs/FullGuide.md#setting-performance-mode) is that `PerformanceMode::LowLatency` is required to get the AAudio fast path (FAST track / MMAP) — without it, Android adds buffering that `None` is explicitly allowed to take. This affects output latency, tap-to-sound, and record/monitor round-trips on every Android device.

## Why it matters

For interactive audio (instruments, DAW-style overdub recording, games) the difference between the fast path and `None` is routinely tens of milliseconds of output latency on mid-range hardware. I measure a ~270 ms gap between `recorder.start()` and the first delivered input buffer on a mid-range Android device at 0.12.x; the input stream's `PerformanceMode::None` is one of the contributors the app can't work around.

## Proposal

Either:

1. Default both streams to `PerformanceMode::LowLatency` (Oboe falls back gracefully when the fast path is unavailable, e.g. unsupported sample rates — and `setFormatConversionAllowed(true)` + `SampleRateConversionQuality` already cover the conversion cases), or
2. Keep `None` as default and expose it as an option (`new AudioContext({ androidPerformanceMode: 'lowLatency' })` and the recorder's `onAudioReady` options), so latency-sensitive apps can opt in without forking.

Given `SharingMode::Exclusive` is already requested (which is itself a low-latency posture), defaulting to `LowLatency` seems consistent with the library's intent.

## Environment

react-native-audio-api 0.12.2 (also checked `main`), Android, new architecture.

mdydek:
Closes #1105 

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
@@ -73,7 +73,7 @@ Result<NoneType, std::string> AndroidAudioRecorder::openAudioStream() {
       ->setDirection(oboe::Direction::Input)
       ->setFormat(oboe::AudioFormat::Float)
       ->setFormatConversionAllowed(true)
-      ->setPerformanceMode(oboe::PerformanceMode::None)
+      ->setPerformanceMode(oboe::PerformanceMode::LowLatency)
       ->setSampleRateConversionQuality(oboe::SampleRateConversionQuality::Medium)
       ->setDataCallback(shared_from_this())
       ->setErrorCallback(shared_from_this());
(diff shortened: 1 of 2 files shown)


maciejmakowski2003 (commented):
have you tested android perf somehow?

PR #1118, opened 2026-06-22
Fix/engine race condition

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- this pr aims to eliminate race condition in the audio native engines. For example one can happen when `audiocontext.resume()` and `node.start()` are called. Resume runs on separate thread pool which handles promises and starting node is done synchronously on js thread. Both of them would try to call inside them functions of engine, which is obvious race condition between two threads.

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


closetcaiman (approved):
lgtm

PR #1122, opened 2026-06-25
feat: add duration-only audio file API

Traviskn:
## Summary

  Adds a `getAudioDuration` utility for reading an audio file’s duration without decoding the full file into an `AudioBuffer`.

  This exposes the API in two places:

  - `getAudioDuration(input)` as a top-level utility
  - `BaseAudioContext#getAudioDuration(input)`

  The native implementation uses a duration-only decoder path so callers can inspect local audio files without allocating memory for the full PCM payload.

  ## Motivation

  Some apps need to validate or display recording durations before playback or upload. Using `decodeAudioData` for that can be expensive because it decodes the entire file into memory, which can cause memory pressure or crashes for longer recordings. In my case, lower end devices would crash with an out-of-memory error when using `decodeAudioData` to get the duration of the resulting audio buffer.

  `getAudioDuration` provides a lighter-weight path for this use case.
  
  I have applied this new API to my own app as a patch and have confirmed that it works.
 
## Checklist

- [ ] Linked relevant issue N/A
- [ ] Updated Web Audio API coverage N/A
- [ ] Added support for web N/A
- [ ] Updated old arch android spec file N/A


mdydek (changes requested):
Thank you for the contribution. Two things I would like to address: 
- I think that it should not be a method of a context, I would like to leave core web audio api classes as close to their web equivalent (there is no property like that)
- web is probably much more flexible when it comes to available resources, I would like to see implementation also on this side (either using audio tag or audiocontext)

Traviskn:
Thanks for the feedback! I’ll update this PR 👍

maciejmakowski2003 (approved):

maciejmakowski2003:
@copilot resolve the merge conflicts in this pull request

PR #1123, opened 2026-06-26
refactor: promises and errors

### LINKED ISSUE #1116: iOS: NSError details from setCategory are dropped, JS sees "unknown error" with empty userInfo
## Summary

When `configureAudioSession` fails (most commonly because `setCategory:mode:options:error:` rejects — see #1115 for one such case on iOS 26), the resulting `NSError` is logged via NSLog but **not propagated up the call chain**. The JS bridge ends up rejecting with `code: 0`, `domain: ""`, `description: ""`, so the JS-side `SessionActivationError` carries no diagnostic info — neither `userInfo.meta`, `nativeError`, nor `details` has anything useful.

This made diagnosing #1115 take several hours of native log capture and library source reading — the actual `AVFoundationErrorDomain` code from iOS was never visible to the app developer.

## Versions

- `react-native-audio-api`: 0.12.2 and 0.13.0-nightly-1a9dfbd-20260618 (both)
- iOS 26.5.1, but the bug is platform-independent — any setCategory rejection on any iOS version reproduces the same symptom.

## Root cause (line-by-line)

In `ios/audioapi/ios/system/AudioSessionManager.mm`:

```objc
- (bool)configureAudioSession
{
  NSError *error = nil;                              // ← LOCAL to this method

  [self.audioSession setCategory:self.desiredCategory
                            mode:self.desiredMode
                         options:self.desiredOptions
                           error:&error];

  if (error != nil) {
    NSLog(@"Error while configuring audio session: %@", [error debugDescription]);
    return false;                                    // ← error not propagated
  }
  ...
}

- (bool)activateSessionIfNeeded:(bool)force error:(NSError **)error
{
  ...
  if (![self configureAudioSession]) {
    return false;                                    // ← outer `error` stays nil
  }
  bool success = [self.audioSession setActive:true withOptions:0 error:error];
  ...
}
```

Then in `AudioAPIModule.mm`'s `setAudioSessionActivity` bridge:

```objc
NSError *error = nil;
auto success = [self.audioSessionManager setActive:enabled error:&error];

if (!success) {
  NSDictionary *meta = @{
    @"nativeCode" : @(error.code),                   // ← 0 (nil error)
    @"nativeDomain" : error.domain ?: @"",           // ← ""
    @"nativeDesc" : error.description ?: @"",       // ← ""
  };
  ...
}
```

Because `configureAudioSession` ate the NSError, the bridge sees `success=false` paired with a nil/empty error object. JS gets `SessionActivationError` with no meta.

## What JS sees

```
Parsed error meta: {"nativeCode": 0, "nativeDesc": "", "nativeDomain": ""}
[SessionActivationError: Failed to activate audio session with unknown error]
```

vs. what JS *should* see if the NSError were forwarded (something like):

```
Parsed error meta: {
  nativeCode: -50,
  nativeDomain: "AVFoundationErrorDomain",
  nativeDesc: "The operation couldn't be completed. (...)",
}
```

## Suggested fix

Make `configureAudioSession` accept an `NSError**` parameter and write the failure into it. Or refactor so `activateSessionIfNeeded` calls `setCategory` itself and the outer `error` pointer it received from the bridge is the one that gets written.

Sketch (option A — minimal change, passes the same error pointer through):

```objc
- (bool)configureAudioSession:(NSError **)outError
{
  if (!self.shouldManageSession || [self areDesiredOptionsSet]) {
    return true;
  }

  NSError *configError = nil;
  [self.audioSession setCategory:self.desiredCategory
                            mode:self.desiredMode
                         options:self.desiredOptions
                           error:&configError];

  if (configError != nil) {
    NSLog(@"Error while configuring audio session: %@", [configError debugDescription]);
    if (outError) *outError = configError;
    return false;
  }
  ...
}

// caller:
- (bool)activateSessionIfNeeded:(bool)force error:(NSError **)error
{
  ...
  if (![self configureAudioSession:error]) {
    return false;
  }
  ...
}
```

Same idea for the `setAllowHapticsAndSystemSoundsDuringRecording` step inside `configureAudioSession` — it currently overwrites the local `error` and the original is lost when both fail in sequence.

## Why this matters

Without it, **any** native AVAudioSession failure surfaces in JS as the same opaque "unknown error". Pairing it with #1115 means an end-user-visible silence bug with zero JS-side signal. Even after the iOS 26 setCategory rejection is patched, library users will hit other AVAudioSession edge cases (other apps holding exclusive session, particular Bluetooth devices, etc.) and need the real iOS error code to act on it.

Happy to PR the change if useful.

mdydek:
Closes #1116 

## ⚠️ Breaking changes ⚠️

- context's methods `resume` and `suspend` now returns `Promise<void>` instead of `Promise<boolean`
- `setAudioSessionActivity` now returns `Promise<void>` instead of `Promise<boolean`
- `setInputDevice` now returns `Promise<void>` instead of `Promise<boolean`

## Introduced changes

- replaced returning value of context method to match web spec
- replaced returning vales of some audio manager methods
- improved ios error logging

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/apps/common-app/src/demos/Record/Record.tsx
+++ b/apps/common-app/src/demos/Record/Record.tsx
@@ -63,7 +63,6 @@ const Record: FC = () => {
       setHasPermissions(true);
     }
 
-    let success = false;
     AudioManager.setAudioSessionOptions({
       iosCategory: 'playAndRecord',
       iosMode: 'default',
@@ -71,15 +70,9 @@ const Record: FC = () => {
     });
 
     try {
-      success = await AudioManager.setAudioSessionActivity(true);
+      await AudioManager.setAudioSessionActivity(true);
     } catch (error) {
       console.error(error);
-      Alert.alert('Error', 'Failed to activate audio session for recording.');
-      setState(RecordingState.Idle);
-      return;
-    }
-
-    if (!success) {
       Alert.alert('Error', 'Failed to activate audio session for recording.');
       setState(RecordingState.Idle);
       return;
--- a/apps/common-app/src/examples/AudioFile/AudioFile.tsx
+++ b/apps/common-app/src/examples/AudioFile/AudioFile.tsx
@@ -35,9 +35,10 @@ const AudioFile: FC = () => {
         iosOptions: [],
       });
 
-      const success = await AudioManager.setAudioSessionActivity(true);
-
-      if (!success) {
+      try {
+        await AudioManager.setAudioSessionActivity(true);
+      } catch (error) {
+        console.error('Failed to activate audio session:', error);
         Alert.alert(
           'Audio Session Error',
           'Failed to activate audio session for playback.'
--- a/apps/common-app/src/examples/Record/Record.tsx
+++ b/apps/common-app/src/examples/Record/Record.tsx
@@ -49,9 +49,10 @@ const Record: FC = () => {
       iosOptions: ['allowBluetoothHFP'],
     });
 
-    const success = await AudioManager.setAudioSessionActivity(true);
-
-    if (!success) {
+    try {
+      await AudioManager.setAudioSessionActivity(true);
+    } catch (error) {
+      console.error('Failed to activate audio session:', error);
       Alert.alert(
         'Audio Session Error',
         'Failed to activate audio session for recording.'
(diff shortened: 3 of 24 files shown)


closetcaiman (approved):
lgtm!

PR #1129, opened 2026-06-29
feat: audio latency

### LINKED ISSUE #1104: Expose AudioContext.baseLatency / outputLatency (Web Audio spec) for latency compensation
## Description

The Web Audio API specifies [`AudioContext.baseLatency`](<https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/baseLatency>) and [`AudioContext.outputLatency`](<https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/outputLatency>). Neither is exposed by react-native-audio-api at 0.12.2 — not natively, and the web wrapper doesn't pass the browser's values through either (no `baseLatency`/`outputLatency` anywhere in the published types).

## Why it matters

Any app that records against playing audio (overdub recording, call-and-response, rhythm games) needs **recording latency compensation**: a take performed in time lands `outputLatency + inputLatency` late on the timeline unless the app shifts it back. Right now there is no API surface to get those numbers, so apps must ship their own native module or a loopback-calibration flow. For scale: on an iPhone 16 speaker route (playAndRecord), I measure a \~188 ms end-to-end record round-trip through the library — all of it invisible to JS today.

The platform data already exists:

* **iOS**: `AVAudioSession.{outputLatency, inputLatency, ioBufferDuration}` map directly (`outputLatency` → spec `outputLatency`; `ioBufferDuration`-derived term → `baseLatency`).
* **Android**: the library owns the Oboe streams — `AudioStream::calculateLatencyMillis()` and stream timestamps are available on the open stream.
* **Web**: pass through the browser's own properties.

## Proposal

1. `context.baseLatency` / `context.outputLatency` per spec (getters, snapshot per call).
2. Bonus for the recording side: an equivalent input figure (e.g. `AudioRecorder.inputLatency` or a field on the `onAudioReady` event) — the spec has nothing here, but `AVAudioSession.inputLatency` and the Oboe input stream both report it.

Together with software-mansion/react-native-audio-api#1103 (populate the recorder `when`), this would make sample-accurate overdub placement possible without any app-side native code.

## Environment

react-native-audio-api 0.12.2, iOS + Android + web, new architecture.

closetcaiman:
Closes #1104 

## ⚠️ Breaking changes ⚠️

- N/A

## Introduced changes

> [!IMPORTANT]
> Identified race condition between `oboe` [background thread](https://github.com/google/oboe/blob/main/src/aaudio/AudioStreamAAudio.cpp#L282) (invoking `onErrorAfterClose()`) and any other thread that accesses `mStream_` -> serialized it via locks (`recursive_mutex` exactly for clearer locking flow)
> Audio Thread running `onAudioReady` does not require locks and thus stays consistent with no-locking policy

- aligned `AudioContext` interface with spec by adding `baseLatency` and `outputLatency`
- added `inputLatency` to `AudioRecorder` (React Native extension, not in spec)
- all values are sourced from native APIs - `AVAudioSession` (iOS) and Oboe (Android)
  - platform differences in reported values are expected
- updated docs for `AudioContext` and `AudioRecorder` with native source details

## Checklist

- [x] Linked relevant issue
- [x] Added support for web


packages/audiodocs/docs/inputs/audio-recorder.mdx
```diff
@@ -568,6 +568,26 @@ export default MyRecorder;
...
+      <td><div class="text-content">
+        ##### inputLatency <MobileOnly />
+      </div></td>
+      <td>
+        <div class="text-content">
+          **React Native extension (not in the Web Audio spec).** Returns the estimated **full** input latency in seconds — from the analog input through the platform to when samples are delivered to the app. Mirrors [`AudioContext.outputLatency`](/docs/core/audio-context#properties) semantics for capture. On iOS: `callbackFrames / sampleRate` (or `AVAudioSession.IOBufferDuration` before the first callback) plus `AVAudioSession.inputLatency`. On Android: `max(Oboe timestamp measurement, callbackFrames / sampleRate)` when Oboe timestamps are available; falls back to `callbackFrames / sampleRate + framesPerBurst / sampleRate`. Re-query before sync-sensitive work; the value can change when the audio route changes. Returns `0` when the recorder is idle or when recording from a graph tap without an open input stream.
+
+          Use together with [`AudioContext.outputLatency`](/docs/core/audio-context#properties) to compensate overdub placement:
+
+          ```tsx
```
maciejmakowski2003:
```suggestion
          **Mobile only extension (not in the Web Audio spec).** Returns the estimated **full** input latency in seconds — from the analog input through the platform to when samples are delivered to the app. Mirrors [`AudioContext.outputLatency`](/docs/core/audio-context#properties) semantics for capture. On iOS: `callbackFrames / sampleRate` (or `AVAudioSession.IOBufferDuration` before the first callback) plus `AVAudioSession.inputLatency`. On Android: `max(Oboe timestamp measurement, callbackFrames / sampleRate)` when Oboe timestamps are available; falls back to `callbackFrames / sampleRate + framesPerBurst / sampleRate`. Re-query before sync-sensitive work; the value can change when the audio route changes. Returns `0` when the recorder is idle or when recording from a graph tap without an open input stream.

          Use together with [`AudioContext.outputLatency`](/docs/core/audio-context#properties) to compensate overdub placement:

          ```tsx
```

packages/react-native-audio-api/common/cpp/audioapi/core/BaseAudioContext.h
```diff
@@ -39,6 +39,7 @@ class BaseAudioContext : public std::enable_shared_from_this<BaseAudioContext> {
   }
   [[nodiscard]] float getSampleRate() const;
   [[nodiscard]] double getCurrentTime() const;
+
```
maciejmakowski2003: ?

packages/react-native-audio-api/src/core/AudioRecorder.ts
```diff
@@ -186,6 +186,10 @@ export default class AudioRecorder {
     return this.recorder.getCurrentDuration();
   }
 
+  get inputLatency(): number {
+    return this.recorder.getInputLatency();
+  }
```
maciejmakowski2003, line 216: why dont we declare it as a readonly property like in `AudioContext`?

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioContextHostObject.h
```diff
@@ -23,5 +23,7 @@ class AudioContextHostObject : public BaseAudioContextHostObject {
   JSI_HOST_FUNCTION_DECL(resume);
   JSI_HOST_FUNCTION_DECL(suspend);
   JSI_HOST_FUNCTION_DECL(createMediaElementSource);
+
+  JSI_PROPERTY_GETTER_DECL(outputLatency);
```
maciejmakowski2003, line 27: where is getter for baseLatency?

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
```diff
@@ -473,6 +476,16 @@ oboe::DataCallbackResult AndroidAudioRecorder::onAudioReady(
...
 
+  if (numFrames > 0) {
+    lastCallbackFrameCount_.store(numFrames, std::memory_order_release);
+
+    // Sample at callback start (maximum of the input latency sawtooth).
+    const auto latencyResult = oboeStream->calculateLatencyMillis();
+    if (latencyResult) {
+      lastInputLatencySeconds_.store(latencyResult.value() / 1000.0, std::memory_order_release);
+    }
+  }
```
maciejmakowski2003: have you check the cost of `oboeStream->calculateLatencyMillis()` invocation?

maciejmakowski2003 (changes requested):

packages/audiodocs/docs/inputs/audio-recorder.mdx
```diff
@@ -365,8 +365,9 @@ const audioRecorder = new AudioRecorder();
 ## Properties
 
 | Name | Type | Description |
-| :----: | :----: | :-------- |
+| :--- | :--- | :--- |
 | `options` | [`AudioRecorderFileOptions`](#audiorecorderfileoptions) \| `null` | Active file-output configuration after [`enableFileOutput()`](#enablefileoutput), or `null` when file output is disabled. |
+| `inputLatency` <MobileOnly/> | `number` | Returns the estimated **full** input latency in seconds (from analog input to app delivery). Mirrors [`AudioContext.outputLatency`](/docs/core/audio-context#properties) semantics. Calculated using underlying platform properties (`AVAudioSession` on iOS; Oboe timestamps/burst sizes on Android). Value may change when the audio route changes. Returns `0` when idle. |
```
maciejmakowski2003: I am pretty sure that `AudioRecorder` is mobile only so it is no need to mark that prop as mobile only.

maciejmakowski2003 (changes requested):

packages/audiodocs/docs/core/audio-context.mdx
```diff
@@ -29,6 +30,11 @@ constructor(options?: AudioContextOptions)
 
 ## Properties
 
+| Name | Type | Description | |
+| :----: | :----: | :-------- | :-: |
+| `baseLatency` | `number` | Processing latency in seconds incurred by the `AudioContext` passing audio from the `AudioDestinationNode` to the audio subsystem. Does not include hardware output latency or audio graph processing time. Equals the render callback duration (`callbackFrames / sampleRate`). Before the first callback fires, falls back to `AVAudioSession.IOBufferDuration` (iOS) or Oboe `framesPerBurst / sampleRate` (Android). Returns `0` when the context is not running. | <ReadOnly /> |
```
mdydek:
```suggestion
| `baseLatency` | `number` | Processing latency in seconds incurred by the `AudioContext` passing audio from the `AudioDestinationNode` to the audio subsystem. Does not include hardware output latency or audio graph processing time. Equals the render callback duration (`callbackFrames / sampleRate`). Returns `0` when the context is not running. | <ReadOnly /> |
```
looking at code it does not behave that way

packages/audiodocs/docs/core/audio-context.mdx
```diff
@@ -29,6 +30,11 @@ constructor(options?: AudioContextOptions)
...
+| `outputLatency` | `number` | Estimated output latency in seconds — the interval between when the audio system is given a buffer and when the first sample is played by the output device. Can change when the audio route changes; re-query before sync-sensitive work. Returns `0` when the context is not running. On iOS backed by `AVAudioSession.outputLatency`; on Android sampled via Oboe timestamps. | <ReadOnly /> |
+
 `AudioContext` does not define any additional properties.
```
mdydek: should be deleted right now since it has some

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/utils/latency/LatencyMonitor.cpp
```diff
@@ -0,0 +1,54 @@
+
+#include <audioapi/android/core/utils/latency/LatencyMonitor.h>
+#include <pthread.h>
```
mdydek: do we need this header, seems very odd if you include thread header

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/utils/latency/LatencyMonitor.h
```diff
@@ -0,0 +1,37 @@
...
+#include <atomic>
+#include <chrono>
+#include <optional>
+#include <thread>
+
+namespace audioapi {
+
+inline constexpr auto LATENCY_MONITOR_SLEEP_DURATION_MS = std::chrono::milliseconds(10);
+inline constexpr auto LATENCY_MONITOR_SMOOTHING_FACTOR = 0.1;
+inline constexpr auto LATENCY_MONITOR_CALLBACK_SIZE = 32;
```
mdydek:
```suggestion
inline constexpr auto LATENCY_MONITOR_CALLBACK_SIZE = sizeof(std::size_t);
```
you are capturing only size of one ptr

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/utils/latency/LatencyMonitor.h
```diff
@@ -0,0 +1,37 @@
...
+
+using LatencyCallback = FatFunction<LATENCY_MONITOR_CALLBACK_SIZE, std::optional<double>()>;
+
+class LatencyMonitor {
```
mdydek: define special member functions

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
```diff
@@ -473,6 +481,10 @@ oboe::DataCallbackResult AndroidAudioRecorder::onAudioReady(
     return oboe::DataCallbackResult::Continue;
   }
 
+  if (numFrames > 0) {
+    lastCallbackFrameCount_.store(numFrames, std::memory_order_release);
```
mdydek, line 476: i don't see any load of this variable on android side

mdydek: is it needed?

mdydek (changes requested):

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/utils/latency/LatencyMonitor.cpp
closetcaiman: It's a leftover from profiling, needed for thread naming. Removed.

packages/react-native-audio-api/ios/audioapi/ios/core/IOSAudioPlayer.h
```diff
@@ -34,6 +34,9 @@ class IOSAudioPlayer : public CommonPlayer {
 
   [[nodiscard]] bool isRunning() const override;
 
+  [[nodiscard]] double getBaseLatency() const;
+  [[nodiscard]] double getOutputLatency() const;
```
mdydek:
```suggestion
  [[nodiscard]] double getBaseLatency() const override;
  [[nodiscard]] double getOutputLatency() const override;
```

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
```diff
@@ -566,4 +581,36 @@ void AndroidAudioRecorder::onErrorAfterClose(oboe::AudioStream *stream, oboe::Re
...
+  double baseLatency = 0.0;
+  const int32_t callbackFrames = lastCallbackFrameCount_.load(std::memory_order_acquire);
+  if (callbackFrames > 0 && streamSampleRate_ > 0.0f) {
+    baseLatency = static_cast<double>(callbackFrames) / static_cast<double>(streamSampleRate_);
+  } else {
+    const int32_t framesPerBurst = mStream_->getFramesPerBurst();
+    if (framesPerBurst > 0 && streamSampleRate_ > 0.0f) {
+      baseLatency = static_cast<double>(framesPerBurst) / static_cast<double>(streamSampleRate_);
+    }
+  }
```
mdydek: i think we should invert this logic, trust oboe first and if it does not return any values, use our estimate, btw is there a time when it will not give right value?

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
```diff
@@ -566,4 +581,36 @@ void AndroidAudioRecorder::onErrorAfterClose(oboe::AudioStream *stream, oboe::Re
...
+    return std::max(latencyResult.value() / 1000.0, baseLatency);
+  }
+
+  const int32_t framesPerBurst = mStream_->getFramesPerBurst();
+  if (framesPerBurst > 0 && streamSampleRate_ > 0.0f) {
+    return baseLatency +
+        static_cast<double>(framesPerBurst) / static_cast<double>(streamSampleRate_);
+  }
+
+  return baseLatency;
```
mdydek: in every return branch you are returning different values:
- max from baselatency and latency of oboe
- baselatency + estimate of inputlatency
- just baselatency
It does not make any sense to me

mdydek (changes requested):

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
closetcaiman: Agreed.

On startup or on deviced with buggy drivers.

mdydek (approved):

PR #1134, opened 2026-06-30
feat: remote url playback

maciejmakowski2003:
Closes #

## ⚠️ Breaking changes ⚠️

- 

## Introduced changes

- Introduced support for remote URL playback. The system first checks if the server supports HTTPS range requests to stream the audio, falling back to a full download if it does not.
-Rework `probeDuration` and `getAudioDuration`, an FFmpeg-based approach to duration probing.

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/src/Audio/useAudioSourceLoader.ts
```diff
@@ -150,14 +157,12 @@ export function useAudioSourceLoader({
...
     const headers = getSourceHeaders(source);
 
     try {
-      if (path.startsWith('http') && !path.endsWith('.m3u8')) {
-        const arrayBuffer = await fetch(path, { headers }).then((response) =>
-          response.arrayBuffer()
-        );
-        sourceRef.current = arrayBuffer;
+      if (isRemoteHttpUrl(path)) {
+        sourceRef.current = await loadRemoteHttpSource(path, headers);
```
mdydek: this makes us more ffmpeg dependent, because in previous implementation we were using miniaudio to decode urls of .mp3,  right now it will try to use ffmpeg all the times, so its clearly a regression

mdydek: probably should have fallback for the miniaudio

mdydek (changes requested):

packages/react-native-audio-api/src/Audio/useAudioSourceLoader.ts
```diff
@@ -199,14 +211,21 @@ export function useAudioSourceLoader({
...
+
+      const probedDuration = useSegmentProbe
+        ? await probeDuration(
+            path,
+            DEFAULT_METADATA_SEGMENT_BYTES,
+            DEFAULT_METADATA_SEGMENT_BYTES,
+            context?.sampleRate,
+            headers
+          )
+        : await probeDurationFromUrl(path, context?.sampleRate, headers);
```
mdydek: why can't we probe duration from url without byte ranges header?

mdydek: maybe we can discard the probeDuration function and in no ffmpeg build in audio tag just treat preload metadata as none, because I think the first function is not that reliable

packages/react-native-audio-api/src/utils/remoteHttpSource.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (headResponse.ok) {
+      const acceptRanges = headResponse.headers
+        .get('Accept-Ranges')
+        ?.toLowerCase();
+      if (acceptRanges === 'bytes') {
+        return true;
+      }
+      if (acceptRanges === 'none') {
+        return false;
+      }
```
mdydek:
```suggestion
      return false;
```

mdydek (changes requested):

mdydek (approved):

PR #1150, opened 2026-07-06
Refactor/prebuilt binaries downloading

### LINKED ISSUE #1140: iOS: cached `Pods/` + fresh `node_modules` skips `prepare_command`, prebuilt binaries missing at link time
### Repro: cached `ios/Pods` + fresh `node_modules` skips `prepare_command`

We hit the same "Build input files cannot be found" failure for<br>`libogg.a`, `libopus.a`, `libopusfile.a`, `libvorbis.a`,<br>`libvorbisenc.a`, `libvorbisfile.a` under<br>`node_modules/react-native-audio-api/common/cpp/audioapi/external/iphoneos/`.

**Setup**

* react-native-audio-api:
* React Native: 0.84.1
* CocoaPods:
* CI: macOS runner, `ios/Pods` restored from cache keyed on<br>`Podfile.lock`, `node_modules` freshly installed (`yarn install`<br>from cache but the folder itself is not persisted between jobs).

**What happens**

1. `yarn install` recreates `node_modules/react-native-audio-api/`<br>from the npm tarball. The tarball does **not** contain the<br>prebuilt static libs under `common/cpp/audioapi/external/`.
2. `pod install` restores `ios/Pods` from cache. Because the<br>`Podfile.lock` hash is unchanged, CocoaPods considers the pod<br>already installed and **does not re-run** `prepare_command`,<br>which is the only step that invokes<br>`scripts/download-prebuilt-binaries.sh`.
3. Xcode fails at link time: the podspec's `-force_load` entries<br>point at `external/iphoneos/lib*.a`, which were never downloaded.

This is not specific to any one CI provider - any workflow that<br>caches `Pods/` independently from `node_modules/` (a very common<br>optimisation to avoid the \~2-3 min `pod install` on every job)<br>will hit it. A local dev running `rm -rf node_modules && yarn`<br>without also nuking `ios/Pods` reproduces it too.

**Root cause (from our side)**

`prepare_command` is a one-shot install-time hook. It is a poor<br>fit for hydrating artifacts that live under `node_modules/` rather<br>than under `Pods/`, because `node_modules/` and `Pods/` have<br>independent lifecycles and independent caches. Any environment<br>where those two caches drift will leave the pod pointing at<br>non-existent files.

**Workaround we shipped**

Extended our root `package.json` postinstall to hydrate the<br>binaries whenever `node_modules` is (re)installed, gated to<br>Darwin so Linux jobs don't pay the download:

```json
"postinstall": "patch-package && ([ \"$(uname)\" = \"Darwin\" ] && (cd node_modules/react-native-audio-api && bash scripts/download-prebuilt-binaries.sh) || true)"
```

This makes the binaries follow `node_modules/` regardless of<br>whether `Pods/` is cached, and matches what the Android<br>`build.gradle` already does (invokes the same script during<br>Gradle configuration).

**Suggested upstream fix**

Move the hydration out of `prepare_command` so it runs whenever<br>`node_modules/react-native-audio-api/` is created, not only when<br>CocoaPods decides to re-run install. Two options:

1. **npm** `postinstall` **in the package's own** `package.json` -<br>guarantees the binaries are present immediately after<br>`yarn install` / `npm install`, independent of CocoaPods<br>caching. Downside: runs on non-Darwin too, so it would need<br>a `uname`/platform guard, or split scripts.
2. `script_phase` **in the podspec** that runs on every build<br>(or on every `pod install`, unconditionally) and no-ops if<br>the binaries are already present. This keeps the fix inside<br>the pod but bypasses CocoaPods' "already installed" short-<br>circuit.

Happy to open a PR for either approach if there's a preferred<br>direction.

**Related closed issues**

* [#867](<https://github.com/software-mansion/react-native-audio-api/issues/867>)
* software-mansion/react-native-audio-api#1048
* software-mansion/react-native-audio-api#1058 (Android/Windows variant)
* software-mansion/react-native-audio-api#1012 (Android/Windows variant)

All four describe symptoms of the same underlying model (binaries<br>hydrated by a one-shot install hook), so it may be worth<br>reopening one of them as a tracker rather than closing per<br>platform.

maciejmakowski2003:
Closes #1140 

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- Split downloading static libs prebuilt for android and iOS
- Aligned downloading on iOS with Android approach. All prebuilts are downloaded in build time or on `pod install` if cache is invalid.
- Downloading script checks presence of all artifacts not only dirs as it used to be
- Renamed jsi utils classes

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/RNAudioAPI.podspec
+++ b/packages/react-native-audio-api/RNAudioAPI.podspec
@@ -75,6 +75,7 @@ Pod::Spec.new do |s|
   ffmpeg_dir = "$(PODS_ROOT)/#{$audio_api_config[:dynamic_frameworks_audio_api_dir]}/#{external_dir_relative}/ffmpeg_ios"
 
   ffmpeg_framework_names =  %w[libavcodec libavformat libavutil libswresample]
+  static_external_libs_names = %w[libopusfile libopus libogg libvorbis libvorbisenc libvorbisfile]
 
   # `prepare_command` hydrates binaries on a fresh `pod install` so CocoaPods can
   # register all vendored `.xcframework`s in `[CP] Copy XCFrameworks`. When `Pods/`
@@ -93,7 +94,7 @@ Pod::Spec.new do |s|
     download_output_files = []
 
     unless $RN_AUDIO_API_STATIC_EXTERNAL_LIBS_DISABLED
-      %w[libopusfile libopus libogg libvorbis libvorbisenc libvorbisfile].each do |lib|
+      static_external_libs_names.each do |lib|
         download_output_files << "#{lib_dir}/#{lib}.a"
       end
     end
@@ -175,14 +176,9 @@ Pod::Spec.new do |s|
     'OTHER_LDFLAGS' => %W[
       $(inherited)
     ]
-    .concat($RN_AUDIO_API_STATIC_EXTERNAL_LIBS_DISABLED ? [] : %W[
-      -force_load #{lib_dir}/libopusfile.a
-      -force_load #{lib_dir}/libopus.a
-      -force_load #{lib_dir}/libogg.a
-      -force_load #{lib_dir}/libvorbis.a
-      -force_load #{lib_dir}/libvorbisenc.a
-      -force_load #{lib_dir}/libvorbisfile.a
-    ])
+    .concat($RN_AUDIO_API_STATIC_EXTERNAL_LIBS_DISABLED ? [] : static_external_libs_names.flat_map { |lib|
+      ['-force_load', "#{lib_dir}/#{lib}.a"]
+    })
     .join(" "),
   }
   # Use install_modules_dependencies helper to install the dependencies if React Native version >=0.71.0.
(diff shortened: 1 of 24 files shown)


packages/react-native-audio-api/RNAudioAPI.podspec
```diff
@@ -67,18 +67,59 @@ Pod::Spec.new do |s|
...
+  # would reject the build when the xcframeworks are missing before we download.
+  #
+  # `output_files` declares everything the linker (`-force_load` static libs in
+  # `s.xcconfig`) and Copy XCFrameworks (FFmpeg) consume, so Xcode's build graph
+  # knows this phase produces them.
+  unless $RN_AUDIO_API_STATIC_EXTERNAL_LIBS_DISABLED && $RN_AUDIO_API_FFMPEG_DISABLED
+    download_output_files = []
+
+    unless $RN_AUDIO_API_STATIC_EXTERNAL_LIBS_DISABLED
+      %w[libopusfile libopus libogg libvorbis libvorbisenc libvorbisfile].each do |lib|
```
mdydek: we can make this array into variable, so it can be reused here and later on when flag `-force_load` is used to link those libraries

mdydek (changes requested):

mdydek (approved):

PR #1151, opened 2026-07-06
feat: analyser and biquad wpt improvements

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/core/effects/BiquadFilterNode.cpp
```diff
@@ -117,218 +120,239 @@ void BiquadFilterNode::getFrequencyResponse(
...
       continue;
     }
 
-    double omega = -PI * normalizedFreq;
-    auto z = std::complex<double>(std::cos(omega), std::sin(omega));
-    auto response = (coeffs.b0 + (coeffs.b1 + coeffs.b2 * z) * z) /
-        (std::complex<double>(1, 0) + (coeffs.a1 + coeffs.a2 * z) * z);
-    magResponseOutput[i] = static_cast<float>(std::abs(response));
-    phaseResponseOutput[i] = static_cast<float>(atan2(imag(response), real(response)));
+    const double omega = kPi * normalizedFreq;
```
maciejmakowski2003, line 143: previously was `double omega = -PI * normalizedFreq;`

packages/react-native-audio-api/src/core/AnalyserNode.ts
```diff
@@ -5,11 +5,18 @@ import { AnalyserOptions } from '../types';
...
+const ANALYSER_ALLOWED_FFT_SIZE: number[] = [
+  32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768,
+] as const;
+
+// The native engine stores these attributes as 32-bit floats, so reading them
+// back loses the exact double the Web Audio spec expects (e.g. 0.8 ->
+// 0.800000011920929). Rounding to float32's ~7 significant digits recovers it.
+function roundFromFloat32(value: number): number {
+  return Number(value.toPrecision(7));
+}
```
maciejmakowski2003: should be in utils file

maciejmakowski2003 (changes requested):

packages/react-native-audio-api/src/core/AnalyserNode.ts
```diff
@@ -5,11 +5,18 @@ import { AnalyserOptions } from '../types';
...
 import { AnalyserOptionsValidator } from '../options-validators';
 
-export default class AnalyserNode extends AudioNode {
-  private static allowedFFTSize: number[] = [
-    32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768,
-  ];
+const ANALYSER_ALLOWED_FFT_SIZE: number[] = [
+  32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384, 32768,
+] as const;
+
```
closetcaiman: Slight repetition with `options-validators.ts`.

packages/react-native-audio-api/src/utils/audioConstants.ts
```diff
@@ -0,0 +1,16 @@
...
+export function assertSupportedSampleRate(sampleRate: number): void {
+  if (
+    sampleRate < MIN_SUPPORTED_SAMPLE_RATE ||
+    sampleRate > MAX_SUPPORTED_SAMPLE_RATE
+  ) {
+    throw new NotSupportedError(
+      `The sample rate provided (${sampleRate}) is outside the range [${MIN_SUPPORTED_SAMPLE_RATE}, ${MAX_SUPPORTED_SAMPLE_RATE}]`
+    );
+  }
+}
```
closetcaiman: Since it's used to validate options, can't that be coupled in that file too? Or even create a better reusable structure here:
```
validation/
  asserts/
  validators/
```

So the placement of checks is actually predictable. In `asserts/` we can keep ad hoc assertions and all that are used by `validators/` in which we would stored coupled, options-specific ones.

packages/react-native-audio-api/src/options-validators.ts
```diff
@@ -2,29 +2,131 @@ import { IndexSizeError, NotSupportedError } from './errors';
...
-    if (options.fftSize && !allowedFFTSize.includes(options.fftSize)) {
+
+    validateAudioNodeOptions(options);
+
+    if (
+      options.fftSize !== undefined &&
+      !ANALYSER_ALLOWED_FFT_SIZE.includes(
+        options.fftSize as (typeof ANALYSER_ALLOWED_FFT_SIZE)[number]
+      )
+    ) {
```
closetcaiman: This cast can be omitted if we create a type:
```js
type FFTSize = typeof ANALYSER_ALLOWED_FFT_SIZE[number];
```

and type `options.fftSize` accordingly. This would also make the type interface even stricter for the end user. It's the result of using `as const` for the array which disallows generic `number` here. 

Alternatively, since we validate this in the runtime, we can just use `ReadonlyArray<number>`

This can be possibly applied to other places where we store constants as arrays.

packages/react-native-audio-api/src/options-validators.ts
```diff
@@ -2,29 +2,131 @@ import { IndexSizeError, NotSupportedError } from './errors';
...
+  maxDecibels: number,
+  minDecibels: number
+): void {
+  if (maxDecibels <= minDecibels) {
+    throw new IndexSizeError(
+      `The maxDecibels value (${maxDecibels}) must be greater than minDecibels`
+    );
+  }
+}
+
```
closetcaiman: Can these be just private methods of `AnalyserOptionsValidator`? For a better segregation and coupling.

packages/react-native-audio-api/src/options-validators.ts
```diff
@@ -2,29 +2,131 @@ import { IndexSizeError, NotSupportedError } from './errors';
...
+      `The channelCountMode value ('${options.channelCountMode}') is not a valid enum value of type ChannelCountMode`
+    );
+  }
+
+  if (
+    options.channelInterpretation !== undefined &&
+    !VALID_CHANNEL_INTERPRETATIONS.includes(options.channelInterpretation)
+  ) {
+    throw new TypeError(
+      `The channelInterpretation value ('${options.channelInterpretation}') is not a valid enum value of type ChannelInterpretation`
```
closetcaiman: `VALID_CHANNEL_INTERPRETATIONS` is not an enum actually, maybe we can make it one? Or rewrite the error message.

packages/react-native-audio-api/common/cpp/audioapi/core/effects/BiquadFilterNode.h
```diff
@@ -104,28 +105,28 @@ class BiquadFilterNode : public AudioNode {
...
-  // delayed samples, one per channel
-  DSPAudioArray x1_;
-  DSPAudioArray x2_;
-  DSPAudioArray y1_;
-  DSPAudioArray y2_;
+  // delayed samples, one per channel (double precision for filter state)
+  std::array<double, MAX_CHANNEL_COUNT> x1_{};
+  std::array<double, MAX_CHANNEL_COUNT> x2_{};
+  std::array<double, MAX_CHANNEL_COUNT> y1_{};
+  std::array<double, MAX_CHANNEL_COUNT> y2_{};
```
closetcaiman, line 112: Why the type change is necessary?

closetcaiman (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/effects/BiquadFilterNode.h
mdydek: audio array operates on floats, here we need to use double to increase accuracy slightly

closetcaiman (approved):

maciejmakowski2003 (approved):

PR #1152, opened 2026-07-07
refactor: removed worklets integration in main package and exposed stable api

maciejmakowski2003:
Part of RNAA-495

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/compatibility/EXTENSION_API.md
```diff
@@ -0,0 +1,152 @@
...
+- `audioapi/AudioAPIModuleInstaller.h` — core global install; not for extensions
+- `audioapi/libs/**`, `audioapi/dsp/**`, `audioapi/external/**`
+- Concrete built-in node HostObjects — reference only, not a stable base API
+
+---
+
+## Versioning
+
+Declare a `react-native-audio-api` **peer dependency** in `package.json` with a
+semver range matching the API you target.
```
mdydek, line 152: I think it doesn't have to be explicitly `peer dependency`, it can also be normal dep

packages/react-native-audio-worklets/package.json
```diff
@@ -0,0 +1,112 @@
...
+  "homepage": "https://github.com/software-mansion/react-native-audio-api#readme",
+  "publishConfig": {
+    "registry": "https://registry.npmjs.org/"
+  },
+  "peerDependencies": {
+    "react": "*",
+    "react-native": "*",
+    "react-native-audio-api": ">= 1.0.0",
+    "react-native-worklets": ">= 0.10.0"
+  },
```
mdydek, line 69: why audio-api and worklets as peer? also there is no version of audio-api 1.0.0 on npm currently, you can point to the specific nightly version and we will update it when the time comes

mdydek (changes requested):
can you also add c++ setup in react-native-audio-worklets, so IDE can correctly map out headers and find them?

mdydek (changes requested):
add also documentation

mdydek (approved):

PR #1182, opened 2026-07-21
refactor: AudioParam

SomePersonFromMars:
Closes RNAA-479

## ⚠️ Breaking changes ⚠️

- Processed `AudioParam` output is now clamped to `[minValue, maxValue]` (spec § 1.6.3). Intrinsic automation math stays unclamped; clipping happens only after modulation in `finalize*`.

## Introduced changes

The PR is an implementation of the [`AudioParam` refactor plan](https://gist.github.com/SomePersonFromMars/2d3a61e385425d36f857ce3189d96fd1).

- Added `GeneralizedAudioParam` / `CompositeAudioParam` and migrated Oscillator, BiquadFilter, and AudioBufferSource computed values onto composites (idempotent `processKRate`/`processARate`).
- Centralized nominal-range clipping at the process output boundary (not in intrinsic `getValueAtTimeUnmodulated`).
- Expanded C++ tests for automation clipping (spec overshooting ramp + linear/exp/target/curve/cancelAndHold cases) and composite param behavior.

## Checklist

- [x] Linked relevant issue
- [ ] Updated relevant documentation
- [x] Added/Conducted relevant tests
- [x] Performed self-review of the code
- [ ] Updated Web Audio API coverage
- [ ] Added support for web
- [ ] Updated old arch android spec file

PR #1184, opened 2026-07-21
test: cover pitchCorrection dropping playbackRate modulation bug

SomePersonFromMars:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- Adds a C++ regression test for `AudioBufferSourceNode` with `pitchCorrection` on and BridgeNode-style `playbackRate` modulation (`inputBuffer_`).
- Asserts source position advances at the modulated rate (~2x); documents the double `processKRateParam` consume that zeros modulation mid-quantum.
- **Note: `AudioBufferSourceNodeTest.PitchCorrectionKeepsPlaybackRateModulationAcrossProcessKRateCalls` *was* failing on purpose**. Now the idempotent `AudioParam` / `refactor/audio-param` fix landed [(see this PR)](https://github.com/software-mansion/react-native-audio-api/pull/1182), which causes the regression test to pass.

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/test/src/core/sources/AudioBufferSourceNodeTest.cpp
```diff
@@ -0,0 +1,147 @@
...
+    channel[0] = modulation;
+  }
+
+  /// Renders one quantum after refilling BridgeNode-style modulation on playbackRate.
+  void renderQuantumWithModulation(TestableAudioBufferSourceNode &node, float modulation) {
+    fillPlaybackRateModulation(node, modulation);
+    node.processNode(QUANTUM);
+    auto clockBuffer =
+        std::make_shared<DSPAudioBuffer>(QUANTUM, 1, static_cast<float>(SAMPLE_RATE));
+    context->processGraph(clockBuffer.get(), QUANTUM);
```
mdydek: as we discussed, currently tests are not setup for graph processing (here your node is not registered, so basically this call processes underlying destination node and it is a noop in this context)

mdydek (changes requested):

packages/react-native-audio-api/common/cpp/test/src/core/sources/AudioBufferSourceNodeTest.cpp
SomePersonFromMars: Necessary changes implemented with respect to the scope of this PR.

mdydek (approved):

PR #1189, opened 2026-07-22
fix: offset default value in AudioBufferQueueSourceNode.start

### LINKED ISSUE #1181: AudioBufferSourceNode start() throws RangeError
### Description

Calling AudioBufferSourceNode's `start()` without providing `offset` causes it to throw RangeError because the `offset` defaults to `-1` in the TypeScript implementation. The documentation states that the default value is `0` which wouldn't produce the error.

As a workaround one must value for `when` and `offset`, so rather than calling `source.start()` you call `source.start(0, 0)`

### Steps to reproduce

1. Call `AudioBufferSourceNode.start()` without any parameters.

https://snack.expo.dev/@jypa/audiobuffersourcenode

0.13.2

### React Native version

0.86.0

### Platforms

iOS

### JavaScript runtime

None

### Workflow

None

### Architecture

None

### Build type

None

### Device

None

### Device model

_No response_

### Acknowledgements

Yes

closetcaiman:
Closes #1181 

## ⚠️ Breaking changes ⚠️

- Introduced explicit `resume(when)` method in `AudioBufferQueueSourceNode`. Calling `start` multiple times now throws `InvalidStateError`. The intented way to resume playback is to use the `resume(when)` method.

## Introduced changes

- changed the default value of `AudioBufferQueueSourceNode.start` `offset`. Previously it was set to `-1` which forced users to explicitly start the node with `.start(0, 0)` as calling `.start()` threw `RangeError` due to `offset` being negative by default.

## Checklist

- [x] Linked relevant issue
- [x] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferQueueSourceNodeHostObject.cpp
```diff
@@ -63,6 +66,16 @@ JSI_HOST_FUNCTION_IMPL(AudioBufferQueueSourceNodeHostObject, pause) {
   return jsi::Value::undefined();
 }
 
+JSI_HOST_FUNCTION_IMPL(AudioBufferQueueSourceNodeHostObject, resume) {
+  auto handle = node_->handle;
+  auto event =
+      [handle, node = bufferQueueSourceNode_, when = args[0].getNumber()](BaseAudioContext &) {
```
mdydek, line 70:
```suggestion
  auto event =
      [handle = node_->handle, node = bufferQueueSourceNode_, when = args[0].getNumber()](BaseAudioContext &) {
```
slightly more unified

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferQueueSourceNodeHostObject.cpp
```diff
@@ -4,7 +4,9 @@
 #include <audioapi/HostObjects/sources/AudioBufferHostObject.h>
 #include <audioapi/core/BaseAudioContext.h>
 #include <audioapi/core/sources/AudioBufferQueueSourceNode.h>
+#include <audioapi/jsi/HostObject.h>
 #include <audioapi/types/NodeOptions.h>
+#include <jsi/jsi.h>
```
mdydek: why those headers were added?

mdydek (changes requested):

mdydek (approved):

PR #1195, opened 2026-07-23
chore: ci job to compare wpt against main

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- added new job in the ci that runs two test harnesses, one on main and one on current branch and then it compares those results, by default main tests won't be passing in 100% thus comparison

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


.github/workflows/wpt-tests.yml
closetcaiman, line 1: I strongly suggest adding the report result in the `pr` comment as `markdown` or any other form of reporting. It will make accessing the results much much easier.

closetcaiman (changes requested):

PR #1196, opened 2026-07-27
fix: sequenced promises in audio and offline contexts

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- as per specification, context promises should be sequenced (https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-pending-promises-slot), previously they run on thread pool, which made the order of those promises unpredictable

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/core/types/ContextPromise.h
closetcaiman, line 1: I think it should be possible to achieve this with only one `template` class. Unless we want explicit readability for valued promises.

closetcaiman, line 1: Is this really a `type` kind of class?

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/OfflineAudioContextHostObject.cpp
```diff
@@ -3,9 +3,34 @@
...
+  auto jsiPromise = std::make_shared<Promise>(std::move(promise));
+  return std::make_shared<OfflineAudioContextResultPromise>(
+      [jsiPromise](const std::shared_ptr<AudioBuffer> &audioBuffer) {
+        auto audioBufferHostObject = std::make_shared<AudioBufferHostObject>(audioBuffer);
+        jsiPromise->resolve([audioBufferHostObject](jsi::Runtime &runtime) {
+          return jsi::Object::createFromHostObject(runtime, audioBufferHostObject);
+        });
+      },
+      [jsiPromise](const std::string &message) { jsiPromise->reject(message); });
+}
```
closetcaiman: Slight repetition with `AudioContextHostObject`. Is it possible to create a proper factory methods for `JSIPromise` -> `ContextPromise` wrapping?

packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.cpp
```diff
@@ -71,72 +73,96 @@ bool AudioContext::tryStartDriver() {
...
   getGraph()->setProducerSelfDrain(true);
   getGraph()->processEvents();
   processAudioEvents();
   audioPlayer_->cleanup();
-}
 
-bool AudioContext::resume() {
-  std::scoped_lock lock(driverMutex_);
+  ContextPromise::resolve(promise);
+  setState(ContextState::CLOSED);
```
closetcaiman: Is the setter needed here if we schedule `statechange` in the promise resolution itself?

packages/react-native-audio-api/common/cpp/audioapi/core/OfflineAudioContext.cpp
```diff
@@ -34,42 +35,58 @@ OfflineAudioContext::OfflineAudioContext(
...
-
+void OfflineAudioContext::renderAudio(const std::shared_ptr<ContextPromise> &resumePromise) {
   // Flush while we are still the sole consumer, then hand the channel to the
   // render thread.
   getGraph()->processEvents();
   getGraph()->setProducerSelfDrain(false);
 
-  std::thread([this]() {
+  std::thread([this, resumePromise = resumePromise]() mutable {
+    setState(ContextState::RUNNING);
```
closetcaiman: If we resolve right after the `statechange`, cannot the `statechange` be invoked on resolution? Same for `ResultPromise`.

closetcaiman (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/OfflineAudioContext.cpp
```diff
@@ -134,7 +135,10 @@ void OfflineAudioContext::startRendering(
 
   renderingStarted_ = true;
   resultPromise_ = promise;
```
closetcaiman, line 135: Why do we store the promise?

packages/react-native-audio-api/common/cpp/audioapi/core/OfflineAudioContext.cpp
```diff
@@ -134,7 +135,10 @@ void OfflineAudioContext::startRendering(
...
-  renderAudio(nullptr);
+  auto runningStatePromise = std::make_shared<ContextPromiseResolver<>>(
+      [self = shared_from_this()]() { self->setState(ContextState::RUNNING); },
+      [](const std::string &) {});
+  renderAudio(runningStatePromise);
```
closetcaiman: Can't this be passed directly from the `OfflineAudioContextHostObject`?

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/BaseAudioContextHostObject.cpp
```diff
@@ -23,10 +23,27 @@
...
+  return std::make_shared<ContextPromiseResolver<>>(
+      [jsiPromise, audioContext, nextState]() {
+        // Spec: update the state attribute in the same follow-up task that
+        // resolves the lifecycle promise (before statechange reactions).
+        audioContext->setState(nextState);
+        jsiPromise->resolve([](jsi::Runtime &runtime) { return jsi::Value::undefined(); });
+      },
+      [jsiPromise](const std::string &message) { jsiPromise->reject(message); });
+}
+
```
closetcaiman: Since it gets the `std::shared_ptr<BaseAudioContext> &audioContext` maybe this could be a method of `ContextPromiseResolver`. Like `ContextPromiseResolver.fromJSI`? This way we don't leak this logic to the `HostObject`.

packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.cpp
```diff
@@ -80,9 +80,9 @@ bool AudioContext::tryStartDriver() {
   return false;
 }
 
-void AudioContext::close(const std::shared_ptr<ContextPromise> &promise) {
+void AudioContext::close(const std::shared_ptr<ContextPromiseResolver<>> &promise) {
```
closetcaiman: I'm not a fan of `<>`. It actually somewhat hides the intent. I'd opt for verbose `ContextPromiseResolver<void>` to indicate that we pass a `promise` that does not return anything.

closetcaiman (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.cpp
mdydek: maybe `using VoidContextPromiseResolver = ContextPromiseResolver<>`?

packages/react-native-audio-api/common/cpp/audioapi/core/OfflineAudioContext.cpp
mdydek: because we can start rendering and then call suspend on the context and then resume, still waiting on the final result

packages/react-native-audio-api/common/cpp/audioapi/core/BaseAudioContext.cpp
```diff
@@ -18,6 +18,19 @@ BaseAudioContext::BaseAudioContext(
...
+              PENDING_PROMISES_OVERFLOW_STRATEGY,
+              PENDING_PROMISES_WAIT_STRATEGY>>(
+              PENDING_PROMISES_CAPACITY,
+              [this](const ContextPromiseTask &task) {
+                if (!task.operation) {
+                  return;
+                }
+                std::scoped_lock lock(driverMutex_);
+                task.operation(*this);
+              })),
```
closetcaiman, line 33: Not a fan of bloating `initialization list`. I'd move to to the constructor.

closetcaiman (approved):

PR #1197, opened 2026-07-27
audioscheduledsource node memory leak fix

### LINKED ISSUE #1187: [Android] Sources stopped before their start time are never reclaimed — each play/stop cycle leaks the full per-source buffer copy (foreground OOM kill)
## Description

We build an ambient-sound engine (myNoise-style overlapping loops) on react-native-audio-api. Each "scene" decodes 5–8 local ogg files (~50–80 MB of float32 PCM per scene) and schedules `AudioBufferSourceNode`s up to 120 s ahead with absolute `start(when)` times (lookahead scheduling, so playback survives JS timer throttling while the screen is off). Stopping a scene calls `stop()` on every created source and drops all JS references to them.

On device, **every play→stop cycle of a single scene leaks ~235 MB of native memory that can never be reclaimed**. After 6–14 scene switches the app is OOM-killed by the kernel *as the foreground process* — no crash record, `am_proc_died [0,pid,...,0,2]` in the events log, RSS goes 0.9 GB → 2.2 GB → (zram squeeze) → SIGKILL.

Measured on a real device (single scene, `load` once, then pure play/stop cycles, `VmRSS` sampled after each cycle):

```
load (5-channel scene, ~78 MB decoded)  rss= 718 MB
cycle 1  rss=  899 MB
cycle 2  rss= 1162 MB
cycle 3  rss= 1396 MB
cycle 4  rss= 1629 MB
cycle 5  rss= 1854 MB
cycle 6  rss= 2082 MB
forced gc()                rss= 2082 MB   (no effect)
```

## Root cause (from reading the 0.13.2 sources)

Three things interact:

1. **`AudioBufferSourceNodeHostObject::setBuffer` makes a full copy of the AudioBuffer per source** (`copiedBuffer = std::make_shared<AudioBuffer>(*buffer);` — acknowledged by the TODO above it). With several sources sharing one buffer, memory is multiplied by the number of sources.

2. **A source stopped at or before its start time stays `SCHEDULED` forever.** In `AudioScheduledSourceNode::updatePlaybackInfo()`, the `isScheduled()` branch takes the `startFrame > lastFrame` early return without ever looking at `stopFrame`. So a scheduled-then-stopped source (a completely normal product of a lookahead scheduler) keeps `playbackState_ == SCHEDULED` until its original — possibly far-future — start time. `AudioGraphManager::canBeDestructed()` requires `use_count() == 1 && (isUnscheduled() || isFinished())`, so the node is not reclaimable and pins its copied buffer.

3. **If the app also `disconnect()`s the stopped source** (a natural cleanup pattern), the node is no longer rendered at all, the state machine can never advance, and the node plus its buffer copy leak permanently. JS-side workarounds cannot recover it:
   - `source.buffer = null` goes through `scheduleAudioEvent(...)`, which is never processed for a node that is no longer rendered;
   - forcing `gc()` does not help (the C++ registry still holds the node, and its state never becomes FINISHED);
   - even `ctx.close()` does not release the memory.

## Minimal repro sketch

```js
const ctx = new AudioContext();
const buffer = await ctx.decodeAudioData(pathToLargeLocalOgg); // e.g. 60 s stereo

for (let cycle = 0; ; cycle++) {
  const sources = [];
  for (let i = 0; i < 8; i++) {
    const src = ctx.createBufferSource();
    src.buffer = buffer;                        // full native copy per source
    src.connect(ctx.destination);
    src.start(ctx.currentTime + 30 + i * 10);   // scheduled ahead, not yet playing
    sources.push(src);
  }
  await sleep(1000);
  for (const src of sources) { src.stop(); src.disconnect(); }
  // drop refs; gc() if available — native memory never returns
}
```

`VmRSS` grows by roughly (buffer bytes × sources) per cycle and is never reclaimed; the process eventually gets OOM-killed with no crash log.

## Environment

- react-native-audio-api **0.13.2** (also reproduced identically on `1.0.0-nightly-5ba1000-20260722`)
- react-native 0.83.1, Hermes, new architecture
- Android 15, vivo V2429A (arm64-v8a), `disableAudioapiFFmpeg=true`

## Suggested fix (verified on device)

Per Web Audio semantics, a source stopped at or before its start time never produces audio and should finish immediately. Handling that first in the `isScheduled()` branch makes the node reclaimable:

```diff
   if (isScheduled()) {
+    // A source stopped at or before its start time will never produce audio:
+    // finish it now so the graph manager's canBeDestructed() predicate
+    // (isUnscheduled() || isFinished()) can reclaim it.
+    if (stopFrame <= startFrame) {
+      startOffset = 0;
+      nonSilentFramesToProcess = 0;
+      playbackState_ = PlaybackState::STOP_SCHEDULED;
+      handleStopScheduled();
+      return;
+    }
     // not yet playing
     if (startFrame > lastFrame) {
       startOffset = 0;
       nonSilentFramesToProcess = 0;
       return;
     }
```

With this patch (applied via patch-package), plus keeping stopped sources connected for a couple of render quanta so the audio thread actually processes the stop, our 40-cycle scene-switch stress test runs with flat RSS (~730–980 MB) where it previously OOM-died at cycle 14.

Two related observations that may deserve their own attention:

- the per-source full buffer copy (point 1) still multiplies transient memory during normal playback; sharing immutable buffers (or copy-on-write) would cut peaks substantially;
- `canBeDestructed()` can never fire for nodes that stop being rendered (e.g. disconnected after `stop()`), since all state transitions happen inside `updatePlaybackInfo()`; a stop-requested-but-unrendered node might deserve a reclaim path of its own.

Happy to open a PR with the one-hunk patch above if you'd take it.

mdydek:
Closes #1187

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- when audio scheduled source node was stopped before start, it leaked memory 
- it cannot be immediately stopped, because `onended` event time fire has to be respected

**Before:**
image

**After:**
image

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
@@ -44,13 +44,14 @@ void AudioScheduledSourceNode::stop(double when) {
   auto stopFrame = dsp::timeToSampleFrame(stopTime_, context->getSampleRate());
   // if stop is requested before start, mark as finished
   if (stopFrame <= startFrame) {
-    // Fire-and-forget: defer finish/disable off the caller thread.
+    playbackState_ = PlaybackState::FINISHED;
+    AudioNode::disable();
+    // Fire-and-forget: defer event dispatch off the caller thread.
     static constexpr double kMillisecondsPerSecond = 1000.0;
     std::thread([this]() {
       std::this_thread::sleep_for(
           std::chrono::milliseconds(static_cast<int>(stopTime_ * kMillisecondsPerSecond)));
-      playbackState_ = PlaybackState::FINISHED;
-      disable();
+      onEndedEvent_.dispatchEmpty();
     }).detach();
     return;
   }
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
@@ -3,14 +3,15 @@
 #include <audioapi/dsp/AudioUtils.hpp>
 #include <audioapi/events/AudioEventHandlerRegistry.h>
 #include <audioapi/utils/AudioArray.hpp>
-
 #if !RN_AUDIO_API_TEST
 #include <audioapi/core/AudioContext.h>
 #endif // RN_AUDIO_API_TEST
 
 #include <algorithm>
+#include <chrono>
 #include <limits>
 #include <memory>
+#include <thread>
 
 namespace audioapi {
 
@@ -38,6 +39,21 @@ void AudioScheduledSourceNode::start(double when) {
 
 void AudioScheduledSourceNode::stop(double when) {
   stopTime_ = when;
+  auto context = context_.lock();
+  auto startFrame = dsp::timeToSampleFrame(startTime_, context->getSampleRate());
+  auto stopFrame = dsp::timeToSampleFrame(stopTime_, context->getSampleRate());
+  // if stop is requested before start, mark as finished
+  if (stopFrame <= startFrame) {
+    // Fire-and-forget: defer finish/disable off the caller thread.
+    static constexpr double kMillisecondsPerSecond = 1000.0;
+    std::thread([this]() {
+      std::this_thread::sleep_for(
+          std::chrono::milliseconds(static_cast<int>(stopTime_ * kMillisecondsPerSecond)));
+      playbackState_ = PlaybackState::FINISHED;
+      disable();
+    }).detach();
+    return;
+  }
 }
 
 bool AudioScheduledSourceNode::isUnscheduled() const {


packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioScheduledSourceNode.cpp
```diff
@@ -38,6 +39,21 @@ void AudioScheduledSourceNode::start(double when) {
...
+  // if stop is requested before start, mark as finished
+  if (stopFrame <= startFrame) {
+    // Fire-and-forget: defer finish/disable off the caller thread.
+    static constexpr double kMillisecondsPerSecond = 1000.0;
+    std::thread([this]() {
+      std::this_thread::sleep_for(
+          std::chrono::milliseconds(static_cast<int>(stopTime_ * kMillisecondsPerSecond)));
+      playbackState_ = PlaybackState::FINISHED;
+      disable();
+    }).detach();
```
closetcaiman, line 55: Is it possible to omit creating another thread?

closetcaiman, line 55: Are we sure that offloading this won't result in `use-after-free`?

closetcaiman (approved):

PR #1207, opened 2026-07-31
fix: render queue event resolution

### LINKED ISSUE #1191: Repeated setTargetAtTime re-scheduling renders as FM warble / inharmonic sidebands on iOS (browsers render a smooth glide)
### Description

Repeated re-scheduling of `AudioParam.setTargetAtTime` on the same param renders as audible FM warble / inharmonic sidebands on iOS, where browsers render a smooth exponential approach to the (moving) target.

We hit this in a meditation app that glides oscillator frequencies with the common "retarget on a timer" pattern: every tick, call `setTargetAtTime(newTarget, ctx.currentTime, tc)` on `OscillatorNode.frequency`. In every browser this produces a clean glide — each call replaces the previous exponential approach and the param converges smoothly. In react-native-audio-api 0.13.2 on iOS the oscillator instead sounds frequency-modulated: a steady drone of 131/198/266 Hz oscillators came out as a dense inharmonic cluster spanning roughly 199–410 Hz (measured with an in-line `AnalyserNode`, 8192-pt FFT, context sample rate 48000).

Notes on severity/scope:

- It is not just fast re-scheduling: 10 Hz retargeting is badly warbled, but even one `setTargetAtTime` every 5 s with a long time constant produces audible pitch instability.
- Same behavior on `BiquadFilterNode.frequency`.
- `linearRampToValueAtTime` is rendered sample-accurately, so `cancelScheduledValues` + `setValueAtTime(param.value)` + `linearRampToValueAtTime` is a working substitute (this is the workaround we shipped).

Possibly related to the `AudioParam` event-handling refactor in #995 / the boundary bug in #1025 — but this reproduces on 0.13.2, i.e. after that fix.

### Steps to reproduce

Minimal repro — one oscillator, retarget `frequency` every 100 ms. Expected (and what browsers do): a smooth glide between 220 and 330 Hz. Actual on iOS: audible warble / sidebands around the pitch.

```tsx
import { AudioContext } from 'react-native-audio-api';

const ctx = new AudioContext();
const osc = ctx.createOscillator();
osc.frequency.value = 220;
osc.connect(ctx.destination);
osc.start();

let high = false;
setInterval(() => {
  high = !high;
  // Browsers: smooth exponential approach toward the new target.
  // RNAA iOS: audible FM warble / inharmonic sidebands.
  osc.frequency.setTargetAtTime(high ? 330 : 220, ctx.currentTime, 0.5);
}, 100);
```

Swapping the `setTargetAtTime` line for the ramp equivalent renders cleanly:

```tsx
const t = ctx.currentTime;
osc.frequency.cancelScheduledValues(t);
osc.frequency.setValueAtTime(osc.frequency.value, t);
osc.frequency.linearRampToValueAtTime(high ? 330 : 220, t + 0.5);
```

To make the artifact visible without listening: insert an `AnalyserNode` before `destination` and FFT the output — with `setTargetAtTime` the spectrum shows a spread of inharmonic partials instead of a single moving peak.

### Secondary observations (mention in case they're useful)

- An `AnalyserNode` on a side branch (connected from a source but not itself on a path to `destination`) is never rendered and reads silence. #860 tracks this; noting it here because it makes debugging this bug harder — the analyser must be in-line.
- `AudioContext.sampleRate` is 48000 on the iOS simulator (worth knowing when analyzing captured PCM; browsers commonly default to 44100).

### Environment

- react-native-audio-api: 0.13.2
- Platform: iOS (reproduced on the iOS simulator; the audible warble was first reported on device)
- React Native: 0.85.3 / Expo SDK 56
- Real-world context: https://deepbreathingexercises.com mobile app — a drone of 3 detuned oscillators re-targeted at 10 Hz became an inharmonic 199–410 Hz cluster; switching to linear ramps restored the intended 131/198/266 Hz stack exactly.

closetcaiman:
Closes #1191 

## ⚠️ Breaking changes ⚠️

- `AudioParam.value` setter will have stricter curve checks (literally the same as `setValueAtTime`) as per spec:

>**value**, of type [float](https://webidl.spec.whatwg.org/#idl-float)
>Setting this attribute has the effect of assigning the requested value to the [[[current value]]](https://webaudio.github.io/web-audio-api/#dom-audioparam-current-value-slot) slot, and calling the [setValueAtTime()](https://webaudio.github.io/web-audio-api/#dom-audioparam-setvalueattime) method with the current [AudioContext](https://webaudio.github.io/web-audio-api/#AudioContext)’s currentTime and [[[current value]]](https://webaudio.github.io/web-audio-api/#dom-audioparam-current-value-slot). **Any exceptions that would be thrown by setValueAtTime() will also be thrown by setting this attribute.**

## Introduced changes

 - `AudioParam.value` setter now invokes `setValueAtTime(value, context.currentTIme)`
   - this solves the bug that occurred when an `automation event` with no `startValue` (like `setTargetAtTime` or `RAMP` events) was scheduled without an explicit `setValueAtTime` call. In result, such event's `startValue` fallbacked to `defaultValue_` which then created audible jumps in pitch and sidebands in `fft` analysis.
   - Before:
   image

   - After:
   image

## Other

> [!NOTE]
>  Identified issues with `WEB IDL` type checking. For example, currently `number.isFinite()` is rarely (if ever) called to check the passed values. Solving this may probably improve `wpt` tests results.

## Checklist

- [x] Linked relevant issue
- [x] Added support for web

### DIFF
--- a/packages/react-native-audio-api/src/core/AudioParam.ts
+++ b/packages/react-native-audio-api/src/core/AudioParam.ts
@@ -29,6 +29,7 @@ export default class AudioParam {
 
   public set value(value: number) {
     this.audioParam.value = value;
+    this.setValueAtTime(value, this.context.currentTime);
   }
 
   public setValueAtTime(value: number, startTime: number): AudioParam {


basiav (approved):

packages/react-native-audio-api/src/core/AudioParam.ts
SomePersonFromMars, line 1: It's worth considering, whether such a business logic lives here. It would probably be beneficial to move it to the C++ layer. On the other hand I understand that executing `setValueAtTime()` at this TS layer gives us `startTime` validation for free. At the same time we probably have too many checks here, as we only use the `checkCurveExclusion` one - that is because we know exactly what is the `startTime` provided inside of `setValueAtTime()`.

The problem obviously isn't significant, but I think it's worth to keep it in mind generally speaking for easier maintenance.

SomePersonFromMars (approved):

Issue #670, opened 2025-09-16
Need Guidance for Duplex Audio Communication in React Native Voice Chat App


fisforfaheem:
### **Project Context**

I'm building a **voice chat application similar to Gemini Live** in React Native (v0.73.6) and struggling with **duplex audio communication** (simultaneous recording and playback). The app is designed for real-time voice conversations with an AI agent.

### **Current Tech Stack**
- **React Native**: 0.73.6
- **Platforms**: iOS and Android
- **Current Audio Packages**:
  - `react-native-audio-api`: ^0.8.2 (iOS)
  - `react-native-audio-session`: ^0.0.6 (iOS)
  - `react-native-audio-recorder-player`: ^3.6.14 (Android)
  - `@react-native-voice/voice`: ^3.2.4
  - `react-native-incall-manager`: ^4.2.1

### **User Story Flow**

1. **User opens app** → HomeScreen with voice chat interface
2. **User taps voice button** → Starts recording user speech
3. **User speaks** → Audio chunks (2048 bytes) sent via WebSocket to backend
4. **AI processes speech** → Backend transcribes and generates response
5. **AI responds** → Audio chunks streamed back via WebSocket
6. **App plays AI response** → While potentially still recording user input
7. **Conversation continues** → Seamless back-and-forth voice interaction

#### **Duplex Communication Problems**
- **iOS**: Audio session conflicts when trying to record while playing
- **Android**: Audio focus issues causing recording to stop during playback
- **Cross-platform**: Inconsistent behavior between platforms
- **Echo/Feedback**: Audio feedback when recording and playing simultaneously

#### **Current Audio Architecture**
```typescript
// Current approach using multiple packages
const audioService = AudioService.getInstance();
const { startRecording, stopRecording, playAudioBuffer } = useAudioDuplex({
  onAudioData: (data: Buffer) => {
    // Send audio chunks to WebSocket
    sendAudioChunk(data);
  }
});
```

#### **Specific Technical Challenges**
1. **Audio Session Management**: iOS AVAudioSession not properly configured for duplex
2. **Buffer Management**: Audio buffer overflow/underflow during simultaneous operations
3. **WebSocket Integration**: Audio chunk timing issues with real-time streaming
4. **Platform Differences**: Different audio APIs requiring separate implementations

### **What I've Tried**

1. **react-native-audio-record** + **react-native-sound** → Failed on iOS duplex
2. **react-native-audio-recorder-player** → File-based, not suitable for streaming
[...]

PR #1210, opened 2026-08-02
feat(android): AudioRecorder input preset option (androidInputPreset)

lpmv:
## Why

The Oboe input stream builder in `AndroidAudioRecorder::openAudioStream` never calls `setInputPreset`, so every capture stream runs on Oboe's implicit default, [`InputPreset::VoiceRecognition`](https://github.com/google/oboe/blob/main/include/oboe/AudioStreamBuilder.h) — the speech-recognition preprocessing chain, which applies **no acoustic echo cancellation**. For duplex voice apps (playing TTS/voice through the same device while recording, VoIP-style — the use case asked about in #670), that makes Android capture echo-raw, while the equivalent iOS setup gets AEC from the `voiceChat` session mode.

## What

An optional constructor option on `AudioRecorder`:

```ts
const recorder = new AudioRecorder({ androidInputPreset: 'voiceCommunication' });
```

mapped to Oboe's `InputPreset` on the capture stream builder. `voiceCommunication` engages the platform AEC/NS chain — the Android twin of iOS `voiceChat`.

- **Backwards compatible**: when the option is omitted (or names an unknown preset) no `setInputPreset` call is made, so existing behavior is preserved exactly.
- iOS ignores the option (its input chain is selected by the AVAudioSession mode); the host object comment says so.
- Presets exposed: `generic`, `camcorder`, `voiceRecognition`, `voiceCommunication`, `unprocessed`, `voicePerformance`.

## Notes

We measured the difference in a duplex probe (1024-sample pcm16 frames at 24 kHz, speaker playback at full volume while recording): on iOS under `voiceChat` the played audio re-enters the mic at +1.9 dB over the room floor; Android capture through the current builder has no platform AEC path at all. Happy to adjust the option's shape (e.g. a general `AudioRecorderOptions` bag was chosen so future per-platform stream options have a home).


mdydek:
looks similar to #1216, will probably try to merge your both approaches into one common api, but need some time to test it

packages/react-native-audio-api/src/AudioAPIModule/globals.d.ts
```diff
@@ -17,7 +17,10 @@ declare global {
     sampleRate: number
   ) => IOfflineAudioContext;
 
-  var createAudioRecorder: () => IAudioRecorder;
+  var createAudioRecorder: (
+    androidInputPreset: string,
+    iosVoiceProcessing: boolean
+  ) => IAudioRecorder;
```
closetcaiman: Wouldn't it be better if we passed an object instead of individual props? In case we would add more options.

packages/react-native-audio-api/src/core/AudioRecorder.ts
closetcaiman, line 1: `withDefaultOptions` helper seems ambiguous now. I think it could handle both file and preset options or split into two verbose helpers then merge, but now the logic is a bit scattered.

packages/react-native-audio-api/ios/audioapi/ios/core/IOSAudioRecorder.mm
```diff
@@ -102,9 +102,12 @@ static void cleanupStartedRecorder(
...
 /// All other necessary fields (like buffers) are initialized in start() method.
 /// This "method" should be called from the JS thread only.
-/// @param audioEventHandlerRegistry Shared pointer to the IAudioEventHandlerRegistry for event handling.
+/// @param audioEventHandlerRegistry Shared pointer to the AudioEventHandlerRegistry for event handling.
+/// @param voiceProcessingEnabled Whether the capture chain runs through Apple's voice-processing I/O
+/// (echo cancellation, noise suppression, automatic gain control).
 IOSAudioRecorder::IOSAudioRecorder(
-    const std::shared_ptr<IAudioEventHandlerRegistry> &audioEventHandlerRegistry)
+    const std::shared_ptr<IAudioEventHandlerRegistry> &audioEventHandlerRegistry,
+    bool voiceProcessingEnabled)
```
closetcaiman: Again, shoudn't this be a member of some `struct IOSAudioRecorderOptions`?

packages/react-native-audio-api/ios/audioapi/ios/core/IOSAudioRecorder.h
```diff
@@ -29,7 +29,9 @@ class AudioFileWriter;
 
 class IOSAudioRecorder : public AudioRecorder {
```
closetcaiman, line 33: Rule of 5.

packages/react-native-audio-api/common/cpp/audioapi/AudioAPIModuleInstaller.h
```diff
@@ -117,8 +119,22 @@ class AudioAPIModuleInstaller {
...
+            iosVoiceProcessing = args[1].getBool();
+          }
+
           auto audioRecorderHostObject = std::make_shared<AudioRecorderHostObject>(
-              audioEventHandlerRegistry, &runtime, jsCallInvoker);
+              audioEventHandlerRegistry,
+              &runtime,
+              jsCallInvoker,
+              androidInputPreset,
+              iosVoiceProcessing);
```
closetcaiman: Do we have to do this parsing inline? Maybe we could extract this logic?

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.h
```diff
@@ -24,7 +24,8 @@ class AndroidAudioRecorder : public oboe::AudioStreamCallback,
                              public std::enable_shared_from_this<AndroidAudioRecorder> {
  public:
   explicit AndroidAudioRecorder(
-      const std::shared_ptr<IAudioEventHandlerRegistry> &audioEventHandlerRegistry);
+      const std::shared_ptr<IAudioEventHandlerRegistry> &audioEventHandlerRegistry,
+      const std::string &inputPreset = "");
```
closetcaiman: Why is this a reference? If `AndroidAudioRecorder` is the utlimate owner of `inputPreset` I think it should just use `std::move` here.

packages/audiodocs/docs/inputs/audio-recorder.mdx
```diff
@@ -379,7 +388,7 @@ Starts the stream from the system audio input device.
 | :---: | :---: | :---- |
 | `options` <Optional /> | [`AudioRecorderStartOptions`](#audiorecorderstartoptions) | Optional recording start configuration. |
 
-#### Returns `Promise<Result<{}>>`.
+###### Returns `Promise<Result<{}>>`.
```
closetcaiman, line 391: Is there a reason for incrementing the heading depth?

packages/audiodocs/docs/inputs/audio-recorder.mdx
```diff
@@ -564,6 +573,40 @@ audioRecorder.clearOnError();
...
+type AndroidInputPreset =
+  | 'generic'
+  | 'camcorder'
+  | 'voiceRecognition'
+  | 'voiceCommunication'
+  | 'unprocessed'
+  | 'voicePerformance';
+```
+
+Maps to the Android [audio input preset](https://developer.android.com/ndk/reference/group/audio#anonymous-enum-9) of the capture stream.
```
closetcaiman: Is this a useful information? Besides, don't we actually use `oboe` `API`? I know they basically wrap the AAudio, but even so the reported number of options do differ: [oboe](https://github.com/google/oboe/blob/0da326e4ef878eac0c032e11ea84ca0a6811aafd/include/oboe/Definitions.h#L470) vs [ndk](https://developer.android.com/ndk/reference/group/audio#anonymous-enum-9)

closetcaiman (changes requested):

packages/react-native-audio-api/src/api.ts
```diff
@@ -38,6 +38,10 @@ export * from './Audio';
 export { default as Audio } from './Audio';
 export { default as AudioControls } from './Audio/controls/AudioControls';
 export type { MediaElementAudioSourceOptions } from './core/MediaElementAudioSourceNode';
+export type {
+  AudioRecorderOptions,
+  AndroidInputPreset,
+} from './core/AudioRecorder';
 export type { default as AudioEventSubscription } from './events/AudioEventSubscription';
 export { default as FilePreset } from './utils/filePresets';
```
closetcaiman, line 43: Is it really a `react` component used in `<audio>`?

closetcaiman (changes requested):

packages/audiodocs/docs/inputs/audio-recorder.mdx
mdydek: yes, with those options

```
---
sidebar_position: 4
toc_min_heading_level: 2
toc_max_heading_level: 5
---
```

h5 also is rendered on the right side as "clickable" shortcut which for sure we don't want to have

closetcaiman (approved):

PR #1211, opened 2026-08-02
feat: AudioManager.getSystemVolume() — synchronous system volume read

lpmv:
## Why

`AudioManager` can observe volume changes (`observeVolumeChanges` + the `volumeChange` event) but has no way to read the volume **before any change has fired**. An app that wants to warn "your volume is at the floor" at the moment the user presses play cannot know the current level until the user happens to press a volume button.

## What

A synchronous read on the exact scale the `volumeChange` event already reports:

```ts
const volume = AudioManager.getSystemVolume(); // 0..1
```

- iOS: `AVAudioSession.outputVolume`
- Android: `STREAM_MUSIC` volume as a fraction of its maximum

Exposed via the TurboModule spec mirroring `getDevicePreferredSampleRate`'s blocking-synchronous form; the iOS implementation delegates to `AudioSessionManager` like its siblings.

## Notes

The JSDoc carries the honest iOS caveat: a cold read before any volume-change event can report a stale or zero value on some iOS versions (KVO on `outputVolume` is the trustworthy stream once it speaks); callers who care can treat an exact 0.0 cold read as suspect. That caveat exists with or without this API — this just gives apps the same read the KVO baseline uses.

### DIFF
--- a/packages/react-native-audio-api/src/system/types.ts
+++ b/packages/react-native-audio-api/src/system/types.ts
@@ -91,4 +91,5 @@ export interface IAudioManager {
   checkNotificationPermissions(): Promise<PermissionStatus>;
   getDevicesInfo(): Promise<AudioDevicesInfo>;
   setInputDevice(deviceId: string): Promise<void>;
+  getSystemVolume(): number;
 }
--- a/packages/react-native-audio-api/src/system/types.ts
+++ b/packages/react-native-audio-api/src/system/types.ts
@@ -80,6 +80,7 @@ export interface IAudioManager {
   observeAudioInterruptions(enabled: boolean): void;
   activelyReclaimSession(enabled: boolean): void;
   observeAudioInterruptions(param: AudioFocusType | boolean | null): void;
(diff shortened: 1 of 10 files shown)


mdydek:
@lpmv can you also please update the documentation? https://github.com/software-mansion/react-native-audio-api/blob/main/packages/audiodocs/docs/system/audio-manager.mdx

packages/react-native-audio-api/src/system/types.ts
```diff
@@ -90,4 +91,5 @@ export interface IAudioManager {
   checkNotificationPermissions(): Promise<PermissionStatus>;
   getDevicesInfo(): Promise<AudioDevicesInfo>;
   setInputDevice(deviceId: string): Promise<void>;
+  getSystemVolume(): number;
```
closetcaiman: Duplicate method declaration with L83.

closetcaiman (changes requested):
Please update the documentation.

closetcaiman (approved):

PR #1218, opened 2026-08-06
refactor: split C++ gtests into smoke/extended runners by category

SomePersonFromMars:
## ⚠️ Breaking changes ⚠️

- None

## Introduced changes

- Split C++ gtests into smoke and extended runners, with extended suites grouped by category (e.g. `graph`).
- CI runs smoke on every non-draft PR; extended categories run when matching paths change or via manual dispatch.
- Local scripts: `yarn test:cpp` / `test:cpp:smoke|extended|full`, plus `validate:cpp` / `validate:cpp-extended`.
- Start CI / Tests / WPT when a draft is marked ready for review, without needing another commit.

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


mdydek (approved):
i think its a good place to unify naming in shell scripts stick to `example-name.sh` convention

.github/workflows/cpp-extended-job.yml
```diff
@@ -0,0 +1,78 @@
...
+        uses: ./.github/actions/checkout-with-node
+
+      - name: Setup Yarn
+        if: steps.decide.outputs.run == 'true'
+        uses: ./.github/actions/setup-yarn
+
+      - name: Setup CMake
+        if: steps.decide.outputs.run == 'true'
+        uses: ./.github/actions/setup-cmake
+
```
closetcaiman, line 75: Can't we reuse `cpp-job.yml`?

.github/workflows/cpp-extended-job.yml
closetcaiman, line 1: The job name is misleading to me. What does it extend? At the end of the workflow it always runs tests, so maybe we could just rename it to `cpp-tests.yml`?

scripts/validate.sh
```diff
@@ -1,14 +1,35 @@
...
+  --cpp-extended   C++ extended (all registered categories)
+  --cpp-full       C++ full (smoke, then all extended categories)
+  --android        Android native build (requires ANDROID_HOME)
+  --ios            iOS native build (macOS only)
+  --full           --fast + C++ extended (all categories) + --android + --ios
+                   (C++ full = smoke from --fast + extended; skips unavailable platforms)
+
+Legacy aliases:
+  --graph          Same as C++ extended category graph only
+                   (prefer --cpp-extended, or yarn test:cpp:extended -- graph)
```
closetcaiman: Is it necessary to keep this? Is it needed for some backwards compatibility with other scripts/workflows?

.github/workflows/tests.yml
```diff
@@ -2,27 +2,32 @@ name: Tests
 
 on:
   pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
   workflow_dispatch:
     inputs:
       run_cpp_tests:
```
closetcaiman, line 8:
```suggestion
      run_cpp_smoke_tests:
```

I'd change it here as well so that it is not confused as `all` suite.

.github/workflows/tests.yml
```diff
@@ -2,27 +2,32 @@ name: Tests
...
         type: boolean
         default: true
       run_js_tests:
         description: Run JS integration tests
         type: boolean
         default: true
+      run_cpp_extended_graph:
+        description: Run C++ extended category graph
+        type: boolean
+        default: false
```
closetcaiman, line 23: I'd reconsider the naming here carefully. `run_cpp_extended_graph` underneath points to the `extended` workflow, but can be read as `extended graph tests` as if there were multiple graph testing suites.

.github/workflows/tests.yml
```diff
@@ -2,27 +2,32 @@ name: Tests
...
 
 jobs:
   cpp-tests:
     if: github.event_name == 'pull_request' || inputs.run_cpp_tests
     uses: ./.github/workflows/cpp-job.yml
     with:
-      name: C++ tests
+      name: C++ smoke
       working-directory: packages/react-native-audio-api
       run: yarn test:cpp
```
closetcaiman, line 32: I think we should align the entry-points. In one place we resort to `yarn` script, in the other we invoke explicitly `extended-job` and pass `category`. Here we could reuse the `extended-job`. I'm against mixing the approaches.

.github/workflows/ci.yml
closetcaiman, line 1: I'd consider sorting workflow files a bit. I know that `.github/worklows` does not like subdirectories, but maybe we could leave only entry-points there? Like `extended-cpp-job` won't be invoked directly so it can be a reusable action - maybe there are more workflows that could be sorted out like that?

closetcaiman (changes requested):
I think in the future we can also extract individual `.md` docs into internal docs. A validation suite manual page would be awesome to have.

.github/workflows/tests.yml
SomePersonFromMars: Indeed that's the intention that they may be multiple graph testing suites:
* smoke - only basic tests
* extended - more time consuming tests

closetcaiman (approved):

.github/workflows/ci.yml
SomePersonFromMars: Yes, that's probably not the right way to use workflows. In either case, that's the convention already present in this repo, so it would require a broader refactor. It means that it is out of scope for now.

PR #1225, opened 2026-08-13
android audio worklets key function

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- on android, there was no key function in one of the host object, which made `getHostObject` call which uses dynamic cast under the hood fails

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-worklets/common/cpp/audioworklets/core/WorkletAudioContext.cpp
```diff
@@ -96,6 +100,10 @@ void WorkletAudioContext::run() {
   // Deadline pacing keeps the soft clock realtime-aligned; sleeping a full
   // quantum after process drifts behind and fills the recorder adapter ring.
   auto nextDeadline = std::chrono::steady_clock::now();
+#ifdef ANDROID
+  // worklet nodes touch jni, so we need to attach this thread to tje jniEnv
```
closetcaiman:
```suggestion
  // worklet nodes touch jni, so we need to attach this thread to the jniEnv
```

closetcaiman (approved):

PR #1228, opened 2026-08-14
fix: os decoder dlopen removal

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- fixed slight bug where m4a files were not decoded on android due to wrong offset read
- bumped min sdk to 24, we support only react native 0.76+ and it was bumped as min version there https://github.com/react-native-community/discussions-and-proposals/discussions/802
- get rid of a very ugly dlopen workaround because of previous min sdk 21

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/AndroidDecoding.cpp
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/AndroidDecoding.cpp
@@ -34,10 +34,11 @@ constexpr int64_t kCodecTimeoutUs = 5000;
 // input without ever flagging EOS output (defensive against a stuck codec).
 constexpr int kMaxTryAgainAfterEos = 200;
 
-struct MediaExtractorDeleter {
-  void operator()(AMediaExtractor *extractor) const {
-    if (extractor != nullptr) {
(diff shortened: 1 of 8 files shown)


packages/react-native-audio-api/android/src/main/cpp/audioapi/android/AndroidDecoding.cpp
```diff
@@ -518,8 +495,13 @@ decoding::DecoderResult AndroidDecoder::open(const decoding::EncodedMemorySource
...
   }
 
-  if (auto configured = configureExtractorMetadata(*state, source.sampleRate);
-      configured.is_err()) {
+  return finishOpen(std::move(state), source.sampleRate);
+}
+
+decoding::DecoderResult AndroidDecoder::finishOpen(
+    std::unique_ptr<AndroidDecoderState> state,
+    int sampleRate) {
```
closetcaiman, line 503: I do like the code deduplication, but I am not a fan of `finishOpen` name. To me it is a bit vague what it actually does. Maybe there is a way to segment `open` more predictably?

closetcaiman (approved):

PR #1229, opened 2026-08-14
fix: problems with rendering and detached threads

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- fast suspend and resume in offline audio context could have a slight window when two rendering threads were running in the same time, which could result in multiple consumers of the events
- removing node lacked self draining
- promise offloader was destroyed too late in the context destructor which could result in use after free
- do not create new threads on audio thread, which results in allocation, changed logic or place of creation new thread

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/core/sources/AudioFileSourceNode.cpp
```diff
@@ -118,13 +118,15 @@ bool AudioFileSourceNode::initDecoder(
...
-  seekDecoderDaemon_ = std::make_unique<SeekDecoderDaemon>(
+  auto deamon = std::make_unique<SeekDecoderDaemon>(
       std::move(daemonOptions),
       decoderState_,
       std::move(commandReceiver),
       std::move(frameSender),
       frameReceiver_);
 
+  seekDecoderThread_ = std::thread(std::move(*deamon));
+
```
closetcaiman: I think `SeekDecoderDaemon` should own the thread, not the `AudioFileSourceNode` since that is an internal property of the daemon worker.

packages/react-native-audio-api/common/cpp/audioapi/core/BaseAudioContext.h
```diff
@@ -164,6 +183,22 @@ class BaseAudioContext : public std::enable_shared_from_this<BaseAudioContext> {
...
+  static constexpr size_t DEFERRED_EMPTY_EVENTS_CAPACITY = 16;
+
+  /// Render-serialized only (audio thread or the synchronous
+  /// `scheduleAudioEvent` path) — no lock, so no other thread may touch it.
+  std::vector<DeferredEmptyEvent> deferredEmptyEvents_;
+
+  void dispatchDueDeferredEvents();
+
   [[nodiscard]] virtual bool isDriverRunning() const = 0;
 };
```
closetcaiman, line 192: Does all this wiring belong to `AudioContext`? Besides can't we reuse some implementation like `BoundedPriorityQueue`?

closetcaiman (changes requested):

closetcaiman (approved):

PR #1232, opened 2026-08-16
feat: respect AudioContextOptions.latencyHint when opening the Android output stream

### LINKED ISSUE #1231: [Android] Hard-coded PerformanceMode::LowLatency causes continuous underruns (crackle) for multi-source playback on low-end devices
## Summary

On Android, `AudioPlayer` opens its Oboe output stream hard-coded to `SharingMode::Exclusive` + `PerformanceMode::LowLatency` with `setFramesPerDataCallback(RENDER_QUANTUM_SIZE)` and no buffer-size request ([AudioPlayer.cpp#L38](https://github.com/software-mansion/react-native-audio-api/blob/fd1d87d/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp#L38)). On entry-level hardware the OS grants the minimum double buffer, and rendering a multi-source graph misses the callback deadline **hundreds of times per second** — heard as continuous crackle. The same graph on a `PerformanceMode::None` stream plays clean on the same device.

Since the whole audio graph renders inside the Oboe data callback, every buffer of headroom matters: `LowLatency` is the right default for interactive use, but there is currently **no way to opt out** for playback-heavy apps, which is exactly the case `AudioContextOptions.latencyHint` exists for in the Web Audio spec. Proposal below; happy to adjust.

## Environment

- `react-native-audio-api` 0.13.2 (the stream configuration is unchanged on current `main`)
- Entry-level Samsung phone (~$200 class), Android, AAudio backend, 48 kHz
- App plays ~11 simultaneous `AudioBufferQueueSourceNode` lanes through per-track gains into a small master chain (all sources `playbackRate` 1.0, no pitch correction — the per-quantum DSP is mostly memcpy + multiplies)

## Measurement

We temporarily patched 0.13.2 to log the granted stream configuration on open and `AudioStream::getXRunCount()` every 2048 callbacks (~5.5 s at 48 kHz / 128 frames).

Granted stream on this device:

```
[oboe] opened api=2 sharing=0 perf=12 rate=48000 burst=96 bufferSize=192 capacity=3072
```

AAudio, Exclusive, LowLatency — granted **192 frames (4 ms) of buffer** out of a 3072-frame capacity, and with no latency tuner it stays there.

While the 11-source session played:

```
[oboe] xruns=1215  bufferSize=192   (t+11s)
[oboe] xruns=2726  bufferSize=192
[oboe] xruns=4284  bufferSize=192
...climbing ~1,100–1,250 per 5.5 s window, continuously...
[oboe] xruns=395578 bufferSize=192  (t+30min)
```

That is **~210–230 underruns per second, sustained** — matching the audible continuous crackle. Two control observations pin it to render load rather than a data or scheduling bug in the app:

- The instant playback stopped, the counter froze at 395,578 and stayed frozen while the callback kept firing over silence (the idle stream renders zeros and meets every deadline).
- With the app idle but a background decode workload running, the counter also climbed — the callback loses the CPU to ordinary app work at a 2.7 ms deadline on little cores.

**A/B on the same device, same session:** reopening the stream with `PerformanceMode::None` (all else identical) → no audible crackle. iOS never exhibits the problem — the AVAudioSession default IO buffer (~23 ms) provides an order of magnitude more scheduling headroom than the 4 ms granted here.

## Proposal: implement `AudioContextOptions.latencyHint`

The Web Audio API already models this exact trade ([spec](https://webaudio.github.io/web-audio-api/#dom-audiocontextoptions-latencyhint)); the library currently ignores the option. Proposed mapping:

| `latencyHint` | Android (Oboe `PerformanceMode`) | iOS | Web |
| --- | --- | --- | --- |
| `'interactive'` (default) | `LowLatency` — current behaviour, unchanged | unchanged | forwarded to browser |
| `'balanced'` | `None` | accepted, no stream change yet | forwarded to browser |
| `'playback'` | `PowerSaving` | accepted, no stream change yet | forwarded to browser |

Design points:

- **Non-breaking:** omitted or unrecognized hints behave exactly as today (`interactive`), matching how browsers treat unknown values.
- The hint rides the `AudioContext` constructor chain (JSI `createAudioContext` gains an optional second argument) down to the platform player; only the Android branch consumes it initially.
- **Numeric hints** (seconds) are out of scope for the first cut — on Android they would map naturally to `setBufferSizeInFrames` within the granted capacity, listed as follow-up.
- **iOS wiring** (e.g. `preferredIOBufferDuration`) is likewise left as follow-up; iOS's default already provides deep-buffer behaviour.

I have a working implementation of the above (validated on the device that reproduced the bug, via a backport of the same change to 0.13.2) and will open a PR referencing this issue.

DanielOrmeno:
## Summary

Implements `AudioContextOptions.latencyHint` (category form) and maps it to each backend's stream configuration, so playback-heavy apps can opt out of the underrun-prone low-latency stream on Android, and latency-sensitive apps can opt into a short buffer on iOS.

Closes #1231 — measurements there: on an entry-level Samsung the hard-coded `LowLatency` stream is granted 192 frames (4 ms) and misses ~210–230 render deadlines per second while an 11-source graph plays. The same graph on `PerformanceMode::None` plays clean.

## Mapping

| `latencyHint` | Android (Oboe `PerformanceMode`) | iOS (`AVAudioSession.preferredIOBufferDuration`) | Web |
| --- | --- | --- | --- |
| omitted | `LowLatency` — unchanged | not requested — unchanged | browser default |
| `'interactive'` | `LowLatency` | 128 frames | forwarded |
| `'balanced'` | `None` | 512 frames | forwarded |
| `'playback'` | `PowerSaving` | not requested — the default is already deep | forwarded |

iOS is the mirror image of the Android problem: the backend never requested a buffer duration, so it ran deep and could not go low. That is why the two columns aren't symmetric, and why an omitted hint is not treated as `interactive` — doing so would drop every existing iOS playback app to a ~2.7 ms buffer. An unrecognised string is likewise treated as no hint, where a browser throws a `TypeError`.

On iOS the duration belongs to the process-wide session rather than to one stream, so live requests reconcile by taking the shortest; `balanced` and `playback` therefore lose to a concurrent `interactive`. Releasing the last request asks for what the session ran at beforehand, snapped to the nearest power-of-two frame count, because the session grants those and rounds a raw request down.

## What changed

- `AudioContextLatencyHint` in `common/cpp/audioapi/core/types/` plus the hint→frames policy; `latencyHintFromString` sits in `js_enum_parser` with the other string→enum conversions.
- Android's `AudioPlayer` maps the hint to a `PerformanceMode` when opening the stream.
- iOS: `AudioSessionManager` reconciles buffer-duration requests per client; `NativeAudioPlayer` holds one while rendering and releases it on stop, suspend, cleanup and every failure path.
- JSI `createAudioContext` gains an optional second string argument, in both the module installer and the WPT binding.
- TS: `AudioContextOptions.latencyHint?: AudioContextLatencyCategory`, forwarded to the browser on web.
- Docs: `audio-context.mdx` constructor table plus a per-platform table.

Non-breaking: every new constructor parameter is defaulted, and an omitted hint is today's behaviour on both platforms.

Out of scope, per #1231: numeric hints in seconds, which would map to `setBufferSizeInFrames` on Android.

## Test plan

- `typecheck`, `lint:js`, `format:check` (clang-format + ktlint), `cpplint` — clean.
- `test:js` — 120/120, including a suite pinning that an omitted hint reaches native as omitted, on both the native and web paths.
- `common/cpp/test/run-tests.sh smoke` — 428/428, including `AudioContextLatencyHintTest` over the parse table and the frames policy. Mutation-checked: remapping `interactive`, making `playback` or an absent hint request a buffer, and changing the balanced multiple each fail the suite.
- `build:ios` (xcodebuild, iOS Simulator) — clean.
- WPT `audiocontextoptions.html` run on this branch and on pristine `main` — 37/41 both, identical.
- iOS behaviour on an iPhone 17 simulator via `context.outputLatency`: omitted 10.10 ms with no request issued, `interactive` 2.77 ms (128 frames granted, at 48 kHz and 44.1 kHz contexts), `balanced` 10.77 ms (512 frames), deep buffer restored after close.
- Android validated earlier on the device from #1231 via a backport to 0.13.2: `interactive` reproduces the underruns, `balanced` plays the 11-source session clean (`AudioStream::getXRunCount()` logging).
- Not run on physical iPhone hardware; the wired and Bluetooth routes are unmeasured.

### DIFF
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.h
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.h
@@ -10,6 +10,7 @@
 #include <mutex>
 
 #include <audioapi/core/CommonPlayer.h>
+#include <audioapi/core/types/AudioContextLatencyHint.h>
 #include <audioapi/utils/AudioBuffer.hpp>
 
 namespace audioapi {
@@ -29,7 +30,8 @@ class AudioPlayer : public CommonPlayer,
       int channelCount,
       std::mutex *driverMutex,
       const std::shared_ptr<AudioContext> &context,
-      std::atomic<uint32_t> &currentRenders);
+      std::atomic<uint32_t> &currentRenders,
+      AudioContextLatencyHint latencyHint = AudioContextLatencyHint::INTERACTIVE);
 
   ~AudioPlayer() override {
     cleanup();
@@ -67,6 +69,7 @@ class AudioPlayer : public CommonPlayer,
   std::atomic<int32_t> lastCallbackFrameCount_{0};
   std::mutex *driverMutex_;
   std::weak_ptr<AudioContext> context_;
+  AudioContextLatencyHint latencyHint_;
 
   bool openAudioStream();
 };
--- a/packages/react-native-audio-api/common/cpp/audioapi/core/types/AudioContextLatencyHint.h
+++ b/packages/react-native-audio-api/common/cpp/audioapi/core/types/AudioContextLatencyHint.h
@@ -1,0 +1,12 @@
+#pragma once
+
+#include <cstdint>
+
+namespace audioapi {
+
+/// Web Audio's AudioContextOptions.latencyHint categories: what the context should
+/// optimize its output stream for. Platform backends map these to their own stream
+/// configuration; INTERACTIVE preserves the pre-hint behaviour and is the default.
+enum class AudioContextLatencyHint : std::uint8_t { INTERACTIVE, BALANCED, PLAYBACK };
+
+} // namespace audioapi
--- a/packages/audiodocs/docs/core/audio-context.mdx
+++ b/packages/audiodocs/docs/core/audio-context.mdx
@@ -21,6 +21,7 @@ constructor(options?: AudioContextOptions)
 | Parameter | Type | Default | |
 | :---: | :---: | :----: | :---- |
 | `sampleRate` <Optional /> | `number` | - | The preferred sample rate for the context. |
+| `latencyHint` <Optional /> | `'interactive' \| 'balanced' \| 'playback'` | `'interactive'` | What the context should optimize its output stream for. `interactive` requests the lowest latency the platform offers; `balanced` and `playback` trade output latency for a deeper buffer, which favours glitch-free sustained playback (many simultaneous sources, low-end devices). On Android these map to Oboe's `LowLatency`, `None` and `PowerSaving` performance modes; on iOS the hint is currently accepted but does not change the stream; on web it is passed to the browser's `AudioContext`. Numeric hints are not supported yet. |
 
 #### Errors
 
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
@@ -13,20 +13,38 @@
 
 namespace audioapi {
 
+namespace {
+
+PerformanceMode performanceModeFor(AudioContextLatencyHint latencyHint) {
+  switch (latencyHint) {
+    case AudioContextLatencyHint::BALANCED:
+      return PerformanceMode::None;
+    case AudioContextLatencyHint::PLAYBACK:
+      return PerformanceMode::PowerSaving;
+    case AudioContextLatencyHint::INTERACTIVE:
+      return PerformanceMode::LowLatency;
+  }
+  return PerformanceMode::LowLatency;
+}
+
+} // namespace
+
 AudioPlayer::AudioPlayer(
     const std::function<void(DSPAudioBuffer *, int)> &renderAudio,
     float sampleRate,
     int channelCount,
     std::mutex *driverMutex,
     const std::shared_ptr<AudioContext> &context,
-    std::atomic<uint32_t> &currentRenders)
+    std::atomic<uint32_t> &currentRenders,
+    AudioContextLatencyHint latencyHint)
     : renderAudio_(renderAudio),
       currentRenders_(currentRenders),
       sampleRate_(sampleRate),
       channelCount_(channelCount),
       isRunning_(false),
       driverMutex_(driverMutex),
-      context_(context) {}
+      context_(context),
+      latencyHint_(latencyHint) {}
 
 bool AudioPlayer::openAudioStream() {
   std::scoped_lock lock(streamMutex_);
@@ -35,7 +53,7 @@ bool AudioPlayer::openAudioStream() {
   builder.setSharingMode(SharingMode::Exclusive)
       ->setFormat(AudioFormat::Float)
       ->setFormatConversionAllowed(true)
-      ->setPerformanceMode(PerformanceMode::LowLatency)
+      ->setPerformanceMode(performanceModeFor(latencyHint_))
       ->setChannelCount(channelCount_)
       ->setSampleRateConversionQuality(SampleRateConversionQuality::Medium)
       ->setFramesPerDataCallback(RENDER_QUANTUM_SIZE)
--- a/packages/react-native-audio-api/common/cpp/audioapi/AudioAPIModuleInstaller.h
+++ b/packages/react-native-audio-api/common/cpp/audioapi/AudioAPIModuleInstaller.h
@@ -60,7 +60,7 @@ class AudioAPIModuleInstaller {
     return jsi::Function::createFromHostFunction(
         *jsiRuntime,
         jsi::PropNameID::forAscii(*jsiRuntime, "createAudioContext"),
-        1,
+        2,
         [jsCallInvoker, audioEventHandlerRegistry](
             jsi::Runtime &runtime,
             const jsi::Value &thisValue,
@@ -68,8 +68,20 @@ class AudioAPIModuleInstaller {
             size_t count) -> jsi::Value {
           auto sampleRate = static_cast<float>(args[0].getNumber());
 
+          // Unknown strings fall back to INTERACTIVE, matching how browsers
+          // treat an unrecognised latencyHint.
+          auto latencyHint = AudioContextLatencyHint::INTERACTIVE;
+          if (count > 1 && args[1].isString()) {
+            auto hint = args[1].getString(runtime).utf8(runtime);
+            if (hint == "balanced") {
+              latencyHint = AudioContextLatencyHint::BALANCED;
+            } else if (hint == "playback") {
+              latencyHint = AudioContextLatencyHint::PLAYBACK;
+            }
+          }
+
           auto audioContextHostObject = std::make_shared<AudioContextHostObject>(
-              sampleRate, audioEventHandlerRegistry, &runtime, jsCallInvoker);
+              sampleRate, audioEventHandlerRegistry, &runtime, jsCallInvoker, latencyHint);
 
           return jsi::Object::createFromHostObject(runtime, audioContextHostObject);
         });
--- a/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioContextHostObject.cpp
+++ b/packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioContextHostObject.cpp
@@ -13,9 +13,10 @@ AudioContextHostObject::AudioContextHostObject(
     float sampleRate,
     const std::shared_ptr<IAudioEventHandlerRegistry> &audioEventHandlerRegistry,
     jsi::Runtime *runtime,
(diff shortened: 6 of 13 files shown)


DanielOrmeno:
Updated the branch against current `main` — the conflict was a single import line in `globals.d.ts` (`AudioRecorderOptions` landed on the line the hint's type was added to). No behaviour change; the rest of the diff is as it was. I merged rather than rebased so the PR history stays intact, but happy to rebase or squash if you'd prefer that.

Verified locally on the updated branch:

- `format:check`, `lint:js`, `typecheck` — clean
- `yarn build && yarn test:js` — 113/113
- `check-audio-enum-sync` — in sync
- C++ still compiled only by CI (no NDK in my environment)

CI hasn't run on this PR at all — the workflow runs are sitting at `action_required`, so I think it needs a maintainer to approve them. Could someone kick that off when convenient?

On next steps, I'd welcome any direction: whether the `interactive`/`balanced`/`playback` → Oboe mapping is the shape you want, and whether the two follow-ups noted in #1231 (numeric hints → `setBufferSizeInFrames`, iOS `preferredIOBufferDuration`) should become separate PRs.

One thing we ran into downstream that may deserve its own issue: on a Samsung SM-A176U1 with a USB audio interface, the hard-coded `SharingMode::Exclusive` output stream crashed the vendor HAL on start (`requestStart` → -899, then the AAudio service died); requesting `Shared` avoids it, and low-latency still goes through AAudio's MMAP path. Glad to file that separately with logs if it's useful.

Thanks, and no rush — just wanted to get the branch mergeable again.

mdydek:
to be honest ideally we wanted to have also ios side implemented in this pr, we played around a little bit, but couldn't achieved any satisfying solution, that's why it was sitting idle for a while, also thank you for sharing that sometimes `exclusive` is also too much

DanielOrmeno:
Makes total sense, I'm having a go at ios too, will update the PR accordingly.

DanielOrmeno:
Pushed the iOS side.

Why it's shaped differently from the Android half: the iOS backend never called `setPreferredIOBufferDuration` at all, so output ran at whatever `AVAudioSession` defaulted to for the category (~21–23 ms). Android hard-codes low latency and can't get headroom; iOS ran deep and couldn't get low latency. So on iOS `balanced` and `playback` are close to no-ops, and `interactive` is the hint that was previously unimplementable.

That changed one thing from the original PR: an omitted hint now stays omitted (`std::optional`) rather than defaulting to `INTERACTIVE`. On Android that's identical to today. On iOS, treating omitted as `interactive` would drop every shipped playback app to a ~2.7 ms buffer — #1231 arriving on iOS. Unrecognised strings are treated the same way rather than throwing, which is a deliberate deviation from the spec's `TypeError`.

Measured on an iPhone 17 simulator via `context.outputLatency`: omitted 10.10 ms with no request issued, `interactive` 2.77 ms (128 frames granted), `balanced` 10.77 ms (512 frames), `playback` 10.77 ms, and the deep buffer restored after the interactive context closed.

Two platform details worth knowing:

- iOS rounds a requested duration *down* to a power-of-two frame count — asking for a 23 ms baseline back was granted 11 ms — so the restore snaps to the nearest power of two.
- The request has to be in frames, not seconds. Deriving it from the context's rate made `interactive` ask for 256 frames on a 44.1 kHz context driving 48 kHz output. The 2.77 ms above holds at both rates.

The hint→frames policy and the string parse moved into `common/cpp` so `common/cpp/test` can reach them; the platform TUs aren't in that target's glob, so neither mapping table had test coverage available before.

On WPT, I built the node addon and ran `audiocontextoptions.html` on this branch and on pristine `main` — 37/41 both, identical. `interactive baseLatency == defaultLatency` only passes because iOS `getBaseLatency()` returns one render quantum regardless of the granted buffer; fixing that looks right but flips the subtest, so it's not in here.

A few things I'd rather you decided:

- `createAudioRecorder` takes one JSI object via `AudioRecorderOptions::CreateFromJSIValue`, while `createAudioContext` hand-decodes `args[1]` in two installers. An `AudioContextOptions` struct is probably the shape now there's a second field — happy to convert.
- `nextPowerOfTwo` now exists twice under `ios/audioapi` (mine and `NativeAudioRecorder.m`'s). Sharing it needs a new C-safe header since the recorder is plain ObjC, so I left it rather than touch that file.
- `AudioEngine` already mints client handles in `sourceRegistrations`, and my reconciliation registry in `AudioSessionManager` is a second one. Folding the frame count onto the source registration may be the better seam.

Nothing has run on physical iPhone hardware — the simulator doesn't grant buffer durations the way a device does, and the wired and Bluetooth routes are unmeasured. Happy to do that before you merge.

CI hasn't run on this branch yet (still `action_required`).

mdydek:
ok few observations when playing around on your branch with various option for the latencies:
- both android emulator and ios simulator do not properly mimic all hardware behavior, thus their reported data and values can differ from the physical devices
- tested on the app, where the audio thread stalls were introduced once per 250 ms for x ms, to get to know real hardware buffer which serves as the protecting layer for the rendering spikes where the audio is rendered longer than usually (upper limits provided):
    - **android:** interactive: 5ms, balanced: 70ms, playback: 70ms
    - **ios:** interactive: 5ms, balanced (iobuffer as 8 x RQ): 20ms, playback (iobuffer as 32 x RQ): 80ms
- on ios requesting greater amount of iobuffer than 4096 results in 1028
    
I think the approach is correct, need probably some adjustments f.e. ios buffer size constants, but overally respecting this implementation should provide a way for better ux of the audio

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.h
```diff
@@ -67,6 +70,7 @@ class AudioPlayer : public CommonPlayer,
   std::atomic<int32_t> lastCallbackFrameCount_{0};
   std::mutex *driverMutex_;
   std::weak_ptr<AudioContext> context_;
+  std::optional<AudioContextLatencyHint> latencyHint_;
```
mdydek, line 73: I don't like the idea that in one of the player the latency hint is its state, whereas in the other it is not

packages/react-native-audio-api/common/cpp/audioapi/core/types/AudioContextLatencyHint.h
```diff
@@ -0,0 +1,31 @@
...
+    case AudioContextLatencyHint::INTERACTIVE:
+      return RENDER_QUANTUM_SIZE;
+    case AudioContextLatencyHint::BALANCED:
+      return 4 * RENDER_QUANTUM_SIZE;
+    case AudioContextLatencyHint::PLAYBACK:
+      return 0;
+  }
+
+  return 0;
+}
```
mdydek, line 29: android completely do not use this function, move it to the utils in the ios side

packages/react-native-audio-api/common/cpp/audioapi/core/types/AudioContextLatencyHint.h
```diff
@@ -0,0 +1,31 @@
...
+    return 0;
+  }
+
+  switch (*latencyHint) {
+    case AudioContextLatencyHint::INTERACTIVE:
+      return RENDER_QUANTUM_SIZE;
+    case AudioContextLatencyHint::BALANCED:
+      return 4 * RENDER_QUANTUM_SIZE;
+    case AudioContextLatencyHint::PLAYBACK:
+      return 0;
```
mdydek, line 25:
```suggestion
    case AudioContextLatencyHint::BALANCED:
      return 8 * RENDER_QUANTUM_SIZE;
    case AudioContextLatencyHint::PLAYBACK:
      return 32 * RENDER_QUANTUM_SIZE;
```
my proposition for the values, 0 is treated as default (8 * RQS), we can enlarge it to 32 * RQS

packages/react-native-audio-api/common/cpp/test/src/core/AudioContextLatencyHintTest.cpp
mdydek, line 1: trivial test, remove it

packages/react-native-audio-api/ios/audioapi/ios/system/AudioSessionManager.mm
```diff
@@ -63,12 +92,88 @@ - (bool)areDesiredOptionsSet
...
+    }
+
+    if (frames == 0 && self.baselineIOBufferDuration <= 0.0) {
+      return;
+    }
+
+    double sampleRate = self.audioSession.sampleRate;
+    if (sampleRate <= 0.0) {
+      return;
+    }
```
mdydek, line 120: looks like kind off too much work for taking min from the dictionary and few returns

mdydek (changes requested):
ios side introduces much more lines than android and I think it could be simplified a little bit

PR #1234, opened 2026-08-17
Feat/android fgs survive task removal

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

- get rid of the icons in the recording notification, the ui is different based on the manufacturer, android versions, thus the current approach is to fallback to simple text buttons, which will be placed by the system possibly in the correct arrangement
- `enableFileOutput` cannot be set once recorder is started
## Introduced changes

With stopWithTask=false a recording outlives the app UI, but a remounted screen (or relaunched app) had no way to learn about it — recorder state was only reachable through the instance that started it:

new JSI globals backed by ActiveRecorderHandle, surfaced as statics: AudioRecorder.isRecordingOngoing() and the consume-once AudioRecorder.takeLastRecordingResult() for files finalized by the notification stop action (mock parity + jest coverage included)
Record demo mounts directly in the live recorder's state, picks up natively stopped files, keeps the recording alive across screen exits and only enables file output when no session is ongoing (re-enabling mid-recording replaces the writer and resets the duration)
deep-link routing for the notification tap (react-navigation linking), duration displays seeded from the recorder instead of assuming a fresh session, and RecordingTime rewritten to plain state — the animated-prop binding went stale on the frozen value while paused and showed zeros

possibility of swiping away notification is configurable, it can be easily swiped, thus ending recording or unswipable, making it stick, but in android 14+ it cannot be made for certain notifications, so it is being reattached in this scenario

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/audiodocs/docs/inputs/audio-recorder.mdx
```diff
@@ -93,6 +93,48 @@ Additionally to be able to record audio while application is in the background,
   </TabItem>
 </Tabs>
 
+### Keeping the recording alive when the app is closed
```
closetcaiman: Is this `android` specific? Maybe we could add a badge?

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioRecorderCallback.h
```diff
@@ -42,7 +42,11 @@ class AudioRecorderCallback {
   void clearOnErrorCallback() {
     assignOnErrorCallbackId(0);
   }
-  void assignOnErrorCallbackId(uint64_t callbackId);
+  // Defined inline so AudioRecorder.cpp doesn't drag this class's whole
```
closetcaiman: ?

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.h
closetcaiman, line 1: Is it possible to omit/reduce this trip to gain access to Recorder?

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/notification/RecordingNotification.kt
```diff
@@ -247,75 +229,86 @@ class RecordingNotification(
...
+    state.usesChronometer =
+      if (options?.hasKey("usesChronometer") == true) {
+        options.getBoolean("usesChronometer")
       } else {
-        state.resumeIconResourceName ?: null
+        state.usesChronometer
       }
-    state.backgroundColor = if (options?.hasKey("color") == true) options.getInt("color") else state.backgroundColor ?: null
+    // Unlike the other options, `paused` resets when absent so the notification never
+    // sticks in the paused look.
```
closetcaiman: Can we refactor this to use some helper methods? All these state modifications share the same logic.

closetcaiman (changes requested):

apps/common-app/src/demos/Record/Record.tsx
```diff
@@ -21,7 +21,17 @@ import Status from './Status';
 import { RecordingState } from './types';
 
 const Record: FC = () => {
-  const [state, setState] = useState<RecordingState>(RecordingState.Idle);
+  // A recording can outlive this screen (and, with `stopWithTask: false`, the whole
+  // app UI). Mounting directly in the right state lets every child initialize from
+  // the live recorder instead of transitioning out of a transient Idle render.
```
michalsek:
```suggestion
  // Recover from "app disabled" state - recording can survive the app kill (android)
```

apps/common-app/src/demos/Record/Record.tsx
```diff
@@ -130,15 +154,22 @@ const Record: FC = () => {
...
 
-    const outputPath = info.paths[0].replace(/[^/]+$/, 'recording.m4a');
+    await loadRecordedAudio(info.paths);
+  }, [loadRecordedAudio]);
 
-    const finalPath = await concatAudioFiles(info.paths, outputPath);
-    const audioBuffer = await audioContext.decodeAudioData(finalPath);
-    setRecordedBuffer(audioBuffer);
+  // The stop action already stopped the recorder natively and hid the notification;
+  // here we only pick up the resulting files and sync the UI.
```
michalsek:
```suggestion
```

apps/common-app/src/demos/Record/Record.tsx
```diff
@@ -254,22 +295,50 @@ const Record: FC = () => {
...
+    const info = AudioRecorder.takeLastRecordingResult();
+    if (info && info.paths.length > 0) {
+      loadRecordedAudio(info.paths);
+    }
+  }, [loadRecordedAudio]);
 
   useEffect(() => {
-    Recorder.enableFileOutput({ rotateIntervalBytes: 1_000_000, format: FileFormat.M4A });
+    // Re-enabling file output during an ongoing recording replaces the file writer,
+    // which starts a new file and resets the duration — skip it when resyncing.
```
michalsek:
```suggestion
```

michalsek: might be also worth changing on the native side to handle that and make enable a no-op in that case

apps/common-app/src/demos/Record/Record.tsx
```diff
@@ -254,22 +295,50 @@ const Record: FC = () => {
...
+    if (!AudioRecorder.isRecordingOngoing()) {
+      Recorder.enableFileOutput({ format: FileFormat.Wav });
+    }
 
     return () => {
+      // The recording and its notification intentionally stay alive when leaving this
+      // screen; they can be stopped from the notification or after coming back.
```
michalsek:
```suggestion
```

apps/common-app/src/demos/Record/Record.tsx
```diff
@@ -254,22 +295,50 @@ const Record: FC = () => {
...
       stopPlayback();
-      Recorder.disableFileOutput();
-      Recorder.stop();
-      AudioManager.setAudioSessionActivity(false);
-      RecordingNotificationManager.hide();
+
+      if (!AudioRecorder.isRecordingOngoing()) {
+        AudioManager.setAudioSessionActivity(false);
+      }
```
michalsek, line 340: same here

apps/common-app/src/demos/Record/RecordingTime.tsx
```diff
@@ -1,71 +1,53 @@
...
-        .toString()
-        .padStart(2, '0');
-      const milliseconds = Math.floor((elapsedSeconds % 1) * 1000)
-        .toString()
-        .padStart(3, '0');
-
-      durationStringSV.value = `${minutes}:${seconds}:${milliseconds}`;
-    }, 100);
+    // Also refresh immediately so a paused or resynced screen shows the real
+    // duration before the first interval tick.
```
michalsek:
```suggestion
```

apps/common-app/src/demos/Record/RecordingVisualization.tsx
```diff
@@ -108,63 +90,23 @@ function drawDefaultWaveform(params: DrawDefaultWaveformParams) {
...
-  );
-
   const translateX = useSharedValue(0);
   const lastIndex = useSharedValue(-1);
-  const durationMS = useSharedValue(0);
+  // The worklet only accumulates duration from buffers it sees while this component
+  // is mounted; when the screen re-attaches to an already-running recording, start
+  // from the recorder's real elapsed time. Seeding here (not in an effect) matters:
+  // TimeStream's children position their ticks from this value during their own
+  // mount, which happens before any parent effect could run.
```
michalsek:
```suggestion
```

apps/common-app/src/demos/Record/TimeStream.tsx
```diff
@@ -23,25 +23,27 @@ interface TimeStreamProps {
   durationMS: SharedValue<number>;
 }
 
-function generateInitialTimestamps() {
+// Seconds around `baseSecond` so the visible window is fully populated even when
+// the stream starts mid-recording (screen re-attached to a live recorder).
```
michalsek:
```suggestion
```

apps/common-app/src/demos/Record/RecordingVisualization.tsx
```diff
@@ -419,6 +286,13 @@ const RecordingVisualization: React.FC<RecordingVisualizationProps> = ({
 
   useEffect(() => {
     if (state === RecordingState.Recording) {
+      if (size.width === 0) {
+        // Canvas not measured yet (mounting straight into an ongoing recording).
+        // Starting the scroll animation now would pin translateX at 0 and draw the
+        // waveform off-screen; this effect re-runs once the size arrives.
```
michalsek:
```suggestion
```

apps/common-app/src/App.tsx
```diff
@@ -181,10 +181,21 @@ const MainTabsScreen: FC = () => {
   );
 };
 
+// Routes notification taps (e.g. the recording notification's `deepLinkUri`)
+// straight to the right screen instead of the app's entry screen.
```
michalsek:
```suggestion
```

packages/react-native-audio-api/tests/mock.test.ts
```diff
@@ -253,6 +253,23 @@ describe('React Native Audio API Mocks', () => {
...
+      recorder.pause();
+      expect(MockAPI.AudioRecorder.isRecordingOngoing()).toBe(true);
+
+      await recorder.stop();
+      expect(MockAPI.AudioRecorder.isRecordingOngoing()).toBe(false);
+    });
+
+    it('should return null when no native stop occurred', () => {
+      expect(MockAPI.AudioRecorder.takeLastRecordingResult()).toBeNull();
+    });
```
michalsek:
```suggestion
```

packages/react-native-audio-api/src/system/notification/types.ts
```diff
@@ -77,14 +77,35 @@ export interface RecordingNotificationInfo {
...
+  stopActionTitle?: string;
+  /**
+   * URI attached to the notification tap intent, e.g. `myapp://record`.
+   * Delivered through React Native's `Linking` (initial URL on cold start,
+   * `url` event otherwise), so it can route to a specific screen. Without it,
+   * tapping the notification opens the app's launcher activity.
+   */
+  deepLinkUri?: string;
+  /** Shows the elapsed recording time in the notification. Default: false. */
+  usesChronometer?: boolean;
```
michalsek, line 102: jsdoc-style

packages/react-native-audio-api/src/plugin/withAudioAPI.ts
```diff
@@ -15,6 +15,13 @@ interface Options {
   androidPermissions: string[];
   androidForegroundService: boolean;
   androidFSTypes: string[];
+  /**
+   * Controls `android:stopWithTask` on the injected foreground service. When
+   * false, swiping the app away from recents keeps the service — and therefore
+   * the app process and any in-progress recording — running (Android calls
+   * onTaskRemoved instead of stopping the service). Defaults to true.
+   */
```
michalsek:
```suggestion
```

packages/react-native-audio-api/src/core/AudioRecorder.ts
```diff
@@ -56,6 +56,31 @@ export default class AudioRecorder {
...
 
+  /**
+   * Checks whether a recording session is ongoing (recording or paused). Native
+   * source of truth that needs no reference to the recorder instance, so a
+   * remounted screen (e.g. after navigating away and back, or reopening an app
+   * whose recording kept running under an Android foreground service with
+   * `stopWithTask: false`) can seed its UI state from it. Reflects the most
+   * recently created `AudioRecorder` — constructing another instance
+   * mid-recording displaces the probed one.
+   */
```
michalsek:
```suggestion
```

packages/react-native-audio-api/src/core/AudioRecorder.ts
```diff
@@ -56,6 +56,31 @@ export default class AudioRecorder {
...
+  }
+
+  /**
+   * Returns the file info of a recording that was stopped natively (via the
+   * recording notification stop action, which finalizes the files even when no
+   * JS listener is reachable), or `null` if there is none. Consume-once: the
+   * result is cleared on read, so a second call returns `null`. Recordings
+   * stopped through {@link stop} resolve their promise with the file info
+   * instead and never appear here.
+   */
```
michalsek:
```suggestion
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioRecorderCallback.h
michalsek: 🤔

michalsek (commented):
a bit of comments for later 👉 👈

packages/audiodocs/docs/system/recording-notification-manager.mdx
```diff
@@ -97,12 +123,16 @@ Adds an event listener for notification actions.
 interface RecordingNotificationInfo {
   title?: string;
   contentText?: string;
-  paused?: boolean; // flag indicating whether to display pauseIcon or resumeIcon
+  paused?: boolean; // flag indicating whether to display the pause or the resume action
```
michalsek:
```suggestion
  paused?: boolean;
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioRecorderCallback.h
maciejmakowski2003: :hurtrealbad:

apps/common-app/src/demos/Record/Record.tsx
maciejmakowski2003: agree

packages/audiodocs/docs/system/recording-notification-manager.mdx
```diff
@@ -97,12 +123,16 @@ Adds an event listener for notification actions.
...
   smallIconResourceName?: string;
   largeIconResourceName?: string;
-  pauseIconResourceName?: string;
-  resumeIconResourceName?: string;
-  color?: number; //
+  color?: number;
+  showStopAction?: boolean; // shows the native stop action, default: false
```
maciejmakowski2003:
```suggestion
  showStopAction?: boolean;
```

apps/common-app/src/demos/Record/Record.tsx
```diff
@@ -254,22 +295,50 @@ const Record: FC = () => {
...
       resumeListener.remove();
-      RecordingNotificationManager.hide();
+      stopListener.remove();
     };
-  }, [onPauseRecording, onResumeRecording]);
+  }, [onPauseRecording, onResumeRecording, onStopRecordingFromNotification]);
+
+  // An ongoing recording is picked up by the state initializer above; here we only
+  // collect the files of a recording that was stopped natively (notification stop
+  // action) while this screen was unmounted.
```
maciejmakowski2003:
```suggestion
  // Collect the files of a recording that was stopped natively while this screen was unmounted.
```

apps/common-app/src/demos/Record/RecordingTime.tsx
```diff
@@ -1,71 +1,53 @@
...
+  }, [state]);
 
-  return (
-    <AnimatedTextInput
-      editable={false}
-      animatedProps={animatedText}
-      style={styles.text}
-    />
-  );
+  return <Text style={styles.text}>{durationString}</Text>;
```
maciejmakowski2003: why we moved to state driven timer?

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/system/NativeRecorderControl.h
maciejmakowski2003, line 1: .hpp ?

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/notification/RecordingNotification.kt
```diff
@@ -29,196 +24,231 @@ class RecordingNotification(
...
+      filter,
+      ContextCompat.RECEIVER_NOT_EXPORTED,
+    )
+    state.initialized = true
   }
 
   override fun show(options: ReadableMap?): Notification {
     initializeNotification()
     val context = reactContext.get() ?: throw IllegalStateException("React context is null")
-    if (options != state.cachedRNOptions) {
```
maciejmakowski2003, line 84: why it is deleted?

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/notification/RecordingNotificationReceiver.kt
```diff
@@ -6,30 +6,110 @@ import android.content.Intent
...
+          } else {
+            NativeRecorderControl.resumeActiveRecording()
+          }
+        // `false` means no recording was in a state this action applies to, so neither
+        // the notification look nor JS may flip.
+        if (toggled) {
+          MediaSessionManager.setRecordingNotificationPaused(paused)
+          dispatchEventToJs(
+            if (paused) AudioEvent.RECORDING_NOTIFICATION_PAUSE else AudioEvent.RECORDING_NOTIFICATION_RESUME,
+          )
```
maciejmakowski2003: it does not look great

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/notification/RecordingNotificationReceiver.kt
```diff
@@ -6,30 +6,110 @@ import android.content.Intent
...
+        pendingResult.finish()
+      }
+    }
+  }
+
+  /** Syncing a live JS runtime is best-effort — in the task-removed scenario the JNI
+   * dispatch can throw, and that must not undo the native work that already completed. */
+  private fun dispatchEventToJs(event: AudioEvent) {
+    try {
+      module.invokeHandlerWithEventNameAndEventBody(event.ordinal, mapOf())
```
maciejmakowski2003, line 98: let's use repo-wide audio API module methods invoking convention

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.h
maciejmakowski2003, line 1: let's rework it to return a state after each method:

```cpp
RecorderState pauseActiveRecording();
RecorderState resumeActiveRecording();
RecorderState stopActiveRecording();
```

thanks to that approach the notification can become a pure function of a state the recorder, the notification becomes a pure function of a state the recorder

maciejmakowski2003 (changes requested):
a few issues identified by claude. worth to check.

- RecordingNotificationReceiver.kt:90 — stopRecordingNatively ignores the stopActiveRecording() return value, so it hides the notification and dispatches recordingNotificationStop even when nothing was stopped. When the handle's slot is empty but a recording is live, this cancels the notification and unsubscribes the mic-typed foreground service while the Oboe stream keeps capturing — no notification, no way to stop. When the stop failed at closeFile(), JS gets a stop event and a null result, silently losing the recording.

- AudioRecorderHostObject.cpp:35 — the constructor unconditionally overwrites ActiveRecorderHandle's single slot. A second, never-started AudioRecorder displaces a recording one, and when that second one is GC'd, clearRecorder sees an identity match and empties the slot outright — the live recording becomes permanently unreachable to every notification action, and isRecordingOngoing() reports false. The one-recorder assumption is documented, but the failure mode is worse than the doc implies.

- AndroidAudioRecorder.cpp:253 — stop() is documented "JS thread only" and mutates recordingSegmentPaths_/filePath_ after releasing stopLock; the notification path now calls it from the receiver's executor. If the user starts a new recording while the old stop is blocked in closeFile(), the executor's trailing clear() / filePath_ = "" wipes the new session's path and its later stop() returns no files.

- ForegroundServiceManager.kt:65 — onServiceDestroyed() clears isServiceRunning unconditionally. A hide-then-show in one tick (recording → playback notification) lets the old instance's onDestroy clear the flag after the restart, and the next unsubscribe then never stops the service — it leaks with no notification behind it.

- RecordingNotification.kt:235 — paused resets to false when absent while every other option is sticky. A partial show({ contentText: … }) during a natively-paused recording flips the button back to Pause; tapping it no-ops (pauseActiveRecording() returns false), so the notification is stuck and the user can't resume from it.

- Record.tsx:136 — demo decodes paths[0] only after concatAudioFiles was dropped; with rotateIntervalBytes (which the new docs recommend for exactly this scenario) all segments but the first are silently discarded.

could you please clean-up comments as well.

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/notification/NotificationRegistry.kt
```diff
@@ -81,30 +90,68 @@ class NotificationRegistry(
...
+      activeNotifications[key] = false
+      ForegroundServiceManager.unsubscribe(notification)
+      Log.d(TAG, "Hiding notification: $key (unsubscribed from foreground service)")
     }
   }
 
+  /**
+   * Hide a notification by its Android notification ID.
+   * Used by native-initiated flows (e.g. the recording stop action) that don't know
+   * the JS-chosen key.
```
SomePersonFromMars: Should a callee know their callers?

```suggestion
```

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/notification/NotificationRegistry.kt
```diff
@@ -81,30 +90,68 @@ class NotificationRegistry(
...
+  fun hideNotification(id: Int) {
+    notifications.entries
+      .firstOrNull { it.value.getNotificationId() == id }
+      ?.let { hideNotification(it.key) }
+  }
+
+  /**
+   * Rebuild and re-post the recording notification with a new paused state.
+   * Used by native-initiated pause/resume so the action button flips even when JS
+   * is unreachable. No-op unless the recording notification is currently visible —
```
SomePersonFromMars: Like above

```suggestion
   * No-op unless the recording notification is currently visible —
```

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/notification/RecordingNotification.kt
```diff
@@ -29,196 +24,231 @@ class RecordingNotification(
...
+    return buildNotification(context)
+  }
 
-    if (state.smallIconResourceName != null) {
-      builder.setSmallIcon(context.resources.getIdentifier(state.smallIconResourceName, "drawable", context.packageName))
-    }
+  /**
+   * Rebuilds with an updated paused flag, leaving the sticky RN options untouched.
+   * Used by native-initiated pause/resume so the action button flips even when JS
+   * is unreachable.
```
SomePersonFromMars:
```suggestion
```

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.cpp
```diff
@@ -0,0 +1,100 @@
...
+    recorder_.reset();
+  }
+}
+
+bool ActiveRecorderHandle::isRecordingOngoing() {
+  std::shared_ptr<AudioRecorder> recorder;
+  {
+    std::scoped_lock lock(mutex_);
+    recorder = recorder_.lock();
+  }
```
SomePersonFromMars: This fragment may be extracted to a helper function `inline std::shared_ptr<AudioRecorder> lockRecorder()`, because it is repeated in almost all methods.

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.h
```diff
@@ -0,0 +1,67 @@
...
+/// recording-notification actions arrive through static JNI with no React context to
+/// walk back to that object — a weak one-slot handle is the minimal bridge that lets
+/// them control the live recorder. On top of native notification control it stashes,
+/// consume-once, the file info of a recording finalized natively while no JS promise
+/// or listener was waiting, and lets a remounted UI seed its state from the native
+/// source of truth via isRecordingOngoing().
+///
+/// Assumes at most one AudioRecorder is alive at a time; setting a new recorder replaces
+/// the previous one.
+class ActiveRecorderHandle {
```
SomePersonFromMars, line 36: Given it is a singleton, the constructor should be private.
```suggestion
class ActiveRecorderHandle {
private:
    ActiveRecorderHandle() = default;
```

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.h
SomePersonFromMars: According to https://en.cppreference.com/cpp/memory/weak_ptr/lock, std::weak_ptr::lock() is an atomic operation, therefore it is unnecessary to lock `mutex_`, when locking `recorder_`.

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.cpp
SomePersonFromMars: Or even better don't use the mutex at all. See https://github.com/software-mansion/react-native-audio-api/pull/1234/changes#r3903505227.

SomePersonFromMars (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.h
closetcaiman: Sure, I had in mind the 'trip' between layers (JNI, etc.) which is not feasible here I'm afraid.

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.cpp
mdydek: we can omit usage of mutex to retrieve shared_ptr from weak_ptr, but we still need it to properly synchronize majority of the calls with the possible clear function being called

SomePersonFromMars: Of course, that's right. I only meant to omit the usage of mutex to retrieve shared_ptr from weak_ptr.

packages/react-native-audio-api/common/cpp/audioapi/core/utils/AudioRecorderCallback.h
mdydek: because this pr introduces first cpp tests that touch recorder, which creates host object as result of the stopping and tests need some workaround it, but right now this workaround lives in the `AudioBufferHostObject.h` file so you can look there for the explanation

apps/common-app/src/demos/Record/RecordingTime.tsx
mdydek: without any particular reason, reverted

apps/common-app/src/demos/Record/RecordingVisualization.tsx
```diff
@@ -498,20 +348,6 @@ const RecordingVisualization: React.FC<RecordingVisualizationProps> = ({
...
-            <Path
-              path={historyWaveformPath}
-              style="stroke"
-              strokeWidth={constants.historyBarWidth}
-              strokeCap="round"
-              color="#ff6259"
-            />
-          </Group>
-        </Canvas>
-      </View>
```
michalsek: why?

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.cpp
```diff
@@ -0,0 +1,90 @@
...
+    if (current && current.get() != recorder) {
+      return;
+    }
+    recorder_.reset();
+  }
+}
+
+RecorderState ActiveRecorderHandle::currentState() {
+  std::shared_ptr<AudioRecorder> recorder = recorder_.lock();
+  std::scoped_lock lock(destructorMutex_);
```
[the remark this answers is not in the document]
SomePersonFromMars: I see that atomicity of C++ smart pointers is more nuanced, than I thought. As I mentioned before, `weak_ptr::lock()` is executed atomically, but `weak_ptr::reset()` isn't. Theoritically it could be mitigated, by using [`std::atomic<std::weak_ptr>` specialization](https://en.cppreference.com/cpp/memory/weak_ptr/atomic2), but it is [unavailable yet in libc++ shipped by Apple's clang](https://discourse.llvm.org/t/rfc-implementing-std-atomic-shared-ptr-t-and-std-atomic-weak-ptr-t-p0718r2-for-libc/91586). Another potential solution is to use [C++11 free functions for atomic operations](https://en.cppreference.com/cpp/memory/shared_ptr/atomic), which support `shared_ptr`. It wouldn't, however be truly lock-free either, because the very same page says:
> These functions are **typically implemented using mutexes**, stored in a global hash table where the pointer value is used as the key.

To conclude, the only solution is to stick to guarding the `weak_ptr` `recorder_` with a mutex.

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/inputs/AudioRecorderHostObject.cpp
```diff
@@ -31,6 +32,7 @@ AudioRecorderHostObject::AudioRecorderHostObject(
 #else
   audioRecorder_ = std::make_shared<IOSAudioRecorder>(audioEventHandlerRegistry, options);
 #endif
+  ActiveRecorderHandle::global().setRecorder(audioRecorder_);
```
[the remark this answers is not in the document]
SomePersonFromMars: This is right. AudioRecorder ought to be a singleton. Because of that we need a policy to enforce it on the JS side, so that the user cannot create more than one host object at the same time. Maybe throwing an exception?

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/inputs/AudioRecorderHostObject.cpp
```diff
@@ -54,13 +55,20 @@ AudioRecorderHostObject::AudioRecorderHostObject(
...
 JSI_HOST_FUNCTION_IMPL(AudioRecorderHostObject, start) {
   auto fileNameOverride = jsiutils::argToString(runtime, args, count, 0, "");
   auto audioRecorder = audioRecorder_;
 
   return promiseVendor_->createAsyncPromise(
       [audioRecorder, fileNameOverride = std::move(fileNameOverride)]() -> PromiseResolver {
         auto result = audioRecorder->start(fileNameOverride);
+        if (result.is_ok()) {
+          ActiveRecorderHandle::global().setRecorder(audioRecorder);
+        }
```
SomePersonFromMars, line 71: Currently, when there are two host objects and both try to start, `audioRecorder->start()` actually fails due to the audio stream being opened in `oboe::SharingMode::Exclusive` mode. It is not obvious from `audioRecorder->start()` semantics and it might change in the future. We should only set a recorder only when the handle is cleared.

SomePersonFromMars: In order to fix it, I suggest adding `trySetRecorder()` to `ActiveRecorderHandle`. It should atomically (the simplest solution is to achieve it via the mutex) verify, whether the recorder is unset and if it is, it should set it, otherwise return false.

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.h
```diff
@@ -0,0 +1,79 @@
...
+  /// and clears it, or std::nullopt when nothing is stashed.
+  std::optional<RecordingStopResult> takeLastRecordingResult();
+
+ private:
+  ActiveRecorderHandle() = default;
+  friend struct ActiveRecorderHandleTestPeer;
+
+  static RecorderState stateOf(const std::shared_ptr<AudioRecorder> &recorder);
+
+  std::mutex destructorMutex_;
```
SomePersonFromMars, line 74: Maybe renaming it simply to `mutex_`? It's job is to simply make the whole structure synchronized.

SomePersonFromMars (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.cpp
```diff
@@ -0,0 +1,90 @@
...
+void ActiveRecorderHandle::clearRecorder(const AudioRecorder *recorder) {
+  std::shared_ptr<AudioRecorder> current;
+  {
+    std::scoped_lock lock(destructorMutex_);
+    current = recorder_.lock();
+    if (current && current.get() != recorder) {
+      return;
+    }
+    recorder_.reset();
+  }
```
closetcaiman: Why do we need scope block here?

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.cpp
```diff
@@ -0,0 +1,90 @@
...
+        RecordingStopResult{.paths = std::move(paths), .size = size, .duration = duration};
+  }
+  return stateOf(recorder);
+}
+
+std::optional<RecordingStopResult> ActiveRecorderHandle::takeLastRecordingResult() {
+  std::scoped_lock lock(destructorMutex_);
+  return std::exchange(lastResult_, std::nullopt);
+}
+
```
closetcaiman, line 112: This read is destructive. I'd consider naming the method accordingly like `pop*` or `consume*`.

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.cpp
```diff
@@ -0,0 +1,90 @@
...
+    std::scoped_lock lock(destructorMutex_);
+    current = recorder_.lock();
+    if (current && current.get() != recorder) {
+      return;
+    }
+    recorder_.reset();
+  }
+}
+
+RecorderState ActiveRecorderHandle::currentState() {
```
closetcaiman: Shouldn't this be a `const` member?

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/sources/AudioBufferHostObject.h
```diff
@@ -1,11 +1,36 @@
...
+  explicit AudioBufferHostObject(const std::shared_ptr<AudioBuffer> &audioBuffer)
+      : audioBuffer_(audioBuffer) {}
+
+  [[nodiscard]] size_t getSizeInBytes() const {
+    return audioBuffer_->getSize() * audioBuffer_->getNumberOfChannels() * sizeof(float) * 2;
+  }
+};
+
+} // namespace audioapi
+
```
closetcaiman: Is this necessary here? Shouldn't this be mocked in the test setup?

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.h
```diff
@@ -0,0 +1,79 @@
...
+/// or listener was waiting, and lets a remounted UI seed its state from the native
+/// source of truth via isRecordingOngoing().
+///
+/// Every control method answers with the recorder state it leaves behind, so a caller
+/// that has no other view of the recorder — the notification after task removal — can
+/// render itself as a pure function of that state instead of tracking its own.
+///
+/// Several AudioRecorder instances can be alive at once (e.g. one per screen), but only
+/// one may occupy this slot
+class ActiveRecorderHandle {
```
closetcaiman, line 28: Given this is a singleton, shouldn't we remove copy/move constructors?

closetcaiman (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.cpp
mdydek: we dont

packages/react-native-audio-api/ios/audioapi/ios/AudioAPIModule.mm
```diff
@@ -123,6 +124,12 @@ - (dispatch_queue_t)methodQueue
         resolve reject : (RCTPromiseRejectBlock)reject)
 {
   dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    // do not disable session if the recording is ongoing, as it will break the recording
+    if (!enabled && ActiveRecorderHandle::global().isRecordingOngoing()) {
+      resolve(nil);
+      return;
+    }
```
maciejmakowski2003: are you sure? I guess it is user intention to do it when calling that method

packages/react-native-audio-api/ios/audioapi/ios/core/IOSAudioRecorder.mm
```diff
@@ -506,27 +506,15 @@ static void cleanupStartedRecorder(
...
-          "Failed to open file for writing: recorder input format is unavailable");
-    }
-
-    auto writerResult = setupFileWriter(properties);
-    if (writerResult.is_err()) {
-      fileOutputEnabled_.store(false, std::memory_order_release);
-      return Result<NoneType, std::string>::Err(writerResult.unwrap_err());
-    }
-  }
-
```
maciejmakowski2003, line 529: why?

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
```diff
@@ -591,8 +587,17 @@ void AndroidAudioRecorder::onErrorAfterClose(oboe::AudioStream *stream, oboe::Re
       return;
     }
 
+    const auto stateBeforeTeardown = state_.load(std::memory_order_acquire);
+
     cleanup();
 
+    // An idle session has nothing to restore — this covers a disconnect delivered
+    // late, after stop() already finished — and reopening here would leave a fresh,
+    // never-started mic stream held while idle.
```
maciejmakowski2003: sloppy comment

maciejmakowski2003: it is trivial

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
```diff
@@ -610,8 +615,13 @@ void AndroidAudioRecorder::onErrorAfterClose(oboe::AudioStream *stream, oboe::Re
       return;
     }
 
-    mStream_->requestStart();
-    state_.store(RecorderState::Recording, std::memory_order_release);
+    // Restore the interrupted session's state instead of unconditionally recording —
+    // a paused session must stay paused, or the reopened stream would silently turn
+    // the microphone back on against an explicit user action.
```
maciejmakowski2003: trivial

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/NativeRecorderControl.kt
maciejmakowski2003, line 1: a lot of sloppy trivial comments

maciejmakowski2003: maybe just refer to cpp counterpart

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.h
maciejmakowski2003, line 1: recheck comments in both .h and .cpp file and use singleton pattern instead of global method

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/AudioRecorder.h
```diff
@@ -61,6 +61,10 @@ class AudioRecorder {
   virtual bool isPaused() const = 0;
   virtual bool isIdle() const = 0;
 
+  /// @brief The state the platform predicates above agree on, which can differ from the
+  /// last requested transition while the audio engine is not running.
```
maciejmakowski2003: trivial

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/inputs/AudioRecorderHostObject.cpp
```diff
@@ -87,8 +95,13 @@ JSI_HOST_FUNCTION_IMPL(AudioRecorderHostObject, stop) {
   return promiseVendor_->createAsyncPromise([audioRecorder]() -> PromiseResolver {
     auto result = audioRecorder->stop();
 
-    return [result =
-                std::move(result)](jsi::Runtime &runtime) -> std::variant<jsi::Value, std::string> {
+    if (result.is_ok()) {
+      ActiveRecorderHandle::global().clearRecorder(audioRecorder.get());
+    }
+
+    using returnValue = std::variant<jsi::Value, std::string>;
```
maciejmakowski2003, line 96: do we really need to declare that type if we use it only once?

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/notification/NotificationRegistry.kt
```diff
@@ -81,30 +90,65 @@ class NotificationRegistry(
...
+      notification.hide()
     } catch (e: Exception) {
       Log.e(TAG, "Error hiding notification $key: ${e.message}", e)
+    } finally {
+      // Even when hide() throws (e.g. the React context was already released), the
+      // registry must record the notification as inactive and let the foreground
+      // service unwind — otherwise it runs forever.
+      activeNotifications[key] = false
+      ForegroundServiceManager.unsubscribe(notification)
+      Log.d(TAG, "Hiding notification: $key (unsubscribed from foreground service)")
```
maciejmakowski2003, line 104: needed?

maciejmakowski2003 (changes requested):
could you test what happens if user dismisses notification. strip off or edit comments added by ai. verify logs usefulness.

maciejmakowski2003:
claude code review findings - could 

Bugs

1. enableFileOutput can break a recording that's starting (Android and iOS). High.
AndroidAudioRecorder.cpp:289 and IOSAudioRecorder.mm:509 now check isIdle() before taking fileWriterMutex_. start() runs on the promise worker thread and holds that mutex, and it only sets Recording at the very end. So this can happen:

JS calls start(), then enableFileOutput().
The isIdle() check passes and enableFileOutput waits on the lock.
start() finishes and sets up the writer.
enableFileOutput then sets fileOutputConfigured_ = false in the middle of the session.

After that, usesFileOutput() is false: onAudioReady stops writing, and stop() never moves or closes fileWriter_, so the file is truncated and never finalized. On Android, onErrorAfterClose briefly sets the state to Idle between cleanup() and reopening the stream, which opens the same window.

Fix: do the isIdle() check after taking the lock.
The doc comment above the Android enableFileOutput still says it "will prepare and open the file for writing immediately" when active, which is no longer true.
Returning Ok while ignoring the properties is silent. Either return Err or document it.

2. ActiveRecorderHandle holds its mutex through recorder->stop(). Medium.
stop() waits until the file is closed and finalized, and pause/resume hold the same mutex. Callers on the JS thread block on it, so tapping Stop in the notification freezes JS until the file is finalized. Those callers:

the isRecordingOngoing() JSI global
~AudioRecorderHostObject (runs during GC)
setRecorder / clearRecorder in the start/stop promise lambdas

Also, recorder is declared after lock, so it's destroyed while the lock is still held. If the handle's copy is the last owner, the recorder's destructor (stream close, file close) runs under the mutex. clearRecorder avoids this on purpose, but the other methods don't.

Fix: copy the shared_ptr under the lock, unlock, do the work, then relock only to write lastResult_.

3. An empty slot counts as IDLE, so tapping Pause/Resume hides the notification. Medium.
RecordingNotificationReceiver.renderNotification(IDLE) calls hideRecordingNotification(), which also unsubscribes the foreground service. The slot is empty in several ordinary cases:

Before start() resolves: setRecorder only runs at the end of the async lambda. If the notification is shown first and the user taps Pause in that window, the notification and the microphone foreground service go away while recording is starting.
After a JS stop(): if the app keeps the notification up, the slot is already cleared.
After GC: the recorder's host object was collected.

Stop on an empty slot also sends RECORDING_NOTIFICATION_STOP to JS even though nothing was stopped. Consider a separate "no recorder" result, and only hide on the stop action.

4. Having more than one recorder isn't enforced. Low to medium.
If recorder A is recording and recorder B starts, B takes the slot. When B stops, clearRecorder(B) empties the slot while A is still recording. Then isRecordingOngoing() returns false, iOS setActive(false) deactivates the session under A, and the notification controls do nothing. The header documents "only one may occupy this slot", but nothing stops it happening.

5. After a JS reload, notification events may go to the old runtime. Low, not verified.
The receiver keeps the AudioAPIModule it was created with in initializeNotification. AudioAPIModule.invalidate() doesn't reset the hybrid, and AudioEventHandlerRegistry keeps a raw jsi::Runtime*. So in the "fresh JS context resyncs" case this branch targets, pause/resume/stop events likely reach the old module instead of the new one. If the old runtime is gone, that could be a native crash, which the Kotlin try/catch in dispatchEventToJs can't catch. Worth testing by reloading while the notification is visible.

6. iOS setAudioSessionActivity(false) does nothing during a recording. Low.
It resolves nil without deactivating, so JS thinks it worked and nothing deactivates the session later. Rejecting, or at least documenting it, would be clearer.

Nits
destructorMutex_ guards all of the handle's state, not just destruction. mutex_ would be a more accurate name.
RecordingStopResult::size and duration are uninitialized. Use double size{} and double duration{}.
options !== state.cachedRNOptions: every JS call creates a new ReadableMap, so this check never skips anything. You can delete cachedRNOptions.

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/inputs/AudioRecorderHostObject.cpp
mdydek: I wanted formatting to behave correctly :)

packages/react-native-audio-api/ios/audioapi/ios/core/IOSAudioRecorder.mm
mdydek: this changed the behavior that we cannot suddenly enable file output in ongoing recorder, thus we don't need that code, we can discuss it internally if we would like to support this, if yes we can revert that

packages/react-native-audio-api/common/cpp/audioapi/core/inputs/ActiveRecorderHandle.h
mdydek: I do not agree with you, because singleton means that it has to be created somewhere and each consecutive call to constructor would return it, it serves slight different purpose

packages/react-native-audio-api/ios/audioapi/ios/AudioAPIModule.mm
mdydek: your right, reverted now disabling session stops the recording first

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/notification/NotificationRegistry.kt
mdydek: it could leave the notification hanging otherwise

SomePersonFromMars (approved):

maciejmakowski2003 (approved):
LGTM

PR #1239, opened 2026-08-18
refactor: split the context state to render and control threads

mdydek:
### Only js state:
resume() flips the single state to RUNNING synchronously. The actual driver open is async and takes a few ms (Oboe/AVAudioSession — normal, not a failure). JS calls suspend() in that window. The guard reads RUNNING (not SUSPENDED), so it correctly decides to call audioPlayer_->suspend() — but the stream hasn't finished opening yet, so there's nothing real to stop. The stop call lands on a half-initialized stream (no-op, or worse, undefined behavior depending on the SDK); the open then completes moments later and audio starts playing — after suspend() already resolved and JS believes the context is suspended.

### Only audio-thread state:
JS calls ctx.close() twice back-to-back (e.g. a cleanup effect firing twice). The "already closed" guard reads the engine-confirmed value, which hasn't updated yet because the engine hasn't processed either call. Both calls pass the guard and both proceed to release the same driver resources — double-free/double-release.

### Solution

Two unrelated states, one serves as the guard for the unlogic calls, the second one guards internal audio engine to not call two things at the same time

### DIFF
--- a/packages/react-native-audio-api/src/core/BaseAudioContext.ts
+++ b/packages/react-native-audio-api/src/core/BaseAudioContext.ts
@@ -48,12 +48,13 @@ export default class BaseAudioContext {
     return this.contextState;
   }
 
-  /* @internal Called by AudioScheduledSourceNode.start(). */
-  public markRunningOnSourceStart(): void {
-    if (this.contextState === 'suspended') {
-      this.contextState = 'running';
-    }
-  }
+  /**
+   * @internal Called by AudioScheduledSourceNode.start(). No-op here: only
+   * AudioContext overrides it, since only AudioContext's native driver can
+   * start implicitly from a source's start() call. OfflineAudioContext only
+   * starts rendering from an explicit startRendering() call.
+   */
+  public markRunningOnSourceStart(): void {}
 
   public async decodeAudioData(
     input: DecodeDataInput,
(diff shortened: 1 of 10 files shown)


packages/react-native-audio-api/src/core/BaseAudioContext.ts
```diff
@@ -38,14 +38,24 @@ export default class BaseAudioContext {
     this.sampleRate = context.sampleRate;
   }
 
+  protected contextState: ContextState = 'suspended';
```
closetcaiman: nit:

I think it be just `state` or if we want to be really explicit `controlThreadState`.

packages/react-native-audio-api/src/core/OfflineAudioContext.ts
```diff
@@ -80,9 +77,11 @@ export default class OfflineAudioContext extends BaseAudioContext {
...
     }
 
-    this.isSuspended = true;
-
-    return (this.context as IOfflineAudioContext).suspend(suspendTime);
+    const result = await (this.context as IOfflineAudioContext).suspend(
+      suspendTime
+    );
+    this.contextState = 'suspended';
+    return result;
```
closetcaiman, line 88: What is the intended behaviour if the promise is rejected? Should the `state` be preserved or always changed to `suspended`?

mdydek: it should be preserved and it works like that currently

closetcaiman (approved):

PR #1099, opened 2026-06-11
build: bump typescript

closetcaiman:
## ⚠️ Breaking changes ⚠️

- 

## Introduced changes

- bumped `typescript` version to `6.0.3`
   - resolved version 6 transpiler warnings (https://github.com/microsoft/TypeScript/issues/62508): 
     - `tsconfig` issues
     - `global` keyword -> `globalThis`
     - `Float32Array` -> `Float32Array<ArrayBuffer>`
  - resolved `vscode` specific `jest` intellisense errors
- refactored `AudioBufferSourceNode.web.ts`
  - moved `wasm` specific code to `/custom`
  - created uniform interface for the `backend` node under the `AudioBufferSourceNode.web.ts` wrapper
  - removed unused code
- moved all `jsi` specific typing into `jsi-interfaces.ts` (the `I*` ones)

## Checklist

- [x] Linked relevant issue
- [x] Updated relevant documentation
- [x] Added/Conducted relevant tests
- [x] Performed self-review of the code
- [x] Updated Web Audio API coverage
- [x] Added support for web
- [x] Updated old arch android spec file

PR #1242, opened 2026-08-19
build: update metro config

closetcaiman:
Closes #

## ⚠️ Breaking changes ⚠️

- N/A

## Introduced changes

- updated `fabric-example/metro.config.js` so that it overrides the consumer resolution to transpiled `JS` files and instead reads source `TS` (metro handles `TS` on the fly) thus allowing for live changes (follows-up changes to package build in #1194 and #1222)
- removed deprecated `baseUrl` field from `fabric-example/tsconfig.json` (kind of stale follow-up of #1099)

## Checklist

- [x] Linked relevant issue
- [x] Added support for web


mdydek (changes requested):
`react-native-audio-worklets`?

mdydek (approved):

PR #1246, opened 2026-08-20
fix: proper gc in the wpt tests

mdydek:
Closes #

The pr kind of evolved, it aimed to eliminate exit code 129 in the ci, but it introduced a few features in between

## Introduced changes

- new context event: `onstatechange` that fires when the context state has changed
- `oncomplete` event in offlineaudiocontext
- wpt side improvements, added batching test files
- several new state variables introduced:
    - `state_` in c++ is the engine truth
    - `publishedState_` in c++ is the js state atribute
    - `lastDispatchedState_` in c++ used to compare whether state has truly changed its value from previous event call, decides whether to call the `onstatechange` event
    - `_state` in js, it serves as immediate guard in the js layer so `close` after `resume` can see `running` temporary js state and it can be queued, it does not wait for true resume to take place

spec is kind of not very clear what exactly each state means: https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-rendering-thread-state-slot, the design has come from the wpt tests that serve as intended way of behaving

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/src/core/AudioBufferSourceNode.ts
```diff
@@ -112,25 +114,31 @@ export default class AudioBufferSourceNode extends AudioBufferBaseSourceNode {
...
     (this.node as IAudioBufferSourceNode).start(when, offset, duration);
+    this.context.markRunningOnSourceStart();
   }
 
   public get onLoopEnded(): ((event: EventEmptyType) => void) | undefined {
     return this.onLoopEndedCallback;
   }
 
   public set onLoopEnded(callback: ((event: EventEmptyType) => void) | null) {
+    // See the note in AudioScheduledSourceNode.onEnded.
```
closetcaiman: what note?

packages/react-native-audio-api/tests/audio-event-subscriptions.test.ts
closetcaiman, line 1: What is the purpose of this? The suite uses some ad hoc mocks. While I am for the better test coverage this seems out of place to me.

packages/react-native-audio-api/src/core/AudioBufferSourceNode.ts
mdydek: leftover

packages/react-native-audio-api/src/core/BaseAudioContext.ts
```diff
@@ -38,14 +43,62 @@ export default class BaseAudioContext {
...
+    if (this.publishedState === nextState) {
+      return;
+    }
+
+    this.publishedState = nextState;
+    setTimeout(() => {
+      this.onstatechange?.({ type: 'statechange', target: this });
+    }, 0);
+  }
+
```
closetcaiman: Isn’t this a JS-side executed `onstatechange`? Shouldn't that be routed and registered into `eventRegistry`?

packages/react-native-audio-api/src/core/BaseAudioContext.ts
```diff
@@ -36,16 +43,87 @@ export default class BaseAudioContext {
...
   }
 
+  /**
+   * The spec's [[control thread state]]: written synchronously the moment an
+   * operation is accepted, so the NEXT call validates against what has already
+   * been requested (e.g. close() right after resume() must see 'running').
+   * Never exposed — the `state` attribute reports acknowledged reality
+   * instead.
+   */
   protected _state: ContextState = 'suspended';
```
closetcaiman, line 53: Is this field actually needed if we moved the getter to the host object?

packages/react-native-audio-api/src/core/AudioContext.ts
```diff
@@ -33,26 +33,26 @@ export default class AudioContext extends BaseAudioContext {
...
 
   async resume(): Promise<undefined> {
     if (this._state === 'closed') {
       throw new InvalidStateError('Cannot resume a closed audio context.');
     }
 
-    this._state = 'running';
-    return (this.context as IAudioContext).resume();
+    this.setControlState('running');
+    await (this.context as IAudioContext).resume();
```
closetcaiman: If the promise rejects the `[[control state]]` stays `running`.

packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.cpp
```diff
@@ -70,10 +70,10 @@ bool AudioContext::tryStartDriver() {
...
   if (audioPlayer_->start()) {
     isInitialized_.store(true, std::memory_order_release);
-    // The driver also starts implicitly, from the first
-    // `AudioScheduledSourceNode::start()`. Publish RUNNING here so the visible state
+    // Publish RUNNING here so the visible state
     // never reports SUSPENDED while the graph is actually rendering; `resume()`
-    // reaches the same state through its promise task.
+    // reaches the same state through its promise task (its dispatch then
+    // dedupes against this one).
     setState(ContextState::RUNNING);
```
closetcaiman, line 77: Actually it does not. It only flips the atomic `state` flag. (?)

packages/react-native-audio-api/src/jsi-interfaces.ts
```diff
@@ -43,6 +43,14 @@ export interface IBaseAudioContext {
   readonly currentTime: number;
   readonly decoder: IAudioDecoder;
 
+  /**
+   * Web Audio `AudioContextState` string; native storage behind the JS `state`
+   * attribute.
+   */
+  readonly state: string;
+  // passing subscriptionId(uint_64 in cpp, string in js) to the cpp
+  onstatechange: string;
```
maciejmakowski2003, line 48: not aligned with repo listeners naming convention

packages/react-native-audio-api/src/jsi-interfaces.ts
```diff
@@ -43,6 +43,14 @@ export interface IBaseAudioContext {
   readonly currentTime: number;
   readonly decoder: IAudioDecoder;
 
+  /**
+   * Web Audio `AudioContextState` string; native storage behind the JS `state`
+   * attribute.
+   */
+  readonly state: string;
+  // passing subscriptionId(uint_64 in cpp, string in js) to the cpp
```
maciejmakowski2003: ai slop comments

packages/react-native-audio-api/src/core/BaseAudioContext.ts
mdydek: because we have to have a way to instantly reject some wrong request at the js layer, then the request goes to cpp and then we split the state into the two parts again, provided in this pr: https://github.com/software-mansion/react-native-audio-api/pull/1239

packages/react-native-audio-api/src/core/BaseAudioContext.ts
```diff
@@ -36,16 +43,87 @@ export default class BaseAudioContext {
...
+   * resume()+suspend() pair acknowledges both before either event lands — a
+   * write here would roll the attribute back to the stale event's value. When a
+   * native-originated state with no acknowledging promise arrives (the planned
+   * `interrupted`), its attribute update must be added here explicitly.
+   * Subclasses extend this to order dependent events (offline `complete`) after
+   * `statechange`.
+   */
+  protected onNativeStateChange(_event: OnStateChangeEventType): void {
+    this.onstatechangeCallback?.({ type: 'statechange', target: this });
+  }
```
closetcaiman: What purpose does this serve? Maybe I don't understand the flow correctly, but it seems like unecessary (?) callback proxy. Please elaborate in PR desc if possible.

packages/react-native-audio-api/src/core/BaseAudioContext.ts
```diff
@@ -36,16 +43,87 @@ export default class BaseAudioContext {
...
+    this.context.onstatechange = this.stateChangeSubscription.subscriptionId;
   }
 
+  /**
+   * The spec's [[control thread state]]: written synchronously the moment an
+   * operation is accepted, so the NEXT call validates against what has already
+   * been requested (e.g. close() right after resume() must see 'running').
+   * Never exposed — the `state` attribute reports acknowledged reality
+   * instead.
+   */
```
closetcaiman, line 52: I'd trim down the comments here and there and reword them so that the message is conveyed in a cleaner way instead of describing the whole state-machine flow in one go.

closetcaiman (changes requested):

closetcaiman (approved):

maciejmakowski2003 (changes requested):

packages/react-native-audio-api/src/jsi-interfaces.ts
mdydek: https://github.com/software-mansion/react-native-audio-api/pull/1286

packages/react-native-audio-api/common/cpp/audioapi/core/types/ContextState.h
```diff
@@ -6,4 +6,18 @@ namespace audioapi {
...
+  switch (state) {
+    case ContextState::SUSPENDED:
+      return "suspended";
+    case ContextState::RUNNING:
+      return "running";
+    case ContextState::CLOSED:
+      return "closed";
+  }
+  return "suspended";
 }
```
maciejmakowski2003: I guess it is not a right place for it

packages/react-native-audio-api/src/core/AudioContext.ts
```diff
@@ -33,37 +38,66 @@ export default class AudioContext extends BaseAudioContext {
...
+    this.setControlState(nextState);
+
+    try {
+      return await nativeTransition();
+    } catch (error) {
+      if (this._transitionSeq === mySeq) {
+        this.setControlState(this.state);
+      }
+      throw error;
     }
```
maciejmakowski2003: I would try to solve it easier. it is not very frequent operation

maciejmakowski2003 (approved):

closetcaiman (approved):

PR #1248, opened 2026-08-21
Fix/repair timeout wpt tests

mdydek:
Closes #

## Introduced changes

- `timeToSampleFrame` now rounds to the nearest frame instead of truncating, which landed boundaries one frame late for inputs like 0.03 × 44100.
- Queue capacity in `Constants.h` raised from 64 to 256 events, with a warning that overflow silently drops events. The host object memory-pressure estimate was
- added `onended` event that acts the same as `onEnded`, latter will be removed in near future for spec completeness
- events now remove their previous native subscription before registering a new one

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/RenderParamEvent.h
```diff
@@ -11,6 +11,15 @@ namespace audioapi {
 /// @brief A RenderParamEvent extends ParamEvent with additional properties and a value calculation
 /// function that can compute the parameter value at any time during the event's active period
 /// based on its type and the current state of the queue.
+///
+/// Times exist in two forms. The inherited startTime/endTime are snapped to the
+/// sample-frame grid by ParamRenderQueue::push and drive ordering and
+/// effect-boundary decisions, so an event scheduled at T takes effect exactly
+/// at frame round(T * sampleRate) — the convention Blink and the WPT reference
+/// use. rawStartTime/rawEndTime keep the times as scheduled and feed the value
```
SomePersonFromMars:
```suggestion
/// at frame round(T * sampleRate)
/// rawStartTime/rawEndTime keep the times as scheduled and feed the value
```

packages/react-native-audio-api/common/cpp/audioapi/dsp/AudioUtils.h
```diff
@@ -7,8 +7,12 @@
 
 namespace audioapi::dsp {
 
+/// Round to the nearest frame, matching Blink and the WPT reference
+/// (audit-util.js). Truncation lands one frame late whenever
```
SomePersonFromMars:
```suggestion
/// Round to the nearest frame. Truncation lands one frame late whenever
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.cpp
SomePersonFromMars, line 1: This PR clearly introduces a concept of a "snapped" time, which isn't well expressed. The confusing fact is introduction of "normal" and "raw" time. I think that "normal" time should keep the original semantics (the semantics of "raw" introduced here). The "raw" time shouldn't exist as it is not obvious what it means. It should be replaced with a "snapped" time.

The question is, where does snapped time live? It would be easiest to put it into `RenderParamEvent`. It would be convenient to calculate it in the constructor, but such a value is calculated using `ParamRenderQueue::sampleRate_` using `ParamRenderQueue::snapToSampleFrameTime()`.

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.cpp
```diff
@@ -22,15 +30,35 @@ std::optional<float> ParamRenderQueue::computeValueAtTime(double time) {
...
+
+  // Ramps stay active until their raw end so sub-frame intervals still
+  // interpolate; other finite events end on their snapped boundary. SetTarget
+  // never ends on its own.
+  double effectiveEndTime = event.isRampType() ? event.getRawEndTime() : event.getEndTime();
+  if (event.getType() != ParamEventType::SET_TARGET && time >= effectiveEndTime) {
+    return event.getEndValue();
+  }
+
+  return event.getCalculateValue()(
```
SomePersonFromMars: I know that it's been here for a while, but now I see that the first four parameters are always

```cpp
event.getRawStartTime(),
event.getRawEndTime(),
event.getStartValue(),
event.getEndValue()
```

Why not introduce `event.calculateValueAt(time)`?

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/RenderParamEvent.h
```diff
@@ -11,6 +11,15 @@ namespace audioapi {
...
+/// interpolation, which the spec defines on real times: a ramp between two
+/// times inside one frame must still interpolate on those times, not on their
+/// collapsed snapped values.
```
SomePersonFromMars: This comment wouldn't be necessary, if there existed `calculateValueAt(time)` method as suggested here https://github.com/software-mansion/react-native-audio-api/pull/1248/changes/BASE..e196cf0e432171e94f2049ff43094661d0989603#r3969640038

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.cpp
```diff
@@ -22,15 +30,35 @@ std::optional<float> ParamRenderQueue::computeValueAtTime(double time) {
...
       time);
 }
 
 bool ParamRenderQueue::push(RenderParamEvent &&event) {
+  // Keep the scheduled times for interpolation; snap the effect boundaries
+  // onto the sample-frame grid for ordering and comparisons.
+  event.setRawStartTime(event.getStartTime());
+  event.setRawEndTime(event.getEndTime());
+  event.setStartTime(snapToSampleFrameTime(event.getStartTime()));
+  event.setEndTime(snapToSampleFrameTime(event.getEndTime()));
```
SomePersonFromMars: Shortly after creation of `event`, the semantics of "normal" time is changed to "snapped" time.

SomePersonFromMars (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.cpp
```diff
@@ -22,15 +30,35 @@ std::optional<float> ParamRenderQueue::computeValueAtTime(double time) {
...
+  if (event.getType() != ParamEventType::SET_TARGET && time >= effectiveEndTime) {
+    return event.getEndValue();
+  }
+
+  return event.getCalculateValue()(
+      event.getRawStartTime(),
+      event.getRawEndTime(),
+      event.getStartValue(),
+      event.getEndValue(),
       time);
```
SomePersonFromMars: I know that it's been here for a while, but now I see that the first four parameters are always

```cpp
event.getRawStartTime(),
event.getRawEndTime(),
event.getStartValue(),
event.getEndValue()
```

Why not introduce `event.calculateValueAt(time)`?

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderEventFactory.h
```diff
@@ -118,14 +118,35 @@ class ParamRenderEventFactory {
...
+  static RenderParamEvent createEvent(
+      double startTime,
+      double endTime,
+      float startValue,
+      float endValue,
+      std::function<float(double, double, float, float, double)> &&calculateValue,
+      ParamEventType type) {
+    auto event =
+        RenderParamEvent(startTime, endTime, startValue, endValue, std::move(calculateValue), type);
+    setRawTimes(event);
```
SomePersonFromMars: Why setting the raw times twice? First in `RenderParamEvent()` constructor and then in `setRawTimes()`

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.cpp
SomePersonFromMars: More specifically, I see two ways of introducing snapped time instead of the raw time and changing semantics of normal time:
1. By reading the code, I see that snapped time is a function of raw time. It also means that they are always updated in pairs - one never changes without the other. Thanks to it, it is possible not to introduce new fields in `RenderParamEvent` at all: Every time a snapped time is required, one can simply calculate it via `snapToSampleFrameTime()`. Thanks to this approach, the state of an event is not so dependent on it's user - the queue, which makes it slightly cleaner.
2. The previous approach is sligntly more CPU intensive. In order to mitigate it, one can simply cache the snapped time in `RenderParamEvent`, by introducing `snappedStartTime_` and `snappedEndTime_`.

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.h
```diff
@@ -28,6 +33,11 @@ class ParamRenderQueue : public ParamQueueBase<RenderParamEvent> {
 
  private:
   float defaultValue_;
+  float sampleRate_;
+
+  /// @brief Snap a time to the exact time of its nearest sample frame
+  /// (round(time * sampleRate) / sampleRate)
+  [[nodiscard]] double snapToSampleFrameTime(double time) const;
```
SomePersonFromMars, line 39: Simply inline it, so it's slightly faster. Also, the comment wouldn't be necessary.
```suggestion
  /// @brief Snap a time to the exact time of its nearest sample frame
  [[nodiscard]] inline double snapToSampleFrameTime(double time) const;
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.cpp
```diff
@@ -2,17 +2,24 @@
...
+#include <cmath>
 #include <optional>
 #include <utility>
 
 namespace audioapi {
 
+double ParamRenderQueue::snapToSampleFrameTime(double time) const {
+  return std::round(time * sampleRate_) / sampleRate_;
+}
+
```
SomePersonFromMars, line 14: May be inlined as mentioned here https://github.com/software-mansion/react-native-audio-api/pull/1248/changes#r3978739851.

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/RenderParamEvent.h
```diff
@@ -11,6 +11,12 @@ namespace audioapi {
 /// @brief A RenderParamEvent extends ParamEvent with additional properties and a value calculation
 /// function that can compute the parameter value at any time during the event's active period
 /// based on its type and the current state of the queue.
+///
+/// startTime/endTime always hold the times exactly as scheduled. The spec defines
+/// the interpolation formulas on those real times (a ramp between two times inside
+/// one sample frame must still interpolate on them), so calculateValueAtTime uses
+/// them unmodified. Whether an event is in effect at a given frame is decided by
+/// ParamRenderQueue, which snaps these times onto the sample-frame grid on demand.
```
SomePersonFromMars: `startTime`/`endTime` don't live in this module. Also, let's not document what those *aren't*, but instead what they *are*.

Also, this isn't a place to document the snapping behavior as it doesn't live here.
```suggestion
```

packages/react-native-audio-api/common/cpp/audioapi/core/AudioParam.cpp
```diff
@@ -74,15 +74,19 @@ std::shared_ptr<DSPAudioBuffer> AudioParam::processARateParam(int framesToProces
   }
 
   float sampleRate = context->getSampleRate();
-  double timeCache = time;
-  double timeStep = 1.0 / sampleRate;
+  // Evaluate each sample at the exact frame time `frame / sampleRate` instead
+  // of accumulating `time += 1/sampleRate`: accumulation drifts by a few ULPs
+  // per quantum, which is enough to observe a snapped event boundary one
+  // frame late.
```
SomePersonFromMars:
```suggestion
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderEventFactory.h
```diff
@@ -3,24 +3,33 @@
...
 /// @brief A factory for creating RenderParamEvents and resolving their values
 /// based on the current state of the queue.
+///
+/// The calculateValue functions are pure interpolation formulas evaluated on an
+/// event's scheduled times; whether an event is in effect at a given frame is
+/// decided by ParamRenderQueue::computeValueAtTime against those times snapped
+/// to the sample-frame grid. Because a frame's exact time can fall fractionally
+/// outside [startTime, endTime) while the frame still belongs to the event, each
+/// formula must extrapolate gracefully (clamp, or accept a sub-ULP overshoot)
+/// instead of branching on the scheduled boundary.
```
SomePersonFromMars:
```suggestion
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderEventFactory.h
```diff
@@ -106,21 +102,20 @@ class ParamRenderEventFactory {
...
-            auto factor = static_cast<float>(
-                (time - startTime) * static_cast<double>(length - 1) / (endTime - startTime) - k);
-            return dsp::linearInterpolate(values->span(), k, k + 1, factor);
+          if (endTime <= startTime) {
+            return endValue;
           }
 
-          return endValue;
+          // Position in the array based on time progress, clamped so a frame
+          // snapped fractionally outside the raw interval stays in range.
```
SomePersonFromMars:
```suggestion
```

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.h
```diff
@@ -8,9 +8,14 @@ namespace audioapi {
 
 /// @brief A queue for managing audio parameter change events on the audio render thread.
 /// @note The invariant of the queue is that its internal buffer always contains non-overlapping events.
+///
+/// Events store their times exactly as scheduled and interpolate on them (see
+/// RenderParamEvent). Effect boundaries, however, live on the sample-frame grid:
+/// an event scheduled at T takes effect at frame round(T * sampleRate)
```
SomePersonFromMars, line 14:
```suggestion
/// @note Events store their times exactly as scheduled and interpolate on them (see
/// RenderParamEvent). Effect boundaries, however, live on the sample-frame grid:
/// an event scheduled at T takes effect at frame round(T * sampleRate)
```

SomePersonFromMars (changes requested):

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.h
mdydek:
```
The original intent of the inline keyword was to serve as an indicator to the optimizer that inline substitution of a function is preferred over function call, that is, instead of executing the function call CPU instruction to transfer control to the function body, a copy of the function body is executed without generating the call. This avoids overhead created by the function call (passing the arguments and retrieving the result) but it may result in a larger executable as the code for the function has to be repeated multiple times.

Since inline substitution is unobservable in the standard semantics, compilers are free to use inline substitution for any function that's not marked inline, and are free to generate function calls to any function marked inline. Those optimization choices do not change the rules regarding multiple definitions and shared statics listed above.

Because the meaning of the keyword inline for functions came to mean "multiple definitions are permitted" rather than "inlining is preferred" since C++98, that meaning was extended to variables.

(since C++17)
```
[cppreference.com](https://en.cppreference.com/cpp/language/inline)

mdydek: i get that we use it like that in many cases in the repository, I think the difference is negligible, but there is no point in introducing it

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.cpp
mdydek: explained in the linked comment

SomePersonFromMars (approved):

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.cpp
```diff
@@ -2,17 +2,24 @@
...
-#include <cstddef>
+#include <cmath>
 #include <optional>
 #include <utility>
 
 namespace audioapi {
 
+double ParamRenderQueue::snapToSampleFrameTime(double time) const {
+  return std::round(time * sampleRate_) / sampleRate_;
+}
```
maciejmakowski2003: dont we have that helper function somewhere?

maciejmakowski2003 (approved):

maciejmakowski2003 (approved):

PR #1251, opened 2026-08-24
fix: lazy init of audio engine

maciejmakowski2003:
## ⚠️ Breaking changes ⚠️

- Removed `AudioEngine` allocation inside iOS audio-api module installation and restricted `AudioEngine` rebuild - apps that use react-native-audio-api only for session / notification management should not be forced to allocate it.
- Restricted session activation on media services reset - the audio session should not be activated after that notification if it had not been active beforehand.

## Introduced changes

- Fixed typo in `SystemNotificationManager.mm` - `handleEngineConfigurationChange` was not invoked.

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/ios/audioapi/ios/system/SystemNotificationManager.mm
+++ b/packages/react-native-audio-api/ios/audioapi/ios/system/SystemNotificationManager.mm
@@ -222,11 +222,9 @@ static NSString *NotificationManagerContext = @"SystemNotificationManagerContext
   switch (routeChangeReason) {
     case AVAudioSessionRouteChangeReasonNewDeviceAvailable:
     case AVAudioSessionRouteChangeReasonOldDeviceUnavailable:
-    case AVAudioSessionRouteChangeReasonRouteConfigurationChange: {
-    handleEngineConfigurationChange:
-      nil;
+    case AVAudioSessionRouteChangeReasonRouteConfigurationChange:
+      [self handleEngineConfigurationChange:nil];
       break;
-    }
     default:
       break;
   }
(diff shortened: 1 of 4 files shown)


packages/react-native-audio-api/ios/audioapi/ios/system/SystemNotificationManager.mm
```diff
@@ -222,11 +222,9 @@ - (void)handleRouteChange:(NSNotification *)notification
   switch (routeChangeReason) {
     case AVAudioSessionRouteChangeReasonNewDeviceAvailable:
     case AVAudioSessionRouteChangeReasonOldDeviceUnavailable:
-    case AVAudioSessionRouteChangeReasonRouteConfigurationChange: {
-    handleEngineConfigurationChange:
-      nil;
+    case AVAudioSessionRouteChangeReasonRouteConfigurationChange:
+      [self handleEngineConfigurationChange:nil];
```
closetcaiman, line 226: ❤️

closetcaiman, line 226: From what I see this handler does not deactivate the session, so it may lead to stale `isActive` flag (says `false`, but the session is active). Maybe we should add the same steps you added below in `handleMediaServicesReset`?

maciejmakowski2003: do you mean that the session can be active after media session restart?

closetcaiman: No, `onConfigurationChange` fires those two:
```objc
- (void)handleEngineConfigurationChange:(NSNotification *)notification
{
  AudioEngine *audioEngine = self.audioAPIModule.audioEngine;
  AudioSessionManager *sessionManager = self.audioAPIModule.audioSessionManager;

  dispatch_async(dispatch_get_main_queue(), ^{
    [sessionManager markInactive];
    [audioEngine restartAudioEngine];
  });
}
```

Which sets `isActive` to false (markInactive), then if there is no graph or no engine `restartEngine` returns early which leaves the `AudioSession` running, but the flag set to `false`, so `setAudioSessionActivity(false)` is no-op.

maciejmakowski2003: this handler will not be fired if there is no engine, that in the cases also implies no graph

closetcaiman: even through `AVAudioSessionRouteChangeNotification`?

closetcaiman (approved):

PR #1252, opened 2026-08-25
fix: AudioContext::resume reanimates the failed Oboe stream due to the change in AudioPlayer::resume 

### LINKED ISSUE #1230: [Android] AudioContext.resume() fails permanently after the Oboe stream dies with a non-Disconnected error (audio silent for the rest of the session)
## Summary

On Android, once the Oboe stream dies in a way that `AudioPlayer::onErrorAfterClose` does not recover from, `AudioContext.resume()` fails **permanently** for the rest of the process. The context reports `state === "suspended"`, every `resume()` call rejects with `"Failed to resume audio context."`, and nothing in the library ever rebuilds the stream. The app is silent until the user force-closes it.

We are seeing this in production (game built on `react-native-audio-api`, buffer-source SFX + music): ~20 Android users in a 14-day window with hundreds to thousands of consecutive resume failures each (worst case ~5,000 in a single stretch), across Pixel and Samsung devices.

## Environment

- `react-native-audio-api`: 0.13.2 (the wedge logic is unchanged on current `main`)
- React Native: 0.83.2 (Fabric)
- Android (production builds); iOS unaffected

## Root cause

1. The only stream-death recovery is `AudioPlayer::onErrorAfterClose` (android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp), and it is gated to `oboe::Result::ErrorDisconnected`:

```cpp
void AudioPlayer::onErrorAfterClose(oboe::AudioStream *stream, oboe::Result error) {
  if (error != oboe::Result::ErrorDisconnected || driverMutex_ == nullptr) {
    return;
  }
  ...
```

Any other terminal error (`ErrorNoService` after an audioserver restart, `ErrorTimeout`, etc.) returns early, leaving `mStream_` pointing at a stream Oboe has already closed. The same dead end is reached when the reopen inside the `ErrorDisconnected` path itself fails (there is no retry).

2. `BaseAudioContext::getState()` reports `SUSPENDED` whenever the driver is not running, so the JS side correctly sees `"suspended"` and calls `resume()`.

3. `AudioContext::resume()` (common/cpp/audioapi/core/AudioContext.cpp) then has no path that can succeed:

```cpp
if (isInitialized_.load(std::memory_order_acquire) && audioPlayer_->resume()) {
  // requestStart() on the closed stream fails -> false
  ...
}
return tryStartDriver();
```

`audioPlayer_->resume()` calls `requestStart()` on the closed stream and fails. Then `tryStartDriver()` immediately returns `false` because `isInitialized_` is still `true`:

```cpp
bool AudioContext::tryStartDriver() {
  ...
  if (isInitialized_.load(std::memory_order_acquire)) {
    return false;   // <- permanent: nothing ever resets isInitialized_ or reopens the stream
  }
```

So after any unrecovered stream death: `resume()` fails forever, and there is no API the app can use to recover short of tearing down the whole `AudioContext` and rebuilding its graph.

## Repro

Hard to trigger on demand, but killing the audio server while the app is in a suspended/idle-audio state reproduces the class of failure:

1. Play some audio through an `AudioContext`, let the context become suspended.
2. `adb shell killall audioserver` (or trigger any stream teardown that surfaces as a non-`ErrorDisconnected` terminal error).
3. Call `ctx.resume()` on the next play. It rejects, and keeps rejecting forever.

## Suggested fix

In `AudioContext::resume()`, when the driver is initialized but `audioPlayer_->resume()` fails, rebuild the stream instead of falling into the `tryStartDriver()` dead end. We are running this patch in production via patch-package:

```cpp
  if (isInitialized_.load(std::memory_order_acquire) && audioPlayer_->resume()) {
    setState(ContextState::RUNNING);
    return true;
  }

#ifdef ANDROID
  // An initialized driver whose stream can no longer start (audioserver
  // restart, route change missed by onErrorAfterClose's ErrorDisconnected-only
  // recovery) previously wedged forever: resume() failed and tryStartDriver()
  // early-returned false on isInitialized_. Rebuild the stream so the next
  // resume() heals the context instead.
  if (isInitialized_.load(std::memory_order_acquire)) {
    audioPlayer_->cleanup();
    if (!audioPlayer_->openAudioStream()) {
      return false;
    }
    isInitialized_.store(false, std::memory_order_release);
  }
#endif

  return tryStartDriver();
}
```

This makes `resume()` self-healing: if the rebuild fails (audio HW genuinely unavailable at that moment), the next `resume()` retries it. Widening `onErrorAfterClose` to recover from all terminal errors would also help, but the `resume()`-side rebuild covers the failed-reopen case too, and it puts recovery on a code path the app is already calling whenever audio is needed.

Happy to open a PR with the change if you'd take it.

basiav:
Closes [#1230](https://github.com/software-mansion/react-native-audio-api/issues/1230)

#### `AudioPlayer::onErrorAfterClose`
- More errors are handled in `AudioPlayer::onErrorAfterClose`: `ErrorDisconnected`, `ErrorTimeout`, `ErrorNoService`, `ErrorInternal`.
- For `ErrorDisconnected`, there is one try of the stream rebuild, if it fails `AudioContext::OnStreamFail` is called.
- For `ErrorTimeout`, `ErrorNoService` and `ErrorInternal` `onStreamFail` is called.

#### `AudioContext::onStreamFail`
- Calls the `AudioPlayer::cleanup` procedure, sets `isInitialized_` to false, invokes `onerror` defined in JS.

## Checklist

- [x] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
@@ -55,21 +55,38 @@ bool AudioPlayer::openAudioStream() {
   return true;
 }
 
+bool AudioPlayer::rebuildStream() {
+  cleanup();
+  return openAudioStream();
+}
+
 bool AudioPlayer::start() {
   std::scoped_lock lock(streamMutex_);
-  if (!isInitialized_.load(std::memory_order_acquire)) {
+
+  // Ensure a live stream exists. A prior Oboe death can leave isInitialized_ true
+  // with a closed/null stream, or a stream that rejects requestStart.
+  if (!isInitialized_.load(std::memory_order_acquire) || mStream_ == nullptr) {
+    if (isInitialized_.load(std::memory_order_acquire) || mStream_ != nullptr) {
+      cleanup();
+    }
     if (!openAudioStream()) {
       return false;
     }
   }
 
-  if (mStream_ != nullptr) {
-    auto result = mStream_->requestStart() == oboe::Result::OK;
-    isRunning_.store(result, std::memory_order_release);
-    return result;
+  if (mStream_ != nullptr && mStream_->requestStart() == oboe::Result::OK) {
+    isRunning_.store(true, std::memory_order_release);
+    return true;
   }
 
-  return false;
+  if (!rebuildStream() || mStream_ == nullptr) {
+    isRunning_.store(false, std::memory_order_release);
+    return false;
+  }
+
+  const bool started = mStream_->requestStart() == oboe::Result::OK;
+  isRunning_.store(started, std::memory_order_release);
+  return started;
 }
 
 void AudioPlayer::stop() {
@@ -88,20 +105,11 @@ bool AudioPlayer::resume() {
   }
 
   if (mStream_ != nullptr) {
-    if (mStream_->requestStart() == oboe::Result::OK) {
-      isRunning_.store(true, std::memory_order_release);
-      return true;
-    }
+    auto result = mStream_->requestStart() == oboe::Result::OK;
+    isRunning_.store(result, std::memory_order_release);
+    return result;
   }
 
-  if (rebuildStream()) {
-    if (mStream_ != nullptr && mStream_->requestStart() == oboe::Result::OK) {
-      isRunning_.store(true, std::memory_order_release);
-      return true;
-    }
-  }
-
-  isRunning_.store(false, std::memory_order_release);
   return false;
 }
 
@@ -167,9 +175,17 @@ AudioPlayer::onAudioReady(AudioStream *oboeStream, void *audioData, int32_t numF
 }
 
 void AudioPlayer::onErrorAfterClose(oboe::AudioStream *stream, oboe::Result error) {
-  // error != oboe::Result::ErrorDisconnected condition is deleted to handle more cases of errors
   if (driverMutex_ == nullptr) {
     return;
+  }
+
+  switch (error) {
+    case oboe::Result::ErrorDisconnected:
+    case oboe::Result::ErrorTimeout:
+    case oboe::Result::ErrorInternal:
+      break;
+    default:
+      return;
   }
 
   // Serialize with start()/resume()/suspend()/close() on the JS / promise-pool threads.
@@ -190,13 +206,14 @@ void AudioPlayer::onErrorAfterClose(oboe::AudioStream *stream, oboe::Result erro
   const bool wasRunning = isRunning_.load(std::memory_order_acquire);
 
   if (!rebuildStream()) {
+    isRunning_.store(false, std::memory_order_release);
     return;
   }
 
   // Restart the stream if it was expected to be running when the error occurred
-  if (wasRunning) {
-    auto result = mStream_->requestStart() == oboe::Result::OK;
-    isRunning_.store(result, std::memory_order_release);
+  if (wasRunning && mStream_ != nullptr) {
+    const bool started = mStream_->requestStart() == oboe::Result::OK;
+    isRunning_.store(started, std::memory_order_release);
   }
 }
 
@@ -242,10 +259,5 @@ double AudioPlayer::getOutputLatency() const {
   }
 
   return minBaseLatency;
-}
-
-bool AudioPlayer::rebuildStream() {
-  cleanup();
-  return openAudioStream();
 }
 } // namespace audioapi
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
@@ -88,11 +88,20 @@ bool AudioPlayer::resume() {
   }
 
   if (mStream_ != nullptr) {
-    auto result = mStream_->requestStart() == oboe::Result::OK;
-    isRunning_.store(result, std::memory_order_release);
-    return result;
+    if (mStream_->requestStart() == oboe::Result::OK) {
+      isRunning_.store(true, std::memory_order_release);
+      return true;
+    }
   }
 
+  if (rebuildStream()) {
+    if (mStream_ != nullptr && mStream_->requestStart() == oboe::Result::OK) {
+      isRunning_.store(true, std::memory_order_release);
+      return true;
+    }
+  }
+
+  isRunning_.store(false, std::memory_order_release);
   return false;
 }
 
@@ -158,7 +167,8 @@ AudioPlayer::onAudioReady(AudioStream *oboeStream, void *audioData, int32_t numF
 }
 
 void AudioPlayer::onErrorAfterClose(oboe::AudioStream *stream, oboe::Result error) {
-  if (error != oboe::Result::ErrorDisconnected || driverMutex_ == nullptr) {
+  // error != oboe::Result::ErrorDisconnected condition is deleted to handle more cases of errors
+  if (driverMutex_ == nullptr) {
     return;
   }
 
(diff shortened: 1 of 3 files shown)


packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
```diff
@@ -176,9 +186,17 @@ void AudioPlayer::onErrorAfterClose(oboe::AudioStream *stream, oboe::Result erro
...
-  cleanup();
-  if (openAudioStream()) {
-    resume();
+  // Check if the stream was expected to be running when the error occurred
+  const bool wasRunning = isRunning_.load(std::memory_order_acquire);
+
+  if (!rebuildStream()) {
+    return;
+  }
+
```
closetcaiman, line 200: If `rebuildStream` fails, `isRunning` flag may stay `true`.

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
```diff
@@ -158,7 +167,8 @@ AudioPlayer::onAudioReady(AudioStream *oboeStream, void *audioData, int32_t numF
 }
 
 void AudioPlayer::onErrorAfterClose(oboe::AudioStream *stream, oboe::Result error) {
-  if (error != oboe::Result::ErrorDisconnected || driverMutex_ == nullptr) {
+  // error != oboe::Result::ErrorDisconnected condition is deleted to handle more cases of errors
+  if (driverMutex_ == nullptr) {
```
closetcaiman, line 166: Removing this guard seems really dangerous to me. Consider this: should all types of errors result in stream rebuild retry? What about some non-recoverable errors, won't that create an infinite loop? Maybe it is better to choose action in response to specific errors -  `oboe::Result` is not that big of an `enum`.

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
```diff
@@ -88,11 +88,20 @@ bool AudioPlayer::resume() {
...
 
+  if (rebuildStream()) {
+    if (mStream_ != nullptr && mStream_->requestStart() == oboe::Result::OK) {
+      isRunning_.store(true, std::memory_order_release);
+      return true;
+    }
+  }
+
+  isRunning_.store(false, std::memory_order_release);
   return false;
```
closetcaiman, line 101: I'm not a fan of `resume` being responsible for rebuilding the stream. Seems like it tries to act both as `resume` and `start`.

closetcaiman (changes requested):

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.h
```diff
@@ -69,6 +69,9 @@ class AudioPlayer : public CommonPlayer,
   std::weak_ptr<AudioContext> context_;
 
   bool openAudioStream();
+  bool rebuildStream();
+
+  std::atomic<bool> isFailed_{false};
```
mdydek: i dont see it being used

packages/react-native-audio-api/common/cpp/audioapi/core/AudioContext.cpp
```diff
@@ -8,9 +8,15 @@
 
 #include <audioapi/core/AudioContext.h>
 #include <audioapi/core/destinations/AudioDestinationNode.h>
+
+#ifdef ANDROID
+#include <android/log.h>
+#endif
+
```
mdydek:
```suggestion
```

packages/react-native-audio-api/src/core/AudioContext.ts
```diff
@@ -28,6 +30,20 @@ export default class AudioContext extends BaseAudioContext {
     return (this.context as IAudioContext).outputLatency;
   }
 
+  /**
+   * Web Audio `onerror` — fired when the output device/stream fails natively.
+   *
+   * @see https://webaudio.github.io/web-audio-api/#dom-audiocontext-onerror
+   */
```
mdydek:
```suggestion
```

packages/react-native-audio-api/src/jsi-interfaces.ts
```diff
@@ -91,6 +91,8 @@ export interface IAudioContext extends IBaseAudioContext {
   close(): Promise<undefined>;
   resume(): Promise<undefined>;
   suspend(): Promise<undefined>;
+  /** Web Audio `onerror` — JS function stored on the native HostObject. */
```
mdydek:
```suggestion
```

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioContextHostObject.cpp
```diff
@@ -78,4 +79,19 @@ JSI_HOST_FUNCTION_IMPL(AudioContextHostObject, createMediaElementSource) {
...
+    return;
+  }
+
+  auto jsFunc = std::make_shared<jsi::Function>(value.getObject(runtime).getFunction(runtime));
+
+  audioContext->setOnError([jsFunc, invoker = callInvoker_, rt = &runtime]() {
+    invoker->invokeAsync([jsFunc, rt]() { jsFunc->call(*rt); });
+  });
+}
+
```
mdydek, line 90: even if it looked simple, it can introduce some errors, such as datarace when swaping the function mid oboe's onError, reference cycle due to strong ptr and that it can be destroyed on different thread. Utilize existing AudioEventHandlerRegistry and compare how events are done using it and implement the `onerror` this way.

mdydek (changes requested):
add also mention in the docs about `onerror`

packages/audiodocs/docs/core/audio-context.mdx
```diff
@@ -75,3 +75,22 @@ Resumes a previously suspended audio context.
 
 Inherits [`onstatechange`](./base-audio-context.mdx#onstatechange) from [`BaseAudioContext`](./base-audio-context.mdx#events); `close`, `suspend` and `resume` each fire it once their returned promise has resolved.
 
+### `onerror`
+
+Sets (or removes, when `null` is assigned) a callback fired when the live output stream fails natively and cannot keep rendering — for example after the audio device disconnects or the OS audio service dies while the context is running.
+
+This matches the Web Audio [`AudioContext.onerror`](https://webaudio.github.io/web-audio-api/#dom-audiocontext-onerror) handler. The callback receives no arguments.
+
+On **Android**, the event is dispatched from the native driver after an Oboe stream error when recovery fails (the context’s driver is cleared so a later [`resume`](./audio-context.mdx#resume) can reopen it). You can assign `onerror` on every platform, but **iOS** does not yet call it from the audio engine, and **web** does not yet forward the browser’s `AudioContext.onerror`.
```
mdydek:
```suggestion
```

this internal info is not necessary

mdydek (approved):

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AudioPlayer.cpp
basiav: Fair point. I've modified this to include 4 common errors: `ErrorDisconnected`, `ErrorTimeout`, `ErrorInternal`, `ErrorNoService`, 2 of them mentioned in the issue. We are not sure what `ErrorTimeout`, `ErrorInternal` and `ErrorNoService` rebuild will turn out in, it depends on the situation in question (especially hardware aspects) and it is not completely clear how to proceed. An infinite loop is possible, however:
- I've let only `ErrorDisconnected` try to rebuild the stream once, if it fails then `AudioContext::onStreamFail` is called.
- I've let `ErrorTimeout`, `ErrorInternal` and `ErrorNoService` only to call `AudioContext::onStreamFail`, so that there is the aspect of invoking `onerror` (absent in a regular stream rebuild), so we are letting some possible logic in (mainly of waiting, e.g. 2 seconds before `AudioContext::resume`). This can help reduce the risk of an infinite loop, which is however possible.

packages/react-native-audio-api/common/cpp/audioapi/HostObjects/AudioContextHostObject.cpp
basiav: corrected

Issue #734, opened 2025-10-10
Select input device for recording?


murphycj:
### Feature Request

I know there is an api to see available devices (`AudioManager.getDevicesInfo`), but is there a way to select the input device to use? If not, is such a feature planned?

PR #1255, opened 2026-08-26
feat(android): implement setInputDevice, which was a no-op

kimchouard:
## The bug

`AudioManager.setInputDevice` works on iOS and does nothing on Android — the module method is a `// TODO: noop for now` that resolves unconditionally. Capture stays on the platform default while the API reports success.

Known gap (the docs say so, and #734 was closed by #926 which shipped iOS only). I hit it building a feature that records a USB-connected instrument on a phone.

## The fix

`AudioInputSelection` holds the preferred device id for the process; `AndroidAudioRecorder::openAudioStream()` reads it and passes it to `oboe::AudioStreamBuilder::setDeviceId()`.

Process-wide rather than per-recorder because Android routes capture per process, and the selection arrives through the `AudioAPIModule` TurboModule, which holds no reference to any recorder. Nothing in `common/cpp` depends on it. I chose to make the existing cross-platform API work rather than add a `RecorderOptions.deviceId` that would diverge from iOS.

Two parts look like more than the job needs, so briefly:

**A stale stream is replaced, not reused.** `stop()` never closes the stream and `openAudioStream()` returned early whenever one existed — so a recorder reuses its first stream for every later `start()`. Setting the device only in the builder would work on the first recording and never again.

**A device that wasn't honoured fails the open.** Oboe applies `setDeviceId` on AAudio only and reports `kUnspecified` under OpenSL ES, so the opened stream's `getDeviceId()` is checked against the request and a mismatch closes the stream and errors. Recording from a device the caller didn't pick seemed worse than not recording. (Deliberately not forcing `setAudioApi(AAudio)` — Oboe's header calls that "extremely risky" on 8.0.)

**Nothing changes for callers who don't use it:** with no selection the value is `oboe::kUnspecified`, `setDeviceId` is never called, the readback is skipped, and the early return behaves exactly as before.

Also: `TYPE_USB_DEVICE` / `TYPE_USB_HEADSET` / `TYPE_USB_ACCESSORY` added to `parseDeviceCategory` (a USB interface enumerated as `"Other (11)"`), and `currentInputs` populated (it was always empty, so `useAudioInput` could never report the current device).

## Questions for you

1. **What should this do while a recorder is running?** iOS reroutes live. Android binds the device as the stream opens, so it can't. This rejects, telling the caller to stop → select → start, on the reasoning that resolving while silently deferring is the worse failure. A live restart means reopening the file, re-`prepare()`ing the callback at a possibly different sample rate and re-`init()`ing the adapter node while the audio thread is in `onAudioReady` — which looked like the duplex-stream work the TODO defers. **Say the word and I'll implement the reroute instead.**
2. **Clearing the preference** isn't possible on either platform today (iOS rejects a nil id before reaching `setPreferredInput:nil`; Android now mirrors that). I had a cross-platform fix and pulled it out — widening `deviceId` to `string | null` in the spec is your API call, not a bug fix. Happy to open it separately.
3. `android/src/oldarch/NativeAudioAPIModuleSpec.java` is hand-maintained and nothing in the repo compiles it. Untouched here, but it can drift from codegen silently.

## Verification

All green: `validate:android` (409 C++ tests, NDK build), `lint:cpp`, `lint:kotlin`, `lint:js`, `format:check:js`, `format:check:android:kotlin`, `typecheck`, `test:js` (79), `check-audio-enum-sync`.

**Not tested on a device.** The readback assert, the reconnect path and reject-while-paused are compiled and reasoned about, never executed on hardware. Happy to check anything specific if you can point me at it.

**`format:check` for C++ is red, but not from this PR** — android 24 / ios 50 / common 1325, identical counts at base commit `c12bd7c9`. My files appear in none of the lists. Measured with clang-format 23.1.0; older releases may report differently.

Left out to keep this focused: `TYPE_BLE_HEADSET` is API 31 and `parseDeviceCategory` is `@RequiresApi(O)`, so it needs an annotation bump.

### DIFF
--- a/packages/audiodocs/docs/system/audio-manager.mdx
+++ b/packages/audiodocs/docs/system/audio-manager.mdx
@@ -167,7 +167,24 @@ Checks if notification permissions were previously granted.
 
 Checks currently used and available devices.
 
+On Android, `currentInputs` reports the device selected through [`setInputDevice`](#setinputdevice) and is empty until one is selected, since the platform does not report which input it would pick on its own. `currentOutputs` is always empty on Android.
+
 #### Returns `Promise<AudioDevicesInfo>`, which is resolved after receiving the answer from the system.
+
+### `setInputDevice`
+
+Selects the device that audio is captured from, using an `id` taken from `getDevicesInfo().availableInputs`.
+
+| Name | Type | Description |
+| :----: | :----: | :---- |
+| `deviceId` | `string` | Identifier of the input device to capture from. |
+
+The two platforms differ in when the selection takes effect:
+
+- **iOS** sets the session's preferred input, which reroutes a running session right away.
+- **Android** binds the device while an Oboe capture stream opens, so the selection applies to recorders started afterwards. Calling it while a recorder is running or paused rejects instead of deferring the switch silently: stop the recorder, select the device, then start it again. A paused recorder still holds its stream and resumes onto the same device, which is why it is refused too. Android also needs the AAudio backend, which is what Oboe picks on Android 8.1 and newer; a recorder that can only open through OpenSL ES fails to start with an explanatory message rather than recording from the wrong device. If the selected device disappears mid-recording, the recorder reports an error through `onError` instead of falling back to the built-in microphone.
+
+#### Returns `Promise<void>`, which is resolved once the device is selected and rejected when the device cannot be found, when the system fails to switch to it, or, on Android, when a recorder is running or paused.
 
 ## Remarks
 
@@ -329,7 +346,7 @@ export type AudioDeviceList = AudioDeviceInfo[];
 export interface AudioDevicesInfo {
   availableInputs: AudioDeviceList;
   availableOutputs: AudioDeviceList;
-  currentInputs: AudioDeviceList; // iOS
+  currentInputs: AudioDeviceList; // iOS, and Android once a device is selected
   currentOutputs: AudioDeviceList; // iOS
 }
 ```
--- a/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
+++ b/packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
@@ -1,5 +1,6 @@
 #include <android/log.h>
 #include <audioapi/android/core/AndroidAudioRecorder.h>
+#include <audioapi/android/core/AudioInputSelection.h>
 #include <audioapi/android/core/utils/AndroidFileWriterBackend.h>
 #include <audioapi/android/core/utils/AndroidRecorderCallback.h>
 
@@ -51,6 +52,30 @@ std::optional<oboe::InputPreset> inputPresetFromString(const std::string &name) 
   }
   return std::nullopt;
 }
+
+/// Runs an action on scope exit unless it is dismissed first, so that a
+/// multi-step operation can undo a claim it took up front without repeating
+/// the undo on every failure path.
+template <typename Action>
+class ScopeExit {
+ public:
+  explicit ScopeExit(Action action) : action_(std::move(action)) {}
+  ~ScopeExit() {
+    if (armed_) {
+      action_();
+    }
+  }
+
+  DELETE_COPY_AND_MOVE(ScopeExit);
+
+  void dismiss() {
+    armed_ = false;
+  }
+
+ private:
+  Action action_;
+  bool armed_ = true;
+};
 } // namespace
 
 AndroidAudioRecorder::AndroidAudioRecorder(
@@ -60,7 +85,8 @@ AndroidAudioRecorder::AndroidAudioRecorder(
       inputPreset_(std::move(options.androidInputPreset)),
       streamSampleRate_(0.0),
       streamChannelCount_(0),
-      streamMaxBufferSizeInFrames_(0) {}
+      streamMaxBufferSizeInFrames_(0),
+      streamDeviceId_(AudioInputSelection::kSystemDefaultDeviceId) {}
 
 /// @brief Destructor ensures that the audio stream and each output type are closed and flushed up remaining data.
 /// callable from the JS thread or handled by audio thread (if js dropped recorder first).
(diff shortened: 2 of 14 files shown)


packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/AudioAPIModule.kt
```diff
@@ -172,12 +182,38 @@ class AudioAPIModule(
...
 
+  /**
+   * Selects the capture device every recorder opens its input stream on.
+   *
+   * Unlike iOS, which reroutes a live session, the selection is bound while an
+   * Oboe input stream opens. Changing it therefore only affects streams opened
+   * afterwards, and is rejected outright while a recorder holds a stream so that
+   * a caller never mistakes a deferred switch for an applied one. A paused
+   * recorder still holds its stream and resumes onto the same device, so it
+   * counts as holding one.
```
mdydek:
```suggestion
```
we try to keep the ai comments minimalistic

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/MediaSessionManager.kt
```diff
@@ -210,39 +211,72 @@ object MediaSessionManager {
...
+      .firstOrNull { it.id.toString() == deviceId }
+
+  fun setPreferredInputDevice(device: AudioDeviceInfo) {
+    this.preferredInputDeviceId = device.id
+  }
+
   @RequiresApi(Build.VERSION_CODES.O)
   fun getDevicesInfo(): ReadableMap {
     val availableInputs = Arguments.createArray()
+    val currentInputs = Arguments.createArray()
```
mdydek: can we have more than one input?

packages/react-native-audio-api/android/src/main/java/com/swmansion/audioapi/system/MediaSessionManager.kt
```diff
@@ -210,39 +211,72 @@ object MediaSessionManager {
...
+   * native capture layer, and `AudioAPIModule.setInputDevice` is the only writer
+   * of either.
+   *
+   * Null means no explicit selection was made and the platform picks the device.
+   * Android offers no way to learn which one that is before a stream opens, so
+   * `currentInputs` stays empty in that case.
+   *
+   * Written from the React Native module thread and read by whichever thread
+   * calls `getDevicesInfo`, hence volatile.
+   */
```
mdydek:
```suggestion
  /**
   * Null means no explicit selection was made and the platform picks the device.
   *
   * Written from the React Native module thread and read by whichever thread
   * calls `getDevicesInfo`, hence volatile.
   */
```

packages/audiodocs/docs/system/audio-manager.mdx
```diff
@@ -167,8 +167,25 @@ Checks if notification permissions were previously granted.
...
+Selects the device that audio is captured from, using an `id` taken from `getDevicesInfo().availableInputs`.
+
+| Name | Type | Description |
+| :----: | :----: | :---- |
+| `deviceId` | `string` | Identifier of the input device to capture from. |
+
+The two platforms differ in when the selection takes effect:
+
+- **iOS** sets the session's preferred input, which reroutes a running session right away.
+- **Android** binds the device while an Oboe capture stream opens, so the selection applies to recorders started afterwards. Calling it while a recorder is running or paused rejects instead of deferring the switch silently: stop the recorder, select the device, then start it again. A paused recorder still holds its stream and resumes onto the same device, which is why it is refused too. Android also needs the AAudio backend, which is what Oboe picks on Android 8.1 and newer; a recorder that can only open through OpenSL ES fails to start with an explanatory message rather than recording from the wrong device. If the selected device disappears mid-recording, the recorder reports an error through `onError` instead of falling back to the built-in microphone.
```
mdydek:
```suggestion
- **Android** binds the device while an Oboe capture stream opens, so the selection applies to recorders started afterwards. Calling it while a recorder is running or paused rejects instead of deferring the switch silently: stop the recorder, select the device, then start it again. Android also needs the AAudio backend, which is what Oboe picks on Android 8.1 and newer; a recorder that can only open through OpenSL ES fails to start with an explanatory message rather than recording from the wrong device. If the selected device disappears mid-recording, the recorder reports an error through `onError` instead of falling back to the built-in microphone.
```

packages/audiodocs/docs/system/audio-manager.mdx
```diff
@@ -167,8 +167,25 @@ Checks if notification permissions were previously granted.
...
+
+#### Returns `Promise<void>`, which is resolved once the device is selected and rejected when the device cannot be found, when the system fails to switch to it, or, on Android, when a recorder is running or paused.
```
mdydek:
```suggestion
#### Returns `Promise<void>`
resolved once the device is selected and rejected when the device cannot be found, when the system fails to switch to it, or, on Android, when a recorder is running or paused.
```

packages/react-native-audio-api/android/src/main/cpp/audioapi/android/core/AndroidAudioRecorder.cpp
```diff
@@ -109,13 +148,38 @@ Result<NoneType, std::string> AndroidAudioRecorder::openAudioStream() {
...
 
   if (result != oboe::Result::OK || mStream_ == nullptr) {
     return Result<NoneType, std::string>::Err(
         "Failed to open audio stream: " + std::string(oboe::convertToText(result)));
   }
 
+  // Oboe honours setDeviceId on the AAudio backend only; OpenSL ES drops the
+  // request and reports kUnspecified instead (see AudioStreamBuilder::setDeviceId).
+  // Recording from a device the caller did not ask for is worse than not
+  // recording at all, so the stream is dropped and the open reported as failed.
```
mdydek:
```suggestion
  // Oboe honours setDeviceId on the AAudio backend only; OpenSL ES drops the
  // request and reports kUnspecified instead
  // Recording from a device the caller did not ask for is worse than not
  // recording at all, so the stream is dropped and the open reported as failed.
```

mdydek (approved):

maciejmakowski2003 (changes requested):
thanks for you work @kimchouard !!! I was discussing your question about recorder behaviour when device input change mid-recording and we concluded that we should go towards some kind of self-healing solution meaning automatic rerouting. let me know if you have time for that enhancement. if not, we will take it over.

Issue #1254, opened 2026-08-26
Android: platform AEC doesn't cancel app's own playback unless the OUTPUT stream uses Usage::VoiceCommunication


arekrgw:
### Feature Request

When building a full-duplex voice assistant (mic capture + TTS playback at the same
time), enabling `androidInputPreset: 'voiceCommunication'` on the `AudioRecorder` alone
was **not enough** to get working acoustic echo cancellation on Android. Our STT kept
transcribing the assistant's own TTS playback.

The missing piece: the **output** stream (`AudioPlayer.cpp`) is opened without any
`Usage`, which defaults to `USAGE_MEDIA`. On many devices, the platform AEC chain
engaged by the `VOICE_COMMUNICATION` input preset only references the
voice-communication output path and media-stream playback leaks into the microphone
uncancelled.

Adding two lines to the output stream builder fixed it completely:

```cpp
// AudioPlayer.cpp — openAudioStream()
builder.setSharingMode(SharingMode::Exclusive)
    ->setFormat(AudioFormat::Float)
    // ...existing config...
    ->setUsage(Usage::VoiceCommunication)   // ← added
    ->setContentType(ContentType::Speech)  // ← added
    // ...
```

### Environment

- `react-native-audio-api` 1.0.0-nightly-f8b57cd-20260825
- React Native 0.86.2 / Expo SDK 57, physical Android device (loudspeaker, no headphones)
- Capture: `AudioRecorder` with `{ androidInputPreset: 'voiceCommunication', iosVoiceProcessing: true }`
- Playback: `AudioContext` playing TTS PCM buffers
- Use case: streaming STT (websocket ASR) while speaking TTS replies (barge-in style interaction)

### Symptom before the patch

- Speech-to-text transcribed the assistant's own spoken replies as if they were user input
- Token-overlap heuristics and LLM-based intent filtering reduced but never eliminated
  the self-echo turns
- Verified the input preset reaches Oboe correctly (`setInputPreset(VoiceCommunication)`
  present in `openAudioStream()`), so the gap was on the output/reference side

### After the patch

- Self-echo eliminated at DSP level (zero phantom transcripts during playback)
- Real barge-in (user talking over the reply) works reliably
- No audio-routing regressions on our test device: playback still comes out of the
  loudspeaker as before

### Suggestions

1. **Library behavior**: consider applying `Usage::VoiceCommunication` (+ `ContentType::Speech`)
   to the output stream automatically when an active recorder uses
[...]

PR #1256, opened 2026-08-27
docs(audio-recorder): note that Android AEC also needs the playback path

recepteksi:
Refs #1254

## ⚠️ Breaking changes ⚠️

- None. Documentation only.

## Introduced changes

- Adds a caution to the AudioRecorder full-duplex section noting that on Android
  the input preset is only the capture half of echo cancellation.

The section currently presents

```tsx
new AudioRecorder({ androidInputPreset: 'voiceCommunication', iosVoiceProcessing: true })
```

as what a full-duplex app needs. That is complete on iOS. On Android the platform
AEC cancels against the **voice-communication output path**, and `AudioContext`
playback defaults to `USAGE_MEDIA`, so on some devices the app's own output still
reaches the microphone with the preset set correctly — which is what #1254
reports, with a proposed fix on the output builder.

This PR does not touch that behaviour: @closetcaiman said in the issue that the
approach needs thinking through, and this is only the documentation half
(suggestion 2 in the report), so it should not constrain that decision.

It also adds a testing note. Headphones remove the acoustic path entirely, so a
broken setup verified over headphones looks fixed — worth stating next to the
recommendation itself.

### How I ran into it

Building a voice assistant on `0.13.3`, which predates `androidInputPreset`, so
our Android capture never engaged the platform AEC at all and the session had to
stop listening while it spoke. Reading upstream to plan the upgrade, the docs
suggested the input preset alone would be enough; #1254 says otherwise. I have
not reproduced the output-path half on a device myself — the wording follows the
issue rather than asserting more than I verified, and is deliberately hedged
("on some devices").

## Checklist

- [x] Linked relevant issue
- [ ] Added/Conducted relevant tests — documentation only
- [ ] Updated Web Audio API coverage — not applicable
- [ ] Added support for web — not applicable
- [ ] Updated old arch android spec file — not applicable

### DIFF
--- a/packages/audiodocs/docs/inputs/audio-recorder.mdx
+++ b/packages/audiodocs/docs/inputs/audio-recorder.mdx
@@ -371,6 +371,23 @@ const audioRecorder = new AudioRecorder({
 });
 ```
 
+:::caution Android: the playback side matters too
+
+On Android the platform AEC engaged by `voiceCommunication` cancels against the
+**voice-communication output path**. Playback started from an `AudioContext`
+does not sit on that path — it defaults to `USAGE_MEDIA` — so on some devices
+the app's own output can still leak into the capture even with the input preset
+set correctly.
+
+If you are building barge-in style interaction and still hear the app
+transcribe itself, this is the likely cause. Progress is tracked in
+[#1254](https://github.com/software-mansion/react-native-audio-api/issues/1254).
+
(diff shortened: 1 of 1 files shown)


mdydek:
we are currently improving and adding features that are connected to the AEC and it will probably be refactored before next release, so your change wouldn't make it to the release anyway, but thank you for the contribution anyway

PR #1257, opened 2026-08-28
fix(ios): recording resume after an interruption

### LINKED ISSUE #1220: iOS: recording never resumes after an interruption (onInterruptionEnd starts the engine without activating the audio session)
### Summary

`AudioEngine.onInterruptionEnd` starts the audio engine without ever activating the audio session, on a session the interruption-began branch explicitly marked inactive. For a recorder that means capture never comes back after a phone call. We patched it locally and **verified on device that recording now resumes mid-call-end with the app in the background**, so this is a confirmed fix rather than a proposal.

> **Note on revisions.** This report has been trimmed. Earlier versions led with `shouldResume` and with a claim that consumers cannot recover from JS. We later measured `shouldResume=1` on every call and found our own symptom had a consumer-side cause (see the last section), so both have been removed rather than left to mislead. The defect below is the one that survived.

### Environment

- `react-native-audio-api` 0.13.2
- React Native 0.85.3, Expo SDK 56, Hermes, iOS device (not reproducible on the simulator)
- session `playAndRecord` / `spokenAudio`, `UIBackgroundModes: audio`, interruption observation OFF

### 1. `onInterruptionEnd` never activates the session

Compare it with `startEngine`, the path taken when the user taps record:

| | `startEngine` (AudioEngine.mm:441) | `onInterruptionEnd` (AudioEngine.mm:366) |
|---|---|---|
| activates the session | `[sessionManager ensureActive:true]` (:452) | **never** |
| checks the input node materialized | yes (:464) | no |
| on failure | returns `false`, caller may retry | sets state `Idle` and returns (:390) |

The session it starts on was explicitly marked inactive when the interruption began:

```objc
// SystemNotificationManager.mm:121
if (interruptionType == AVAudioSessionInterruptionTypeBegan) {
    dispatch_async(dispatch_get_main_queue(), ^{
      [audioEngine onInterruptionBegin];
      [sessionManager markInactive];      // isActive = false
    });
```

Nothing reactivates it in between, so `startAndReturnError` at :384 runs against an inactive session and cannot succeed with an input node attached.

Three things compound it:

1. **The failure poisons the state.** `self.state = Idle` at :390, while the guard at :371 requires `Interrupted`. Once it has failed it can never run again for that recording, so a re-delivered ended notification and `activelyReclaimSession`'s `onInterruptionEnd:true` (:303) are both no-ops.
2. **Ordering matters more than just "activate before start".** `stopEngine` (:375) already sets `Idle` (:510), so an activation inserted lower down still returns `Idle` on refusal. And `rebuildAudioEngine` (:376) materializes the input node by reading `currentInputConnectionFormat` (:189), which is unavailable while the session is inactive. The activation has to come before **any** engine call, which is what `startEngine` does.
3. **No missing-input-node guard.** Unlike `startEngine` (:482), a failed materialization here is indistinguishable from success, so the engine can start with no sink attached: `isRecording()` reports true and not one buffer is ever delivered.

The shape we are running locally, and would be glad to send as a PR:

```objc
- (void)onInterruptionEnd:(bool)shouldResume
{
  std::scoped_lock lock(_engineLock);
  NSError *error = nil;

  if (self.state != AudioEngineState::AudioEngineStateInterrupted) {
    return;
  }

  if (!shouldResume && self.inputRegistration == nil) {
    [self stopEngine];
    [self rebuildAudioEngine];
    self.state = AudioEngineState::AudioEngineStatePaused;
    return;
  }

  // BEFORE any engine call: the format read inside rebuildAudioEngine needs an active session, and a
  // refusal here must return with Interrupted still intact so a later trigger can retry.
  if (![self.sessionManager ensureActive:true error:&error]) {
    NSLog(@"Error while activating audio session after interruption: %@", [error debugDescription]);
    return;
  }

  [self stopEngine];
  [self rebuildAudioEngine];

  if (self.inputRegistration != nil && self.inputNode == nil) {
    NSLog(@"Error while materializing the audio input node after interruption: missing live input format");
    self.state = AudioEngineState::AudioEngineStateInterrupted;
    return;
  }

  [self.audioEngine prepare];
  [self.audioEngine startAndReturnError:&error];

  if (error != nil) {
    NSLog(@"Error while restarting the audio engine after interruption: %@", [error debugDescription]);
    self.state = AudioEngineState::AudioEngineStateInterrupted;   // not Idle: stopEngine already set that
    return;
  }

  self.state = AudioEngineState::AudioEngineStateRunning;
  self.sessionDeactivationInvalidatedGraph = false;
}
```

The `inputRegistration == nil` condition on the `shouldResume` bail is a separate opinion, not part of the bug: a recorder the user explicitly started should always come back, while playback keeps the polite behaviour. Drop it if you disagree; the rest stands on its own.

Device log with the patch applied, app backgrounded, real incoming call (`Interrupted=3`, `Running=1`):

```
[audioapi] onInterruptionEnd shouldResume=1 state=3 hasInput=1   <- guard passed, no activation error follows
[audioapi] onInterruptionEnd shouldResume=1 state=1 hasInput=1   <- duplicate notification correctly bails
```

Capture resumed into the same file, with only the call's own audio missing, before the app was reopened.

### 2. Docs and behaviour disagree on `observeAudioInterruptions`

The [AudioManager docs](https://docs.swmansion.com/react-native-audio-api/docs/system/audio-manager/) say:

> On iOS, it enables/disables event emission only.

That reads as a pure observability toggle. But the interruption-ended branch is an either/or:

```objc
// SystemNotificationManager.mm ~line 138
if (self.audioInterruptionsObserved) {
  // emit the JS "ended" event, and nothing else
} else {
  dispatch_async(dispatch_get_main_queue(), ^{ [audioEngine onInterruptionEnd:shouldResume]; });
}
```

So subscribing to interruption events hands recovery to JavaScript, which for a recorder is a trap: iOS can suspend the runtime while the session is inactive, so the app takes ownership of a recovery it may not be scheduled to perform. The same either/or appears at lines 171 and 177.

Either run the native recovery regardless of whether events are observed (they are not mutually exclusive concerns), or document that observing transfers recovery responsibility to the consumer.

### 3. Two smaller things

- **A no-op deactivation still has side effects.** `setActive:false` short-circuits on `!self.isActive` (AudioSessionManager.mm:155) and never touches AVAudioSession, but `AudioAPIModule.mm:180` runs `handleSessionDeactivation` regardless of whether anything was deactivated, and that forces the engine state to `Paused` (:352, :357). A consumer calling `setAudioSessionActivity(false)` to reclaim a session therefore moves its engine out of `Interrupted`, which is the state `onInterruptionEnd` requires.
- **`resume()` gives JS nothing to act on.** `startEngine` returns `bool`, but `startIfNecessary`'s result is dropped and `IOSAudioRecorder::resume` returns `void`, so a failed engine start reaches JS as nothing at all, only an `NSLog`. A consumer cannot distinguish "resumed" from "silently did nothing", which makes any retry policy guesswork.

Footnote on a related hazard we avoided rather than hit: `IOSAudioRecorder::resume` (:446) stores `Recording` unconditionally even when `[nativeRecorder_ resume]` failed to start the engine. If `inputArmed` was cleared by a preceding `pause()` and the engine is later restarted by the native path (which sets `Running` without re-arming), the recorder reports `isRecording() == true` while the receiver block drops every buffer at `NativeAudioRecorder.m:40`, with no error anywhere and no way to re-arm. Storing `Recording` only on success, or treating `inputArmed == NO` as not-recording, would close it.

### For other consumers hitting "recording never resumes": check your session options first

Our symptom had a second, consumer-side cause, and it is worth stating plainly so nobody else spends as long on it as we did. With a non-mixable `playAndRecord` session, iOS refuses activation to a **backgrounded** app:

```
[CannotInterruptOthers] code: 560557684    // refused, app in background
```

Adding `mixWithOthers` to `iosOptions` changed the error to `[InsufficientPriority] 561017449`, which appears only while the call is genuinely still in progress, and after that both the native path and ours activated successfully from the background. A mixable session does not need to interrupt anyone, so it can be activated when a non-mixable one cannot.

This is not a library bug, and the library exposes the option already. It does interact with the defect above, though: the missing session activation in `onInterruptionEnd` is what made the whole thing undiagnosable, because the native path failed for a reason no log ever reported.

### Related

- #1209 looks like the same class of problem from a different trigger.
- #1013 is a crash in the reattach path this depends on. Could you confirm the fix is in 0.13.2?

SomePersonFromMars:
Closes #1220

## ⚠️ Breaking changes ⚠️

- **iOS - `observeAudioInterruptions(true)` no longer replaces native recovery.** Native `onInterruptionBegin` / `onInterruptionEnd` always run. The flag only enables JS `interruption` events, which matches the existing `AudioManager` docs. Apps that subscribed because observing used to skip native resume should not own engine recovery for a recorder (in particular, do not call `Recorder.pause()` on `began` - that moves the engine to `Paused` and disables the native retry). Playback can still pause on `began`; native does not resume players.

## Introduced changes

Fixes iOS recording that never came back after an interruption. **Updated library policy:** the same take resumes when the app is, or becomes, **foreground**. Background restart after the interrupter leaves is best-effort; iOS can still refuse I/O while the app is backgrounded.

### Responds to the issue

**§1 `onInterruptionEnd` never activated the session; failure poisoned `Idle`; no input-node guard**

- Activate the session *before* any engine teardown/rebuild (matches `startEngine`).
- On activation or `startAndReturnError` failure, stay `Interrupted` and return (do not set `Idle`). A later retry can run.
- If a recorder is attached but the input sink did not materialize, stay `Interrupted` and treat it as capture lost (do not look like a successful start with no buffers).
- Always try to resume a recorder even when `shouldResume` is false; playback-only still pauses when the OS says not to resume.

**§2 Docs vs behavior for `observeAudioInterruptions`**

- When JS is configured to `observeAudioInterruptions`, an appropriate handler runs only when `AudioEngine` actually transitioned to a new state. For example, if the engine was `Interrupted`, the handler runs on a *successful* transition to `Running`.
- Native recovery now runs independently of whether JS is observing interruptions, in accordance with the documentation.

**§3 No-op `setActive:false` still ran deactivation**

- `setAudioSessionActivity(false)` only calls session-deactivation handling if the session was actually managed and active *before* `setActive`. It no longer transitions `Interrupted` to `Paused`, which used to disable `onInterruptionEnd`.

**§3 / footnote - `resume()` / store `Recording` on a dead engine**

- `resume()` still does not have a return type, but it is now guaranteed that `isRecording()` is true if and only if the audio engine is not in an interrupted state and the recorder is actually running.
- According to my experiments and research, it is now most likely never necessary to implement a custom retry policy in JS.
    - `AudioEngine`'s `onInterruptionEnd` recovery after `AVAudioSessionInterruptionTypeEnded` can still fail. That can happen, for example, due to a race with a higher-priority app or internal iOS privacy policy. In that case we retry once the app returns to the foreground, which is then guaranteed by iOS to succeed.
    - At the same time, I can imagine a situation where an active retry policy might be useful, but I am unable to trigger a scenario in which the first background interrupt recovery fails and the next one succeeds.
        - If an iOS privacy policy was the cause, I suppose consecutive retries would finish with the same error. It is hard to judge, though, because this is not well documented.
        - On the other hand, if a race with another app caused the failure, then once that app finishes its activity it typically announces an interruption-ended notification as well, which lets our app try to recover.
    - If a reproducible scenario is found in which an active retry policy is necessary, it can be considered again.
    - Relatedly, I do not think it is useful for a JS app developer to handle native iOS errors (for the modules affected by this PR) themselves. If there is a use case for that, it can be considered again. Such a feature would require an architectural update, because the codebase is not currently prepared to surface native iOS errors to JS.

**Consumer note - `mixWithOthers` / `CannotInterruptOthers`**

- Documented: mixable `playAndRecord` is required to avoid `CannotInterruptOthers` when activating in the background, and is not sufficient for background restart. The Record demo enables `mixWithOthers` with `default` mode.

### Additional useful tweaks and fixes

- **Retry on foreground.** Once recovery from an interruption fails and does not lead to an audio-engine transition, engine recovery is retried when the app returns to the foreground via `WillEnterForegroundNotification` and `DidBecomeActiveNotification`. iOS guarantees that a foreground audio start succeeds.
- **Generalized `handleInputConfigurationChange` to handle errors.** A failed engine resume now finalizes the audio recorder.
- **File writers after capture loss.** Capture loss closes the current segment. A later successful resume (`HardwareChanged`) must open a new rotating segment instead of calling `rotateFiles()` on a writer that is already closed. `IOSRotatingFileWriter::reprepareStreamFormat` reopens when the inner file is closed; `rotateFiles()` no-ops if nothing is open.
- **Player detach on failed `startIfNecessary`**, matching the recorder. A later recorder start must not play leftover source audio.
- **Record demo:** `observeAudioInterruptions(true)`; freeze the waveform on `began`, unfreeze only on `ended`.

### Docs

- New `AudioManager` subsection "Resume recording after an interruption".
- Extended instructions related to iOS audio session configuration.

### Manual tests and reproductions used while debugging

This issue is hard to reproduce with automated tests. During research and implementation I used the following manual tests, with the `apps/fabric-example` app and the included Recorder demo.

* Start record; go background; interrupt with Siri; stop record
* Start record; go background; interrupt with Siri; tell it to open FabricExample; stop record
* Start record; go background; interrupt with meet.jit.si; disconnect; stop record
* Start record; go background; interrupt with meet.jit.si; disconnect and close the Safari tab; stop record
* Start record; interrupt with FaceTime; stop record
* Start record; go background; interrupt with FaceTime; stop record
* Record a short take and stop; start exclusive audio (could stay in foreground and accept FaceTime); play the recorded take; end the exclusive audio; end the recorded playback before it ends; record again
* Record a short take and stop; start exclusive audio (could stay in foreground and accept FaceTime); play the recorded take; end the exclusive audio; wait until playback ends; record again
* Record a short take and stop; start exclusive audio (could stay in foreground and accept FaceTime); play the recorded take and wait until playback reaches the end; end the exclusive audio; end the recorded playback; record again
* Record a short take and stop; start exclusive audio (could stay in foreground and accept FaceTime); play the recorded take and wait until playback reaches the end; end the exclusive audio; end the recorded playback; record again (do this step quickly enough after ending the exclusive audio)

> **Safari WebRTC microphone session side note.** One of the most mysterious errors I found, while experimenting with a meet.jit.si interruption, is the following:
>
> ```
> Error while restarting the audio engine after interruption: Error Domain=com.apple.coreaudio.avfaudio Code=2003329396 "(null)" UserInfo={failed call=err = PerformCommand(*ioNode, kAUStartIO, NULL, 0)}
> ```
>
> It is the result of executing `[self.audioEngine startAndReturnError:&error]` in `AudioEngine`'s `onInterruptionEnd`. Code `2003329396` corresponds to the `'what'` FourCC, which is `AVAudioSessionErrorCodeUnspecified`. It is not well documented, and I am unable to resume the session in the background with this error even with multiple active delayed retries, until I return to the foreground. I have two hypotheses that explain it: the first is iOS privacy policy; the second is that the interruption modifies the shared iOS audio session settings, after which a recovery to the original ones is impossible in the background.
>
> I mention it because I thought it was a good example for an active retry, but such a policy does not bear fruit.

## Checklist

- [x] Linked relevant issue
- [ ] Updated Web Audio API coverage - N/A (iOS session / engine, not a Web Audio node)
- [ ] Added support for web - N/A
- [ ] Updated old arch android spec file - N/A (no TurboModule spec change)


apps/CLAUDE.md
```diff
@@ -281,6 +281,10 @@ if (status !== 'Granted') return;
 
 ### Interruption handling
 
+Enable emission with `AudioManager.observeAudioInterruptions(true)`, then listen.
+
+**Playback:** pause on `began` (native does not resume players). The AudioFile example resumes on `ended` when it had been playing.
```
maciejmakowski2003, line 286:
```suggestion
**Playback:** native pause on `began`, but does not resume any audio contexts. The AudioFile example resumes on `ended` when it had been playing.
```

packages/audiodocs/docs/system/audio-manager.mdx
maciejmakowski2003, line 1: could you research and add a section about background resume for playback? I am quite sure that below audio session settings allow it.

```ts
AudioManager.setAudioSessionOptions({
    iosCategory: 'playback',
    iosMode: 'spokenAudio',
  })
```

packages/react-native-audio-api/ios/audioapi/ios/system/SystemNotificationManager.mm
```diff
@@ -108,6 +118,60 @@ - (void)observeValueForKeyPath:(NSString *)keyPath
...
+- (void)emitInterruptionBeganIfAccepted:(bool)accepted
+{
+  if (!self.audioInterruptionsObserved || !accepted) {
+    return;
+  }
+
+  [self.audioAPIModule invokeHandlerWithEventName:audioapi::AudioEvent::INTERRUPTION
+                                          payload:audioapi::InterruptionPayload{
+                                                      .type = "began", .shouldResume = false}];
+}
```
maciejmakowski2003, line 140: let's inline it

packages/react-native-audio-api/ios/audioapi/ios/system/SystemNotificationManager.mm
```diff
@@ -108,6 +118,60 @@ - (void)observeValueForKeyPath:(NSString *)keyPath
...
+    [self emitInterruptionEndedIfTransitioned:outcome shouldResume:shouldResume];
+  });
+}
+
+- (void)retryInterruptedRecordingIfNeeded
+{
+  AudioEngine *audioEngine = self.audioAPIModule.audioEngine;
+
+  if (self.interruptionEndedDelivered && [audioEngine getState] == AudioEngineStateInterrupted) {
+    [self performInterruptionEndOnEngine:audioEngine shouldResume:true];
```
maciejmakowski2003, line 171: verify hardcoded shouldResume: true

maciejmakowski2003 (changes requested):
could you perform the same research on android? in perfect world we would like to provide SotA for interruptions om both platforms.

packages/react-native-audio-api/ios/audioapi/ios/system/AudioEngine.mm
```diff
@@ -428,22 +430,37 @@ - (void)markSessionDeactivationInvalidatedGraph
...
+    [self stopEngine];
+    [self rebuildAudioEngine];
+    self.state = AudioEngineState::AudioEngineStatePaused;
+    [self notifyInput:AudioEngineInputNotificationHardwareChanged];
+    return AudioEngineInterruptionEndOutcomePaused;
+  }
+
+  if (![self.sessionManager ensureActive:true error:&error]) {
+    NSLog(@"Error while activating audio session after interruption: %@", [error debugDescription]);
+    return AudioEngineInterruptionEndOutcomeStillInterrupted;
```
maciejmakowski2003, line 452: what if we have both player and recorder paused <==> engine paused and interruption happens? does it mean that engine will auto-start after interruption

packages/react-native-audio-api/ios/audioapi/ios/system/SystemNotificationManager.mm
SomePersonFromMars: Objective C doesn't really have static methods, because sending messages always happens in runtime via dynamic dispatch. The closest thing we can get would be the following:

```objc
static inline void privateHelper(MyClass *self, NSString *value) {
    [self doSomething];
}

@implementation MyClass
- (void)publicMethod {
    privateHelper(self, @"Hello");
}
@end
```

I think, though, that it would introduce too much complexity.

PR #1260, opened 2026-09-01
fix(ios)!: inputNode race condition

closetcaiman:
## ⚠️ Breaking changes ⚠️

- N/A

## Introduced changes

- added a mutex on `getLiveInputNode` path so that it does not race with `AudioRecorder::start` which caused `use-after-free` and a hard crash.
- modified the mutex do be [recursive](https://en.cppreference.com/cpp/thread/recursive_mutex), so that it does not deadlock when re-locking from `notifyConfiguration` path.

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web

### DIFF
--- a/packages/react-native-audio-api/ios/audioapi/ios/system/AudioEngine.mm
+++ b/packages/react-native-audio-api/ios/audioapi/ios/system/AudioEngine.mm
@@ -172,13 +172,11 @@ static AudioEngine *_sharedInstance = nil;
 {
   std::scoped_lock lock(_engineLock);
 
-  AVAudioEngine *engine = self.audioEngine;
-
-  if (engine == nil) {
+  if (self.audioEngine == nil) {
     return nil;
   }
 
-  AVAudioInputNode *engineInputNode = engine.inputNode;
+  AVAudioInputNode *engineInputNode = self.audioEngine.inputNode;
 
   if (engineInputNode == nil) {
     return nil;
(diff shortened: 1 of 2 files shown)


packages/react-native-audio-api/ios/audioapi/ios/system/AudioEngine.mm
```diff
@@ -170,11 +170,15 @@ - (void)materializeSourceNodeWithId:(NSString *)sourceNodeId
 
 - (AVAudioFormat *)liveInputFormat
 {
-  if (self.audioEngine == nil) {
+  std::scoped_lock lock(_engineLock);
+
+  AVAudioEngine *engine = self.audioEngine;
```
SomePersonFromMars: Saving this pointer here is unnecessary and could be confusing, as the mutex is already guarding it. Even assuming that another method doesn't lock, it doesn't follow the contract then.

SomePersonFromMars (changes requested):

packages/react-native-audio-api/ios/audioapi/ios/system/AudioEngine.mm
closetcaiman: I had a log there before. Good to catch that, thanks!

SomePersonFromMars (approved):
Great job!

maciejmakowski2003 (approved):

Issue #1266, opened 2026-09-02
[iOS] Overnight background audio: AudioBuffer memory freed only by Hermes GC finalizers, steady per-native-call leak, and cancelScheduledValues not cancelling in-flight automation


catone:
### Description

We're building a generative sleep-soundscape app (`UIBackgroundModes: audio`, plugin `iosBackgroundMode: true`) that plays continuously for 8+ hours with the screen locked. Across multiple overnight device tests the app was repeatedly killed by jetsam with a footprint of ~1.9–2.75GB (kill reason `proc-thrashing` on the final kills; state `active` since the audio session keeps it scheduled). Steady-state audio memory for our loaded kit is well under 500MB, so ~2GB+ of that is leaked.

We progressively rearchitected the app to isolate causes, and found two distinct native memory leaks plus one verified `AudioParam` automation bug (we initially suspected two others — stale `.value` reads and stacked-`setTargetAtTime` glitches — and ruled both out by direct measurement; details in item 3, since the ruled-out hypotheses may save you triage time). Filing together because they came from one investigation; happy to split into separate issues if you prefer.

#### 1. Native memory behind a dropped `AudioBuffer` is only freed by Hermes GC finalizers — an idle background app never runs GC, so every mid-session decode leaks until jetsam

Our original design decoded bed stems just-in-time and dropped the JS reference when a stem rotated out. Each decode+drop cycle grew the footprint by ~23MB (the decoded PCM size), and that memory was never reclaimed while the app idled in the background — Hermes GC simply doesn't run when the JS thread is nearly idle, so the finalizers that release the native buffer never fire. The footprint climbed monotonically all night until jetsam at ~2.6GB.

Evidence that this is GC-dependence rather than an app-side retain: in the foreground (simulator and device with normal UI/JS activity driving GC), the identical code produces a sawtooth — the buffers *are* eventually freed; only the idle-background case climbs monotonically. (We also added a periodic `globalThis.gc?.()` call as a defensive measure, but could not confirm it is exposed in release Hermes builds, so we can't say whether an explicit GC would have reclaimed the memory.)

[...]

PR #1287, opened 2026-09-15
fix: handle current event in the cancel events also

mdydek:
Closes part of #1266, part about `cancelScheduledValues`

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

-

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.cpp
```diff
@@ -104,7 +104,34 @@ float ParamRenderQueue::getValueOfPreviousEventAt(const RenderParamEvent &event,
...
+  // An event may already have been promoted out of the queue; the erase above cannot see it
+  if (currentEvent_ && currentEvent_->getAutomationTime() >= cancelTime) {
+    restoreValueFromBeforeCurrentEvent();
+  }
+}
+
+void ParamRenderQueue::restoreValueFromBeforeCurrentEvent() {
+  currentEvent_ = ParamRenderEventFactory::createSetValueEvent(
+      currentEvent_->getStartValue(), currentEvent_->getStartTime());
+}
```
closetcaiman: This method is used exactly in one place and replaces 1 LOC. Just inline this.

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.cpp
```diff
@@ -104,7 +104,34 @@ float ParamRenderQueue::getValueOfPreviousEventAt(const RenderParamEvent &event,
...
+
 void ParamRenderQueue::cancelAndHoldAtTime(double cancelTime) {
+  // An in-flight ramp lives in currentEvent_, not in the queue
+  if (currentEvent_ && currentEvent_->isRampType() && cancelTime < currentEvent_->getEndTime()) {
+    truncateCurrentEventAt(cancelTime);
+    eventQueue_.erase(eventQueue_.upperBound(cancelTime), eventQueue_.end());
+    return;
+  }
+
```
closetcaiman, line 132: I'm not a fan of this being an early return. It does not match the style of the spec-defined step purging. I'd opt for merging the truncation into the step 3 with eventual erase as fallthrough:

```cpp
if (currentEvent_ && currentEvent_->isRampType() && cancelTime < currentEvent_->getEndTime()) {
truncateCurrentEventAt(cancelTime);
} else if (e2It != eventQueue_.end() && e2It->isRampType()) {
  ...
} else {
  ...
}

eventQueue_.erase(eventQueue_.upperBound(cancelTime), eventQueue_.end());
```

closetcaiman (changes requested):

maciejmakowski2003 (approved):

packages/react-native-audio-api/common/cpp/audioapi/core/utils/param/ParamRenderQueue.cpp
maciejmakowski2003: I would keep it. it is easier to understand the code with single responsibility methods split

closetcaiman (approved):

PR #1288, opened 2026-09-15
refactor: each context get its own event queue

mdydek:
Closes #

## ⚠️ Breaking changes ⚠️

-

## Introduced changes

- fixes sometimes happening bugs especially in wpt tests, in which there were many context created which used the same queue for the events and it crashed the token invariant introducing multiple producer on single producer channel, now each context gets its own token

## Checklist

- [ ] Linked relevant issue
- [ ] Added support for web


packages/react-native-audio-api/common/cpp/audioapi/events/AudioEventProducer.h
maciejmakowski2003, line 1: verify comments

packages/react-native-audio-api/common/cpp/audioapi/events/DeferredEventQueue.hpp
```diff
@@ -74,16 +78,20 @@ class DeferredEventQueue {
...
-        deferred.event, deferred.callbackId, AudioEventPayload{EmptyPayload{}});
+        *audioEventProducer_,
+        deferred.event,
+        deferred.callbackId,
+        AudioEventPayload{EmptyPayload{}});
   }
 
   BoundedPriorityQueue<DeferredEvent, MAX_PENDING_EVENTS, ByDueTime> pending_;
   std::shared_ptr<IAudioEventHandlerRegistry> registry_;
+  std::shared_ptr<AudioEventProducer> audioEventProducer_;
```
maciejmakowski2003, line 94: does it really has to hold audioEventProducer_?

maciejmakowski2003 (changes requested):
verify comments

packages/react-native-audio-api/common/cpp/audioapi/events/DeferredEventQueue.hpp
mdydek: why not, if this queue is designed to always call events from audio thread, why it cannot capture it?

maciejmakowski2003 (approved):

